PR #2, opened 2019-09-13
Delprete focchi

mfocchi:
Adding friction cones constraints in the WF
adding cost for obstacle avoidance  
added heightmap class
some bugfixes 
added paraller thread for visualization


cmastalli:
Two things:
 1. This is a mirror repository, so you have to create PR here: https://gepgitlab.laas.fr/loco-3d/crocoddyl
 2. The PR has to be done to the devel branch.

PR #604, opened 2019-11-20
CoMPosition: add set_cref to update reference, ShootingProblem: add x0 setter

wxmerkt:


cmastalli:
Thank you for adding these extra functionalities.

Issue #282, opened 2019-11-18
Replace asssert with if + throw exception


wxmerkt:
In GitLab by @wxmerkt on Nov 5, 2019, 17:32

As in Pinocchio - function argument and sanity checks should not be `assert`.

PR #627, opened 2019-11-30
Replace assert by if + throw exception #282

cmastalli:
This PR is mainly focused on #282.

Additionally, it creates setter function of the reference values in all the cost functions.

Finally, it creates the Python bindings of the the `ActivationNumDiff` class.


include/crocoddyl/core/utils/exception.hpp
```diff
@@ -0,0 +1,32 @@
...
+#ifndef CROCODDYL_CORE_UTILS_EXCEPTION_HPP_
+#define CROCODDYL_CORE_UTILS_EXCEPTION_HPP_
+
+#include <exception>
+#include <iostream>
+#include <sstream>
+
+namespace crocoddyl {
+
+class CrocoddylException : public std::exception {
```
wxmerkt: Why are we introducing a new exception type `CrocoddylException` rather than using `invalid_argument` for instance?

src/core/integrator/euler.cpp
```diff
@@ -31,6 +31,11 @@ IntegratedActionModelEuler::IntegratedActionModelEuler(boost::shared_ptr<Differe
     set_u_lb(differential_->get_u_lb());
     set_u_ub(differential_->get_u_ub());
   }
+  if (time_step_ < 0.) {
+    time_step_ = 1e-3;
+    time_step2_ = time_step_ * time_step_;
+    throw CrocoddylException("dt has positive value, set to 1e-3");
```
wxmerkt: So, here the values first get initialised to a default but then it throws nonetheless. Was this meant to be a warning?

src/core/numdiff/action.cpp
```diff
@@ -1,7 +1,8 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2018-2019, LAAS-CNRS, New York University, Max Planck Gesellshaft
+// Copyright (C) 2018-2020, LAAS-CNRS, University of Edinburgh, New York University,
+// Max Planck Gesellshaft
```
wxmerkt: Typo in German: "Gesellschaft" (c missing)

src/core/numdiff/activation.cpp
```diff
@@ -1,7 +1,8 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2018-2019, LAAS-CNRS, New York University, Max Planck Gesellshaft
+// Copyright (C) 2018-2020, LAAS-CNRS, University of Edinburgh, New York University,
+// Max Planck Gesellshaft
```
wxmerkt: Typo as above

src/core/numdiff/diff-action.cpp
```diff
@@ -1,7 +1,8 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2018-2019, LAAS-CNRS, New York University, Max Planck Gesellshaft
+// Copyright (C) 2018-2020, LAAS-CNRS, University of Edinburgh, New York University,
+// Max Planck Gesellshaft
```
wxmerkt: Typo as above

src/core/numdiff/state.cpp
```diff
@@ -1,7 +1,8 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2018-2019, LAAS-CNRS, New York University, Max Planck Gesellshaft
+// Copyright (C) 2018-2020, LAAS-CNRS, University of Edinburgh, New York University,
+// Max Planck Gesellshaft
```
wxmerkt: Typo as above

src/multibody/costs/control.cpp
```diff
@@ -64,4 +76,6 @@ void CostModelControl::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
 
 const Eigen::VectorXd& CostModelControl::get_uref() const { return uref_; }
 
+void CostModelControl::set_uref(const Eigen::VectorXd& uref_in) { uref_ = uref_in; }
```
wxmerkt: We should check dimensions here

src/multibody/costs/state.cpp
```diff
@@ -97,4 +133,6 @@ boost::shared_ptr<CostDataAbstract> CostModelState::createData(pinocchio::Data*
 
 const Eigen::VectorXd& CostModelState::get_xref() const { return xref_; }
 
+void CostModelState::set_xref(const Eigen::VectorXd& xref_in) { xref_ = xref_in; }
```
wxmerkt: We should check dimensions here

include/crocoddyl/core/utils/exception.hpp
cmastalli: Because we need to register the exceptions in the Python world. Otherwise, when the exception is triggered in Python, you will get message saying unidentified c++ exception.

This is the recommended approach in boost python

src/core/integrator/euler.cpp
cmastalli: The step time cannot be negative. If the user made stupy things, then the policy is rewrite a default dt value and let the user knows about it.

cmastalli:
@wxmerkt do you why the jobs aren't passing? I didn't change that code, so I imagine it's something related to example-robot-data.

Additionally, it seems the version of yapf has changed. So the format isn't passing. I couldn't see the version. Yapf is annoying, in every version the format changes :(

include/crocoddyl/core/utils/exception.hpp
wxmerkt: I have been throwing std::invalid_argument and Boost Python has correctly translated them without this message. This is how it's done in Pinocchio for instance

src/core/integrator/euler.cpp
wxmerkt: Yes, that is clear. However: There is no point in first assigning th values and then throwing an exception. Either assign a default and show a warning, or throw an exception.

wxmerkt:
The build jobs are failing due to unit tests. We are using the latest robotpkg packages - if they are up to date it should pass?

Can you please let me know which version of yapf we should be installing? We can then fix it so this won't happen again

cmastalli:
Uhm ... I will double check the unittest, but it seems weird since I didn't change and they passed in my PC.

We're using yapf 0.28.0

wxmerkt:
yapf version is now fixed to 0.28.0. If you rebase on devel it should trigger it using the correct version

include/crocoddyl/core/utils/exception.hpp
cmastalli: Done, thanks for the tip

src/core/integrator/euler.cpp
cmastalli: I decided to show warning with `std::cerr`

src/core/numdiff/action.cpp
cmastalli: Thanks for catching these typos

src/multibody/costs/control.cpp
cmastalli: done

src/multibody/costs/state.cpp
cmastalli: done

src/core/solvers/ddp.cpp
```diff
@@ -212,17 +212,17 @@ void SolverDDP::backwardPass() {
     }
 
     if (raiseIfNaN(Vx_[t].lpNorm<Eigen::Infinity>())) {
-      throw CrocoddylException("backward_error");
+      throw std::invalid_argument("backward_error");
```
wxmerkt: These are probably `std::runtime_error` or similar

src/core/solvers/ddp.cpp
```diff
@@ -212,17 +212,17 @@ void SolverDDP::backwardPass() {
...
     }
     if (raiseIfNaN(Vxx_[t].lpNorm<Eigen::Infinity>())) {
-      throw CrocoddylException("backward_error");
+      throw std::invalid_argument("backward_error");
```
wxmerkt: These are probably `std::runtime_error` or similar

src/core/solvers/ddp.cpp
```diff
@@ -237,10 +237,10 @@ void SolverDDP::forwardPass(const double& steplength) {
     cost_try_ += d->cost;
 
     if (raiseIfNaN(cost_try_)) {
-      throw CrocoddylException("forward_error");
+      throw std::invalid_argument("forward_error");
```
wxmerkt: These are probably `std::runtime_error` or similar

src/core/solvers/ddp.cpp
```diff
@@ -237,10 +237,10 @@ void SolverDDP::forwardPass(const double& steplength) {
...
     }
     if (raiseIfNaN(xs_try_[t + 1].lpNorm<Eigen::Infinity>())) {
-      throw CrocoddylException("forward_error");
+      throw std::invalid_argument("forward_error");
```
wxmerkt: These are probably `std::runtime_error` or similar

src/core/solvers/ddp.cpp
```diff
@@ -250,7 +250,7 @@ void SolverDDP::forwardPass(const double& steplength) {
   cost_try_ += d->cost;
 
   if (raiseIfNaN(cost_try_)) {
-    throw CrocoddylException("forward_error");
+    throw std::invalid_argument("forward_error");
```
wxmerkt: These are probably `std::runtime_error` or similar

src/core/solvers/fddp.cpp
```diff
@@ -182,10 +182,10 @@ void SolverFDDP::forwardPass(const double& steplength) {
     cost_try_ += d->cost;
 
     if (raiseIfNaN(cost_try_)) {
-      throw CrocoddylException("forward_error");
+      throw std::invalid_argument("forward_error");
```
wxmerkt: These are probably `std::runtime_error` or similar

src/core/solvers/fddp.cpp
```diff
@@ -182,10 +182,10 @@ void SolverFDDP::forwardPass(const double& steplength) {
...
     }
     if (raiseIfNaN(xnext_.lpNorm<Eigen::Infinity>())) {
-      throw CrocoddylException("forward_error");
+      throw std::invalid_argument("forward_error");
```
wxmerkt: These are probably `std::runtime_error` or similar

src/core/solvers/fddp.cpp
```diff
@@ -201,7 +201,7 @@ void SolverFDDP::forwardPass(const double& steplength) {
   cost_try_ += d->cost;
 
   if (raiseIfNaN(cost_try_)) {
-    throw CrocoddylException("forward_error");
+    throw std::invalid_argument("forward_error");
```
wxmerkt: These are probably `std::runtime_error` or similar

src/core/solvers/ddp.cpp
cmastalli: Nice catch! done

src/core/solvers/ddp.cpp
cmastalli: done

cmastalli:
@MaximilienNaveau I have removed some assert unit-test in the multiple-impulse since they were substituted by `if + throw` exceptions

PR #630, opened 2019-12-02
Created a to_string function for c++98 compatibility

cmastalli:
I know we have discussed about to stick only with c++11 and higher. This solution can be quickly adapted if we decided to move forward.

### DIFF
--- a/bindings/python/crocoddyl/core/action-base.hpp
+++ b/bindings/python/crocoddyl/core/action-base.hpp
@@ -24,10 +24,10 @@ class ActionModelAbstract_wrap : public ActionModelAbstract, public bp::wrapper<
   void calc(const boost::shared_ptr<ActionDataAbstract>& data, const Eigen::Ref<const Eigen::VectorXd>& x,
             const Eigen::Ref<const Eigen::VectorXd>& u) {
     if (static_cast<std::size_t>(x.size()) != state_->get_nx()) {
-      throw std::invalid_argument("x has wrong dimension (it should be " + to_string(state_->get_nx()) + ")");
+      throw std::invalid_argument("x has wrong dimension (it should be " + std::to_string(state_->get_nx()) + ")");
     }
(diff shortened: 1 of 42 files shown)


wxmerkt:
Can we put this into the `std` namespace if the C++ version is less than 2011 to avoid having to change so many files when we do make the switch to C+11?

cmastalli:
> switch

done! I am waiting for the results on the gepgitlab CI. In github, we don't have 16.04 jobs; so I can't fully check it here. Should we include them in Github?

wxmerkt:
We can include 16.04 in Travis, but it may slow down the CI quite a bit (each additional build ~30min)

cmastalli:
> We can include 16.04 in Travis, but it may slow down the CI quite a bit (each additional build ~30min)

I see
@nim65s do you have any suggestion?

nim65s:
I'll suggest to keep travis configuration as minimal as possible. It should be able to show quickly simple issues, while we can use gitlab CI for long and exhaustive builds.
My point is I'll spend a lot of time on tuning gitlab CI anyway, so it would be better to avoid spending too much time in travis configuration and maintainance in individual projects.

cmastalli:
> I'll suggest to keep travis configuration as minimal as possible. It should be able to show quickly simple issues, while we can use gitlab CI for long and exhaustive builds.
> My point is I'll spend a lot of time on tuning gitlab CI for all projects at once anyway, so it would be better to avoid spending too much time in travis configuration and maintenance in individual projects.

For your the minimalist version of Travis only includes 18.04, don't you?

wxmerkt:
The Travis has already been set up for 16.04 and 18.04 now. We can take 16.04 out again if desired.

nim65s:
if including more than one distribution is strictly adding one more line with the distribution name, and that it doesn't require way more time to run, we can keep it. If this require any additionnal work, we shouldn't, IMO.

PR #633, opened 2019-12-02
Update Travis to include 16.04 builds

wxmerkt:
Replaces #632 

Test if this works and how much this slows down Travis/CI

Check output from fork if this works:

https://travis-ci.com/wxmerkt/crocoddyl/builds/138956994


cmastalli:
thanks for moving this forward

PR #635, opened 2019-12-02
Add Travis badge to README

wxmerkt:
For master branch. Alternatively could change to devel.


cmastalli:
I think the master is more meaningful for the users. Developers can always investigate deeper using Travis.

cmastalli:
Just a comment -- please do not push any branch into the main repo.

I want to keep it only with master and devel ;)

wxmerkt:
Yes, that is clear. This one was necessary to trigger the CI correctly to see if the matrix works. Which is why I closed #634 which was a PR from a fork.

Issue #640, opened 2019-12-03
Use const ref


jcarpent:
https://github.com/loco-3d/crocoddyl/blob/d59b9be9f6c3f90b07dda12e5c14d8889cdc2214/src/core/optctrl/shooting.cpp#L114

Issue #631, opened 2019-12-02
Contributing guidelines vs credits


wxmerkt:
The `CONTRIBUTING.md` currently lists who has contributed in the past (list of contributors). However, GitHub interprets this special file as the "Contributing guidelines". So when someone opens an issue/pull request they are shown "Contributing guidelines", which aren't guidelines but a list of contributors. Is this intended?

![image](https://user-images.githubusercontent.com/1664508/69956687-8f7cff00-14f8-11ea-9b2c-baab5ea9fb02.png)

cmastalli:
is this the same https://github.com/loco-3d/crocoddyl/wiki/Developer-Guide?

wxmerkt:
Adding the content of the wiki link you said into CONTRIBUTING.md would likely do the trick (or no CONTRIBUTING.md at all). The point is: Having credits in CONTRIBUTING.md is confusing

cmastalli:
> Adding the content of the wiki link you said into CONTRIBUTING.md would likely do the trick (or no CONTRIBUTING.md at all). The point is: Having credits in CONTRIBUTING.md is confusing

Roger that! I created an issue #634

proyan:
I believe this is done. I'll close this.

PR #642, opened 2019-12-03
Used const ref for all setter functions + updated CONTRIBUTING file

cmastalli:
This PR solves two issues: #640 and #631.


include/crocoddyl/core/integrator/euler.hpp
```diff
@@ -28,7 +28,7 @@ class IntegratedActionModelEuler : public ActionModelAbstract {
 
   const boost::shared_ptr<DifferentialActionModelAbstract>& get_differential() const;
   const double& get_dt() const;
-  void set_dt(double dt);
+  void set_dt(const double& dt);
```
wxmerkt: For primitive types there is no benefit of using const-ref. The cost of creating a reference and the size of it is effectively the same as passing by value and thus creating a copy.

include/crocoddyl/core/solvers/ddp.hpp
```diff
@@ -56,13 +56,13 @@ class SolverDDP : public SolverAbstract {
   const std::vector<Eigen::VectorXd>& get_k() const;
   const std::vector<Eigen::VectorXd>& get_gaps() const;
 
-  void set_regfactor(double reg_factor);
-  void set_regmin(double regmin);
-  void set_regmax(double regmax);
+  void set_regfactor(const double& reg_factor);
```
wxmerkt: As above, there is no point in using const ref for primitive types

include/crocoddyl/multibody/costs/cost-sum.hpp
```diff
@@ -143,12 +143,12 @@ struct CostDataSum {
   Eigen::MatrixXd get_Lxu() const { return Lxu; }
   Eigen::MatrixXd get_Luu() const { return Luu; }
 
-  void set_r(Eigen::VectorXd _r) { r = _r; }
-  void set_Lx(Eigen::VectorXd _Lx) { Lx = _Lx; }
-  void set_Lu(Eigen::VectorXd _Lu) { Lu = _Lu; }
-  void set_Lxx(Eigen::MatrixXd _Lxx) { Lxx = _Lxx; }
-  void set_Lxu(Eigen::MatrixXd _Lxu) { Lxu = _Lxu; }
-  void set_Luu(Eigen::MatrixXd _Luu) { Luu = _Luu; }
+  void set_r(const Eigen::VectorXd& _r) { r = _r; }
```
wxmerkt: I believe we should add size checks here for all following methods

include/crocoddyl/core/integrator/euler.hpp
cmastalli: I know about the primitives, however I believe it's a good rule of thumb to use it in these cases.

wxmerkt:
Further reading on the dont-pass-primitive-types-by-const-ref:

- https://stackoverflow.com/a/14013189
- https://stackoverflow.com/a/673836

Point being it's counter-productive as it may break the compiler's built-in optimisation

include/crocoddyl/core/integrator/euler.hpp
cmastalli: @jcarpent, @nim65s and @olivier-stasse 

Should we use only object for primitive types (double, int, bool)?

include/crocoddyl/multibody/costs/cost-sum.hpp
cmastalli: Done. Thanks for the feedback

cmastalli:
> Further reading on the dont-pass-primitive-types-by-const-ref:
> 
>     * https://stackoverflow.com/a/14013189
> 
>     * https://stackoverflow.com/a/673836
> 
> 
> Point being it's counter-productive as it may break the compiler's built-in optimisation

According to first link, the double are a little bigger than its reference. So it is questionable if we should pass only objects for double cases.

include/crocoddyl/core/integrator/euler.hpp
nim65s: I'd say no, we should just keep `(double dt)` for primitives types, but I don't have references about this pattern right now.

wxmerkt:
> Furthermore, I don't think extra aliasing matters for setter functions. These functions will be typically call once during configuration time, and rarely call again at runtime.

Than why do the premature and unnecessary optimisation which may break built-in compiler optimisation?

include/crocoddyl/core/integrator/euler.hpp
cmastalli: @nim65s Sorry I didn't get your point. Do you mean "use `double`, instead of `const double&`"?

According to this link (https://stackoverflow.com/a/14013189), the size of a `double` are bigger than reference. So, for `double`, it could be better to use `const&` if extra aliasing it's not an issue (which I believe for these functions the answer is "no issue"`)

wxmerkt: I'd say: If it's a primitive, pass by value. Only pass by const-ref for larger objects. So for here `(double dt)`

From the same link:
> pass-by-value avoids aliasing, allowing the optimizer to really do its thing

nim65s: @cmastalli : yes, I mean "use `double`, instead of `const double&`", but if you want to benchmark it, and if that benchmark shows a good (performances gains / complexity loss) ratio, we can discuss it.

CONTRIBUTING.md
```diff
@@ -1,21 +1,63 @@
-[Carlos Mastalli](https://cmastalli.github.io/) (University of Edinburgh): main developer and manager of the project
+## I. About
+Crocoddyl has been written in C++ for efficiency and uses the  Eigen  library  for  linear  algebra  routines.  It  comes  with Python bindings for easy prototyping. Crocoddyl is currently supported  for  most  Linux  distributions,  with  plans  to  release on Mac OS X and Windows. The project is fully open-source under  the  permissive  BSD-3-Clause  license  and  is  hosted on  GitHub.
 
-[Nicolas Mansard](http://projects.laas.fr/gepetto/index.php/Members/NicolasMansard) (LAAS-CNRS): project instructor
+## II. Design pattern
+Crocoddyl is designed using virtualization pattern for easy prototyping and yet efficient implementation. Virtualization have shown to be more efficient than static polymorphism through CRTP for system dynamics higher than 1. The real benefits of static polymorphism are in very small system (with dimension less than 6), however virtualization is still competitive.
```
jcarpent: The second part of this paragraph seems to be over claiming. In general, static polymorphism by using CRTP and/or Variant are much more efficient.

cmastalli: I benchmark it. For this particular case, it was worse for humanoid problems, same for quadruped and lightly better for unicycle.

jcarpent: But when I've seen your bench, it was not correctly done I would say. I would remove this statement. Think to Eigen for instance ;)

include/crocoddyl/core/integrator/euler.hpp
jcarpent: In my case, I would just use const double to say to the compiler that this quantity won't be modified ;). But I do agree with @wxmerkt, this is not necessary at first glance, except when you use AutoDiiff frameworks!

include/crocoddyl/multibody/costs/cost-sum.hpp
```diff
@@ -143,12 +143,45 @@ struct CostDataSum {
...
+    Lx = _Lx;
+  }
+  void set_Lu(const Eigen::VectorXd& _Lu) {
+    if (Lu.size() != _Lu.size()) {
+      throw std::invalid_argument("Lu has wrong dimension (it should be " + std::to_string(Lu.size()) + ")");
+    }
+    Lu = _Lu;
+  }
+  void set_Lxx(const Eigen::MatrixXd& _Lxx) {
+    if (Lxx.rows() != _Lxx.rows() && Lxx.cols() != _Lxx.cols()) {
```
jcarpent: This statement is not correct. It should be an OR test.

jcarpent: The same applies to other places.

jcarpent:
@cmastalli I think you're going too fast when merging. You should wait a bit from external review if you want to avoid mistakes.

jcarpent:
In addition, you should test your assert, or at least provides some unit tests for each function you're adding: it takes some additional times when implementing it, but it will save the time of everybody in the future.

include/crocoddyl/multibody/costs/cost-sum.hpp
cmastalli: Thanks for the feedback! I fixed it, we'll see it in the next PR.

include/crocoddyl/core/integrator/euler.hpp
cmastalli: @jcarpent thanks for the feedback.

CONTRIBUTING.md
cmastalli: I have relaxed this claim by basically saying that it is competitive design. I will add in the next PR. Thanks for the feedback!

Issue #615, opened 2019-11-27
Sharing data for impulse-based cost functions


cmastalli:

proyan:
Why was this closed? Has this been resolved already?

cmastalli:
> Why was this closed? Has this been resolved already?

Yes, it has. I will later add friction cone cost for impulse dynamics (now only for contact one).

Issue #295, opened 2019-11-18
Sharing data for force-based cost functions


wxmerkt:
In GitLab by @cmastalli on Nov 15, 2019, 17:17

null

wxmerkt:
In GitLab by @cmastalli on Nov 15, 2019, 17:17

assigned to @cmastalli

PR #644, opened 2019-12-05
Create data collector for sharing force, impulse and actuation data

cmastalli:
This PR allows to include cost functions that depend on the force, impulse or actuation data (#295 and #615). To do so, I created an abstraction for data collection with throw exception in case of users mistakes

Given an action data, we create a tailored data collection that is shared through any cost function. Different cost functions requires different kind of data collectors. And, in each cost function, we check that it has been passed a right data collector. This test is possible due to a proper design hierarchy of data collection abstractions.

This method doesn't affect the run-time performance due to casting. All needed castings are done inside the creation of a specific cost data. All code has Python bindings

Additionally, there are other minor tasks in this PR:
 1. removed claim about virtualization against CRTP
 2. fixed throw exception logic in cost-sum


src/multibody/costs/centroidal-momentum.cpp
```diff
@@ -61,20 +62,20 @@ void CostModelCentroidalMomentum::calcDiff(const boost::shared_ptr<CostDataAbstr
   Eigen::Ref<pinocchio::Data::Matrix6x> Rv = data->Rx.rightCols(nv);
 
   activation_->calcDiff(data->activation, data->r, recalc);
-  pinocchio::getCentroidalDynamicsDerivatives(state_->get_pinocchio(), *data->pinocchio, Rq, d->dhd_dq, d->dhd_dv, Rv);
+  pinocchio::getCentroidalDynamicsDerivatives(state_->get_pinocchio(), *d->pinocchio, Rq, d->dhd_dq, d->dhd_dv, Rv);
 
   // The derivative computation in pinocchio does not take the frame of reference into
```
jcarpent: Which quantity do you want to compute here?

jcarpent: It is because h_ref is expressed in the frame aligned with the world coordinates system expressed around the world frame?

jcarpent: If my statement is valid, then:
```
data->Rx.block<3, 1>(3, i) -= d->pinocchio->Jcom.col(i).cross(d->pinocchio->hg.linear()); 
```
and should be replaced by:
```
     data->Rx.block<3, 1>(3, i) -= d->pinocchio->Jcom.col(i).cross(h_ref);
```
I think.

unittest/bindings/test_force_cost.py
```diff
@@ -25,22 +29,16 @@
...
 
 x = ROBOT_STATE.rand()
 u = pinocchio.utils.rand(ACTUATION.nu)
 MODEL.calcDiff(DATA, x, u)
 MODEL_ND.calcDiff(dnum, x, u)
-np.allclose(DATA.Fx, dnum.Fx, atol=MODEL_ND.disturbance)
-np.allclose(DATA.Fu, dnum.Fu, atol=MODEL_ND.disturbance)
-np.allclose(DATA.Lx, dnum.Lx, atol=MODEL_ND.disturbance)
-np.allclose(DATA.Lu, dnum.Lu, atol=MODEL_ND.disturbance)
+assert (np.allclose(DATA.Fx, dnum.Fx, atol=1e5 * MODEL_ND.disturbance))
```
wxmerkt: I think this is a better option: https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_allclose.html

wxmerkt: Also, 1e5?

cmastalli:
@wxmerkt this tolerance value comes from a long discussion about NumDiff unittest (during Python development). The values depends on the disturbance used in the NumDiff routine, which it depends on the epsilon machine. For a 64bits machine, it's around 1e-4.

unittest/bindings/test_force_cost.py
cmastalli: I need to work on this unittest (and test_momentum_cost) later. I won't spend much time now (i.e. this PR) in improving it.
Thanks for the feedback, I will consider it for the next PR.

cmastalli:
The reason of the fails in the CI is due to a change in Pinocchio version, as reported #645 
I will proceed with this PR as it is.

wxmerkt:
> @wxmerkt this tolerance value comes from a long discussion about NumDiff unittest (during Python development). The value depends on the disturbance used in the NumDiff routine, which it depends on the epsilon machine. For a 64bits machine, it's around 1e-4.

We should add this comment in the code. If there are discussions which arent captured on issues or in comments, it's not easy to follow

wxmerkt:
This PR didn't just fail on Xenial but also on Bionic

cmastalli:
> This PR didn't just fail on Xenial but also on Bionic

I am fixing now. But there is still the problem with Pinocchio, I will pointing you out  to this issue in my next PR

PR #646, opened 2019-12-06
Adds a more verbose exception type and replaces asserts

VladimirIvan:
- Adds `crocoddyl::Exception` that prints file name, line number, method name and a message when thrown.
- Adds `throw_pretty(msg)` and `assert_pretty(condition, msg)` macros that construct and throw the above exception. `msg` is a string stream.
- Replaces `throw std::...` with `throw_pretty` in the whole repo.
- Replaces `assert` with `assert_pretty` in the whole repo. There is no performance hit on my PC when running the benchmarks.

Usage:
```
...
throw_pretty("Problem detected at iteration: " << i);
assert_pretty(size!=n, "Wrong size, got " << size << " expected " <<n);
```

Success story:
An assert muted in release build caused wrong size of MPC initialisation to segfault and kill ipython kernel. This took 2h to debug.
The new exception printed a message that resulted in a fix that took 10s to implement in the user code in python.


cmastalli:
@VladimirIvan thanks for this nice work :smiley: 

I have one question -- did you try the throw exceptions in Python? The reason is that I think you need to register a custom throw exception in boost python. See this post for more information: https://stackoverflow.com/questions/2261858/boostpython-export-custom-exception

jcarpent:
eigenpy is doing the job for you normally.

cmastalli:
@VladimirIvan I am handling the modifications, could you give me right to push my commits?

PR #647, opened 2019-12-06
Improve packaging + fix test

jcarpent:
@cmastalli As in #645 you did not provide the write access on top of your commits, I open this PR which provides different fixes with recent versions of Boost mainly.
It also provides support for OSX systems.
More will come soon to speed up (really!) the compilation of Python bindings.

PR #649, opened 2019-12-06
Formatting code

cmastalli:
This is a follow-up PR of #647 which only applies the right format.

Fyi @jcarpent, we're checking the code format for c++ and Python. You don't need to learn the code format, we use the Gepetto linter tools. See here for more information: https://github.com/loco-3d/crocoddyl/wiki/Developer-Guide


cmastalli:
@wxmerkt I forgot to pointing out. In this PR, I fixed the error that I introduced and I documented the numdiff routine.

Again these unittest are temporal, @MaximilienNaveau will write them in c++.

PR #651, opened 2019-12-06
Fixed throw exception in solvers + auto-indexing in Python bindings

cmastalli:
The previous PR #646 introduced a new custom exception for Crocoddyl. This is exception was not catch by the solvers during expected runtime errors in the backward and forward passes. This PR fixed this issue.

Additionally, it was reported the need of having auto-indexing Python bindings, i.e. to be able to write something like:
```python
model.calc(data=mydata, x=myx, u=myu)
```
which now it is possible.


src/core/utils/exception.cpp
```diff
@@ -1,12 +1,13 @@
...
 // BSD 3-Clause License
 //
-// Copyright (C) 2018-2019, LAAS-CNRS
+// Copyright (C) 2018-2020, University of Edinburgh
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
 #include "crocoddyl/core/utils/exception.hpp"
+#include "crocoddyl/core/utils/to-string.hpp"
```
wxmerkt: I would put this into the header "exception.hpp"

cmastalli: I always include header in the place that are needed. Is this against a good-coding practice? If so, could you point out to a refer for my own curiosity?

VladimirIvan: It's not needed at all, just use the string stream:
```
throw_pretty("Cost item doesn't have the same control dimension (it should be " << nu_ << ")");
```

wxmerkt: If you don't add it to the header, you will have to add it to every single file that uses `std::to_string`, which is in every place where we use `throw_pretty`

cmastalli: We don't need to include in each place with this methods, indeed, I didn't have to do it.

Modifying each throw_pretty is plausible too, but what happens if we change in future the method? I believe it's more flexible to use `std::to_string`

wxmerkt: I think you missed my point. My point was: in many files where we use `throw_pretty`, there is an implicit `std::to_string`. So each of those files needs an include to the header `to-string.hpp`. If you include this header in `exception.hpp`, it is automatically included already.

cmastalli: I saw your point, however, the files are compiling in Ubuntu 16.04 without need of including 'to-string.hpp'. Are you sure that your statement is right?

In my understanding, it's not needed because we're linking a library that contains it.

VladimirIvan:
I removed `std::to_string` and used a `std::stringstream` instead. That's why it now compiles on 16.04.

cmastalli:
It was compiling before your PR too. @VladimirIvan

PR #652, opened 2019-12-06
Fix CI for C++98 include of to-string.hpp

wxmerkt:


wxmerkt:
This can easily be added to #651. Closing

cmastalli:
> This can easily be added to #651. Closing

I agree. I just want to move faster things

PR #658, opened 2020-01-05
Important improment in box ddp and box qp

cmastalli:
This PR introduces important improvement in the box DDP and box QP solvers.

***1. Box QP: active set***
One of the main problem was due to a bug in the box QP, incorrect output of the active set. Additionally, our box QP is able to accept infeasible warm-start in contrast to the original proposed method by Tassa, this is done by a "cheap" projection to the feasible set. In conclusion, this PR proposes a completely new box QP solver, which is based on Cholesky factorization.

***2. Box DDP: infeasible warm-start and expected model***
Our box DDP is also different to the proposed method by Tassa in which an initial forward pass is needed for computing a feasible xs,us point to the backward pass. Note that if the first iteration is infeasible then we cannot properly predict the dus bounds. Therefore, I propose to run  a unbound backward pass (i.e. DDP backward pass) in such case; it shows great numerical behavior in all examples.

Furthermore, the expected model is properly computed by clamping the gradient direction (Qu) as done by the box QP. Finally, ascend direction can appears in the box DDP, so I updated the step acceptance in the DDP solver.

***3. Other minor tasks***
This PR includes unittest code for the the box QP and extra examples for control-bound cases (with its log files):
 - quadrotor
 - quadrupedal walk
 - bipedal walk
 - humanoid manipulation (new example using the Talos robot).

Additionally, I added benchmark code for the box QP, which shows an improvement with respect to previous code.
Note that I have included Python bindings for the box QP.

***4. Conclusion***
With this new box DDP, the number of iterations remains similar compared with the DDP solver (i.e. with probably few more iterations). The computation time remains to be benchmarked, however, most of the cases the QP is solved in 1 iteration, making it as fast as a unconstrained QP.


src/core/solvers/box-qp.cpp
```diff
@@ -0,0 +1,258 @@
...
+        for (std::size_t i = 0; i < nf_; ++i) {
+          const std::size_t& fi = free_idx_[i];
+          for (std::size_t j = 0; j < nf_; ++j) {
+            Hff_(i, j) = H(fi, free_idx_[j]);
+          }
+        }
+        if (reg_ != 0.) {
+          Hff_.diagonal().array() += reg_;
+        }
+        Hff_inv_llt_ = Eigen::LLT<Eigen::MatrixXd>(nf_);
```
jcarpent: You're allocating memory for nothing.

src/core/solvers/box-qp.cpp
```diff
@@ -0,0 +1,258 @@
...
+        Hff_inv_llt_.compute(Hff_);
+        Eigen::ComputationInfo info = Hff_inv_llt_.info();
```
jcarpent: Use reference here.

src/core/solvers/box-qp.cpp
```diff
@@ -0,0 +1,258 @@
...
+        if (info != Eigen::Success) {
+          throw_pretty("backward_error");
+        }
+        Hff_inv_.setIdentity(nf_, nf_);
+        Hff_inv_llt_.solveInPlace(Hff_inv_);
+      }
+      solution_.Hff_inv = Hff_inv_;
```
jcarpent: You can avoid one copy here, by working directly with `solution_.Hff_inv`

src/core/solvers/box-qp.cpp
```diff
@@ -0,0 +1,258 @@
...
+    }
+    Hff_inv_llt_.solveInPlace(dxf_);
+    dxf_ -= xf_;
+    dx_.setZero();
+    for (std::size_t i = 0; i < nf_; ++i) {
+      dx_(free_idx_[i]) = dxf_(i);
+    }
+
+    // Try different step lengths
+    double fold = 0.5 * x_.dot(H * x_) + q.dot(x_);
```
jcarpent: You may avoid some memory allocation here.

src/core/solvers/box-qp.cpp
```diff
@@ -0,0 +1,258 @@
...
+    for (std::vector<double>::const_iterator it = alphas_.begin(); it != alphas_.end(); ++it) {
+      double steplength = *it;
+      for (std::size_t i = 0; i < nx_; ++i) {
+        xnew_(i) = std::max(std::min(x_(i) + steplength * dx_(i), ub(i)), lb(i));
+      }
+      double fnew = 0.5 * xnew_.dot(H * xnew_) + q.dot(xnew_);
```
jcarpent: The same applies here.

src/core/solvers/box-qp.cpp
```diff
@@ -0,0 +1,258 @@
...
+    std::cerr << "Warning: reg value has to be positive." << std::endl;
+  }
+
+  // Reserve the space and compute alphas
+  clamped_idx_.reserve(nx_);
+  free_idx_.reserve(nx_);
+  const std::size_t& n_alphas_ = 10;
+  alphas_.resize(n_alphas_);
+  for (std::size_t n = 0; n < n_alphas_; ++n) {
+    alphas_[n] = 1. / pow(2., static_cast<double>(n));
```
jcarpent: This version of pow is inefficient, but it might not be a serious problem as it is called once.

src/core/solvers/box-qp.cpp
cmastalli: The dimension of free-set changes dynamically ("nf_"). So, I can't avoid this operation. Is there a better way to do it (e.g. to reserve memory) in which I am not aware of?

src/core/solvers/box-qp.cpp
cmastalli: done

src/core/solvers/box-qp.cpp
cmastalli: done. I also avoid copy of the free and clamped indexes

src/core/solvers/box-qp.cpp
cmastalli: done

src/core/solvers/box-qp.cpp
cmastalli: done

src/core/solvers/box-qp.cpp
cmastalli: Yes, I preferred to use it (for code readability) since it's needed only at config-time.

src/core/solvers/box-qp.cpp
jcarpent: You're doing the allocation twice here (initialization + copy operator). So, the answer is yes, you can do better.

cmastalli: I see your point. I will provide a solution in the next PR.

Thanks for code review!

PR #660, opened 2020-01-08
Fixed CI-github CI issue for 16.04-Debug + added examples in CI-github

cmastalli:
This PR fixes the CI issue by compiling Crocoddyl using c++11 standard (due to Pinocchio is installed using c++11).

It also adds the example for CI-Release jobs; debug cases are too slow. This solution doesn't substitute the proposed method by @nim65s. @nim65s - do you think we should propagate this solution to gepgitlab? or wait for the final solution?


.travis/build-and-test.sh
```diff
@@ -4,6 +4,29 @@ set -xe
...
+fi
+
 make -j1
 make test
+
+# Run the examples in Release mode
+if [ $CMAKE_BUILD_TYPE == "Release" ]
+then
+  # Examples
+  make -s examples-double_pendulum
```
wxmerkt: Why do these need to be made separately?!

nim65s:
I don't follow you. propagate which solution to gepgitlab ?

nim65s:
Got it, I'll open another PR.

.travis/build-and-test.sh
cmastalli: @nim65s's PR have changes it. Now they are added as unittests. Check it out this for further information: https://github.com/loco-3d/crocoddyl/pull/662/files

PR #662, opened 2020-01-08
Run examples as tests

nim65s:
Follows #660 

While here, I saw that other python files in `examples/notebook/*.py` ; shouldn't we also run those as tests ?
@imaroger had issues with `cartpole_swing_up.py`.

### DIFF
--- a/.travis/build-and-test.sh
+++ b/.travis/build-and-test.sh
@@ -14,19 +14,3 @@ fi
 
 make -j1
 make test
-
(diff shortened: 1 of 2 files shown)


nim65s:
on gitlab-ci: https://gepgitlab.laas.fr/gsaurel/crocoddyl/pipelines/7832

cmastalli:
Thank for the nice work :)

I wasn't aware of that, I will work on it asap.

PR #663, opened 2020-01-09
Actuation squashing

PepMS:
This PR addresses the creation of a base class for ActuationModel with suqashing functions.


PepMS:
Yes, you are right. The squashing function does not depend on the actuation model. 
Also, you might want to test different squashing functions.
Right now I have in another branch where [I derived the squash base class](https://github.com/PepMS/crocoddyl/blob/uav_problem/src/multibody/actuations/uav-uam-squash.cpp).
As the squashing function depends on several parameters and I do not want them all scattered inside the derived actuation class, I've created a SquashingParams function.
However, the calculation of the squashing function value and the derivatives is done inside the Actuation class, so this not a very neat solution either.

I can implement the specific calculation and derivatives for the squashing function inside the ActuationSquashingAbstract class directly. How we are gonna deal with the parameters of the squashing function?

PepMS:
In [this branch](https://github.com/PepMS/crocoddyl/tree/actuation_squashing_derived_bind) I've implemented the Squashing function with an Abstract class and its associated Data structure, following the same philosophy as in the whole Crocoddyl. I personally like it better.

Besides, I've also done the corresponding bindings, which I needed in order to run experiments.
I've tried to separate the development of different new classes in different commits and branches.
So, in case you like it, it can be merged in several PR.

PepMS:
Closed this because it will be tackled in several PRs.

PR #664, opened 2020-01-09
Added getter and setter functions for solvers + improved Python bindings

cmastalli:
This PR:
 1. defines getter and setter functions which are needed for c++ implementation,
 2. changes copy_const_reference instead of return_by_value (Python binding returned policy), 
 3. returns the const block in Euler integrator, and
 4. fixes an issue in the Euler python bindings (wrong returning of data.dx)

@PepMS thanks for reporting the needed of these getter functions

### DIFF
--- a/src/core/solver-base.cpp
+++ b/src/core/solver-base.cpp
@@ -132,7 +132,7 @@ void SolverAbstract::set_ureg(const double& ureg) {
 }
 
 void SolverAbstract::set_th_acceptstep(const double& th_acceptstep) {
-  if (0. >= th_acceptstep && th_acceptstep >= 1) {
+  if (0. >= th_acceptstep || th_acceptstep > 1) {
     throw_pretty("Invalid argument: "
                  << "th_acceptstep value should between 0 and 1.");
   }
--- a/src/core/solver-base.cpp
+++ b/src/core/solver-base.cpp
@@ -91,7 +91,7 @@ const std::vector<Eigen::VectorXd>& SolverAbstract::get_xs() const { return xs_;
 
 const std::vector<Eigen::VectorXd>& SolverAbstract::get_us() const { return us_; }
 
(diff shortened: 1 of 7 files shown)


src/core/solver-base.cpp
```diff
@@ -105,12 +103,92 @@ const double& SolverAbstract::get_xreg() const { return xreg_; }
...
+
+const double& SolverAbstract::get_th_stop() const { return th_stop_; }
+
+const std::size_t& SolverAbstract::get_iter() const { return iter_; }
+
+void SolverAbstract::set_xs(const std::vector<Eigen::VectorXd>& xs) {
+  const std::size_t& T = problem_->get_T();
+  if (xs.size() != T + 1) {
+    throw_pretty("Invalid argument: "
+                 << "xs list has to be " + std::to_string(T + 1));
```
wxmerkt: I think you can do all of this with `<<` and avoid the `to_string` entirely.

src/core/solvers/ddp.cpp
```diff
@@ -438,15 +438,15 @@ void SolverDDP::set_alphas(const std::vector<double>& alphas) {
 }
 
 void SolverDDP::set_th_stepdec(const double& th_stepdec) {
-  if (0. >= th_stepdec && th_stepdec >= 1.) {
+  if (0. >= th_stepdec || th_stepdec > 1.) {
```
wxmerkt: Great catch!

src/core/solver-base.cpp
cmastalli: I think so. I agree with you that it would better to remove std:to_string and it header inclusion.

We have this type of comment in the entire library. I propose to solve it in an dedicated PR. I have opened an issue: https://github.com/loco-3d/crocoddyl/issues/665

PR #669, opened 2020-01-21
Fix bug, add unittest for floating-base DifferentialFwdDynamicsDerived

wxmerkt:
This PR fixes a couple issues when working with floating-base systems based on `DifferentialFwdDynamicsDerived` and adds a unit test for it.

- Fixes a bug related to `nq`,`nv` (swapped; difference when the floating base uses quaternion)
- Uses the correct `tau` vector to pass to ABA (the actuation->tau should be passed rather than the raw control vector)
- Adds a unit test with the Anymal robot using a floating base
- Fixes the arm_manipulation example visualisation
- Minor typos


cmastalli:
Thanks again!

PR #671, opened 2020-01-23
Change gap notation + improved diagnotic tools

cmastalli:
This PR proposes to improve the notation used for the gaps internally in the solvers, now it is renamed as 'fs' as in the Crocoddyl paper.

Additionally, it improves the naming of internal variables used by some diagnostics tools: logger and save logfile.

Finally, it improves the display of contact forces, friction cone and frame trajectories by showing always on top.


cmastalli:
@wxmerkt I have added documentation for all variables inside the solver classes.

I will later work on the entire documentation of classes.

PR #675, opened 2020-01-28
[CMake] drop multicontact-api dependency

nim65s:
Hi,

I think this dependency is useless.


cmastalli:
Yes, you're right!

Thanks for cleaning up the cmake code!

PR #683, opened 2020-02-07
[DAMFreefwddyn] remove redundant forwardKin call

proyan:
ForwardKin is not needed after aba or computeAllTerms

### DIFF
--- a/src/multibody/actions/free-fwddyn.cpp
+++ b/src/multibody/actions/free-fwddyn.cpp
@@ -57,6 +57,7 @@ void DifferentialActionModelFreeFwdDynamics::calc(const boost::shared_ptr<Differ
   // Computing the dynamics using ABA or manually for armature case
   if (with_armature_) {
     d->xout = pinocchio::aba(pinocchio_, d->pinocchio, q, v, d->multibody.actuation->tau);
+    pinocchio::updateGlobalPlacements(pinocchio_, d->pinocchio);
   } else {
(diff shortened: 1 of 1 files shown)


proyan:
Looks like I missed something.. lets wait.

proyan:
Got it. aba doesn't do oMi updates.

src/multibody/actions/free-fwddyn.cpp
```diff
@@ -57,6 +57,7 @@ void DifferentialActionModelFreeFwdDynamics::calc(const boost::shared_ptr<Differ
   // Computing the dynamics using ABA or manually for armature case
   if (with_armature_) {
     d->xout = pinocchio::aba(pinocchio_, d->pinocchio, q, v, d->multibody.actuation->tau);
+    pinocchio::updateGlobalPlacements(pinocchio_, d->pinocchio);
```
cmastalli: I double check the Pinocchio code, it seems `aba` doesn't compute `oMi`. Thus, it is OK from this point of view.

But `updateGlobalPlacements` doesn't update the spatial velocities (`data.v`).
And these quantities are needed to compute the frame-velocity cost.
Therefore, the solution is to substitute it by `forwardKinematics`.

proyan: aba updates data.v

cmastalli: Ok, let me check in my PC.
For me, I shouldn't appear any difference in the log files.
I will come back to this in few minutes.

Issue #672, opened 2020-01-24
Create a non-const reference to get_models in solver-base


proyan:
I would like to be able to update the models once they have been created inside the solver class. If you are running the MPC, it is counter-productive to re-define the entire problem again and again just for a small change in one feature.

If you are okay, I can push the changes immediately

wxmerkt:
I've asked for a similar API change before but was told that this requires to make sure that the data matches the updated models which would require further added checks. I.e., such a non-const change needs to include data checks in the calc/calcDiff calls.
An alternative would be to update the references in the costs etc. (which is what we currently do).

proyan:
>  An alternative would be to update the references in the costs etc. (which is what we currently do).

Could you elaborate please? Once the SolverModel object has been created, it has the CostModel and the cost references inside the action models. In order to update the references without re-creating the problem, you would need to access and modify the action models which are in SolverModel. How do you do that?

wxmerkt:
An example from our MPC code is e.g.:
```python
new_cref = ddp.problem.terminalModel.differential.costs.costs['comTrack'].cost.cref.copy()
new_cref[0] = 0.1
ddp.problem.terminalModel.differential.costs.costs['comTrack'].cost.cref = new_cref
```

proyan:
I see. I'm working in c++, and I think the only way around for me would be to save a pointer to the cost reference which is being shared, and update it this way.

I think I can make do, but I don't think this would be the right approach. This would still avoid the data checks that supposedly are preventing a proper solution. Moreover, this is limiting to specific problem types. For example if we have the contact planner in the loop, we might want to change the differential action model to add or remove contacts. Or perhaps change the DAM Model types.
I don't like this being used as a proper solution.

I am more in favor of having a data-checking function before we call solver.solve in debug mode. Otherwise If a user wants to update the model, he should be allowed to do that.

proyan:
Thanks for your solution @wxmerkt

cmastalli:
This is not the right way of doing it. If we need to change the model, then we have to interact with the shooting problem object directly.

However, I agree that we should have the following changes:
 - remove `get_models` and `get_datas` since it contains in the shooting problem
 - give freedom to set a different shooting problem

I will provide a solution soon!

**Additional comment**
I am currently working on a MPC formulation that will works for quadrupeds, bipeds, centaurs and humanoids. This implementation will handle all these details.

cmastalli:
I will close this issue, feel free to open it.

[...]

PR #684, opened 2020-02-07
Added setter functions + removed unnecessary computation of calc

cmastalli:
This PR proposes:
 - updates the example-robot-data version (and bipedal-ubound log) due to a fixed bug on this packages,
 - setter functions for the running / terminal models in shooting problem,
 - setter function for the differential action model in euler integrator,
 - `recalc` option for `calcDiff(data, x)` cases (action models),
 - `recalc` option for `calcDiff(xs, us)` in shooting problem, and
 - to avoid unnecessary extra computation of `calc` of each node

The latter point comes from the fact that we have already computed `calc` during the rollout. Thus, we don't call `calc` after the first iteration.

@proyan this is for you. I have tested the code against all log files + unittest code.


cmastalli:
This PR is blocked by the release of example-robot-data v3.1.2.
I will restart the CI when it has been finished

proyan:
Hey Carlos,
In this PR, you are doing a lot of things. It is hard to do a review.

The setter function is not what we need! What is needed for MPC is to be able to change one single parameter of the original problem, without having to redefine a new problem completely. These setters would be useless if you want to be efficient.

Simply add a non-const getter to the variables, as I initially asked for in #672 . Or make them unprotected.

cmastalli:
@proyan 

To be able to set cost parameters in c++, the user needs to downcast various classes.
This is a working example in c++:
```cpp
  // Downcast an action model to integrated action model
  crocoddyl::IntegratedActionModelEuler* m0 =
      static_cast<crocoddyl::IntegratedActionModelEuler*>(ddp.get_problem()->get_terminalModel().get());
  // Downcast an euler integrated action model to specific differential action model
  crocoddyl::DifferentialActionModelFreeFwdDynamics* m1 =
      static_cast<crocoddyl::DifferentialActionModelFreeFwdDynamics*>(m0->get_differential().get());
  // Retrieve the cost-base model
  const boost::shared_ptr<crocoddyl::CostModelAbstract>& c =
      m1->get_costs()->get_costs().find("gripperPose")->second.cost;
  // Downcast the cost abstraction to specific cost
  crocoddyl::CostModelFramePlacement* mycost = static_cast<crocoddyl::CostModelFramePlacement*>(c.get());
```

To make the c++ easier to write, the solution is to define as much as possible the objects and properties in the abstract class. For instance, we will need to define a cost model in the ActionModelAbstract class.
Despite that this is possible, we need to agree that we want to do that.
Unfortunately, other cases are not possible. For instance, we cannot define differential in the action-base class.

This solution will reduce the among of needed downcast. I could work on it, but in another PR. I need to modify many places in the code.

Therefore, I will proceed to merge this PR. And continuing discussing with you about next improvements.

include/crocoddyl/core/solvers/ddp.hpp
```diff
@@ -29,7 +29,7 @@ class SolverDDP : public SolverAbstract {
   virtual double tryStep(const double& steplength = 1);
   virtual double stoppingCriteria();
   virtual const Eigen::Vector2d& expectedImprovement();
-  virtual double calc();
+  virtual double calcDiff();
```
proyan: Hi Carlos,

Could we please avoid this type of arbitrary API changes? It breaks some of my code, and is a bit frustrating to find.

PR #685, opened 2020-02-07
[efficiency] remove redundant calls to update all frames. only update…

proyan:
… the ones that are required

This reduces the time used by calc (100 nodes) by about 100us.

### DIFF
--- a/src/multibody/actions/free-fwddyn.cpp
+++ b/src/multibody/actions/free-fwddyn.cpp
@@ -99,10 +99,8 @@ void DifferentialActionModelFreeFwdDynamics::calcDiff(const boost::shared_ptr<Di
 
   // Computing the dynamics derivatives
   if (with_armature_) {
-    pinocchio::computeABADerivatives(pinocchio_, d->pinocchio, q, v, d->multibody.actuation->tau);
-    d->Fx.leftCols(nv) = d->pinocchio.ddq_dq;
-    d->Fx.rightCols(nv) = d->pinocchio.ddq_dv;
-    d->Fx += d->pinocchio.Minv * d->multibody.actuation->dtau_dx;
+    pinocchio::computeABADerivatives(pinocchio_, d->pinocchio, q, v, d->multibody.actuation->tau, d->Fx.leftCols(nv), d->Fx.rightCols(nv), d->pinocchio.Minv);
+    d->Fx.noalias() += d->pinocchio.Minv * d->multibody.actuation->dtau_dx;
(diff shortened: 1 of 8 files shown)


src/multibody/actions/free-fwddyn.cpp
```diff
@@ -69,7 +69,6 @@ void DifferentialActionModelFreeFwdDynamics::calc(const boost::shared_ptr<Differ
   }
 
   // Computing the cost value and residuals
-  pinocchio::updateFramePlacements(pinocchio_, d->pinocchio);
```
cmastalli: Could we replicate this on `contact-fwddyn.cpp`?

proyan:
The failed builds are because of:
```
The following packages have unmet dependencies:

 robotpkg-py27-example-robot-data : Depends: robotpkg-example-robot-data (= 3.1.0) but 3.1.2 is to be installed or

                                             robotpkg-example-robot-data (= 3.0.0) but 3.1.2 is to be installed or

                                             robotpkg-example-robot-data (= 3.1.1) but 3.1.2 is to be installed

E: Unable to correct problems, you have held broken packages
```

@cmastalli Could you please have a look at the CI issues?

src/multibody/actions/free-fwddyn.cpp
proyan: I'm on the calcdiff right now :) , afterwards I'll look at contactDAM.

cmastalli: Great! thanks!
Do not forget to share the results.

proyan:
I checked the example logs. They are unchanged.

proyan:
@cmastalli I think this is ready. Could you please merge this?

cmastalli:
@proyan - It misses the changes in the contact-fwddyn. Otherwise, there is duplication of updating the frame placement.
I have pushed those changes, and after passing the CI, I will merge them.

Thank you for this nice contribution.

proyan:
I didn't want to put the contact fwddynamics in this PR, because I didn't want to put the time to be sure it doesn't break anything. This PR, I'm sure that it doesn't.

But if we are adding the contact-fwddynamics as well, we have to consider the costs dependent on this dynamics (contact-force and contact-friction-cone), and also propagate this change to impulse dynamics, impulses, and impulse based costs.

Don't merge this yet in that case. I'll make the changes and push.

proyan:
Did you check that it doesn't break with respect to the contact-force and contact-force friction cone costs?

cmastalli:
Yes, I checked it.

proyan:
Cool. Thanks.

PR #686, opened 2020-02-09
[bug] callback case 2

PepMS:
The reduction ocst where printed in the wrong order for the callbackverbose() case 2.


cmastalli:
Thanks a lot :)

PR #689, opened 2020-02-12
Remove unneccessary branchings from the code.

proyan:
In this PR:
1) I've removed the unnecessary calls to if(recalc) everywhere in the code. This reduces unnecessary branchings that were introduced, and reduces the dependence on the branch prediction of the architecture.

2) I've checked against the examples and the unittesting that the output is unchanged.

3) I've also removed an unnecessary call to computejointjacobians inside damfree.calcdiff

4) I've also updated a bug in the unittests.

No API is affected by this PR. However, after the PR is merged, calcdiff(data, x, u) would NOT update the calc at all!
Currently calcdiff(...,recalc) updates the calc by default. This default behaviour would be changed. Now if calc has to be called, it has to be called explicitly. The only place where we actually need to do a recalc is in the shootingproblem. So, we can just call a calc from there and not have to propagate the boolean everywhere inside the code.


proyan:
@nim65s Could we please wait to include this PR in the 1.1.0 release?

proyan:
Finally, 

When we redefined the asserts in https://github.com/loco-3d/crocoddyl/blob/devel/include/crocoddyl/core/utils/exception.hpp#L29, unfortunately we overrode the implementation of asserts which depends on NDEBUG. 

The last commit now makes it dependent on NDEBUG, and if NDEBUG is defined, asserting does nothing.

include/crocoddyl/core/utils/exception.hpp
```diff
@@ -26,13 +26,16 @@
     throw crocoddyl::Exception(ss.str(), __FILE__, __PRETTY_FUNCTION__, __LINE__); \
   }
 
+#ifdef NDEBUG
+#define assert_pretty(condition, m) ((void)0)
```
wxmerkt: Please do not add and merge this to master. Have you been able to see speed-ups with this line added, and if so, how much? The risk of someone developing code in `RelWithDebInfo` or another mode where `NDEBUG` may be added as a flag is too high.

proyan: Is this a policy in crocoddyl to keep asserts active ALL the time? If so, I didn't know that.

I noticed a high number of cache data misses in the asserts when analyzing the code. Normally the asserts are disabled in release mode, are we not following the same principle here?

wxmerkt: > I noticed a high number of cache data misses in the asserts when analyzing the code.  
What's the bottom-line impact here concretely?

Often code is in Release or RelWithDebInfo or from the binaries (!) when developing or debugging problem formulations e.g. using Python or dependent code. We had _weeks_ of time lost due to slight bugs or argument errors which did not show or notice. If this has a noticeable effect for one application, it can be activated using a compile-time switch locally but it should not be the default.

cmastalli: We  want to keep the assert in release mode as @wxmerkt explained.

cmastalli:
> @nim65s Could we please wait to include this PR in the 1.1.0 release?

@proyan I disagree. This modifications are not urgent.
The PR is huge and it will take time to review it entirely.

include/crocoddyl/core/utils/exception.hpp
proyan: I'm okay with that, I thought this was an error since `<assert.h>` has a different policy.
I'll revert this commit.
Thanks for the feedback @wxmerkt.

For a problem of 4DoF robot, with 100 nodes and no multithreading, I saw an benefit of ~10us. I'm using Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz,

proyan:
The modifications are not urgent, but there is a change in the default behaviour of calcDiff. Since you are doing a major version bump from 1.0.0 to 1.1.0, I think this change in behaviour should be included in a major increase. In that case, 1.0.0 would be safer for the existing code.

But I'll let you and @nim65s handle the releases.

cmastalli:
> The modifications are not urgent, but there is a change in the default behaviour of calcDiff. Since you are doing a major version bump from 1.0.0 to 1.1.0, I think this change in behaviour should be included in a major increase. In that case, 1.0.0 would be safer for the existing code.
> 
> But I'll let you and @nim65s handle the releases.

I understand, but there is already people building models using the devel-branch API.
 A quick merge it wouldn't prevent this inconvenient for our users.

The good thing is that it shouldn't break any user Python code, but it is not the same for c++ case.

src/core/action-base.cpp
```diff
@@ -29,8 +29,8 @@ void ActionModelAbstract::calc(const boost::shared_ptr<ActionDataAbstract>& data
 }
 
 void ActionModelAbstract::calcDiff(const boost::shared_ptr<ActionDataAbstract>& data,
-                                   const Eigen::Ref<const Eigen::VectorXd>& x, const bool& recalc) {
-  calcDiff(data, x, unone_, recalc);
+                                   const Eigen::Ref<const Eigen::VectorXd>& x, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code.

src/core/actions/diff-lqr.cpp
```diff
@@ -52,7 +52,7 @@ void DifferentialActionModelLQR::calc(const boost::shared_ptr<DifferentialAction
 
 void DifferentialActionModelLQR::calcDiff(const boost::shared_ptr<DifferentialActionDataAbstract>& data,
                                           const Eigen::Ref<const Eigen::VectorXd>& x,
-                                          const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
+                                          const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/actions/lqr.cpp
```diff
@@ -47,7 +47,7 @@ void ActionModelLQR::calc(const boost::shared_ptr<ActionDataAbstract>& data,
 
 void ActionModelLQR::calcDiff(const boost::shared_ptr<ActionDataAbstract>& data,
                               const Eigen::Ref<const Eigen::VectorXd>& x, const Eigen::Ref<const Eigen::VectorXd>& u,
-                              const bool& recalc) {
+                              const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/actions/unicycle.cpp
```diff
@@ -40,7 +40,7 @@ void ActionModelUnicycle::calc(const boost::shared_ptr<ActionDataAbstract>& data
 
 void ActionModelUnicycle::calcDiff(const boost::shared_ptr<ActionDataAbstract>& data,
                                    const Eigen::Ref<const Eigen::VectorXd>& x,
-                                   const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
+                                   const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/activations/quadratic.cpp
```diff
@@ -25,14 +25,12 @@ void ActivationModelQuad::calc(const boost::shared_ptr<ActivationDataAbstract>&
 }
 
 void ActivationModelQuad::calcDiff(const boost::shared_ptr<ActivationDataAbstract>& data,
-                                   const Eigen::Ref<const Eigen::VectorXd>& r, const bool& recalc) {
+                                   const Eigen::Ref<const Eigen::VectorXd>& r, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/activations/smooth-abs.cpp
```diff
@@ -28,14 +28,11 @@ void ActivationModelSmoothAbs::calc(const boost::shared_ptr<ActivationDataAbstra
 }
 
 void ActivationModelSmoothAbs::calcDiff(const boost::shared_ptr<ActivationDataAbstract>& data,
-                                        const Eigen::Ref<const Eigen::VectorXd>& r, const bool& recalc) {
+                                        const Eigen::Ref<const Eigen::VectorXd>& r, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/activations/weighted-quadratic-barrier.cpp
```diff
@@ -34,15 +34,11 @@ void ActivationModelWeightedQuadraticBarrier::calc(const boost::shared_ptr<Activ
 }
 
 void ActivationModelWeightedQuadraticBarrier::calcDiff(const boost::shared_ptr<ActivationDataAbstract>& data,
-                                                       const Eigen::Ref<const Eigen::VectorXd>& r,
-                                                       const bool& recalc) {
+                                                       const Eigen::Ref<const Eigen::VectorXd>& r, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/activations/weighted-quadratic.cpp
```diff
@@ -29,14 +29,11 @@ void ActivationModelWeightedQuad::calc(const boost::shared_ptr<ActivationDataAbs
 }
 
 void ActivationModelWeightedQuad::calcDiff(const boost::shared_ptr<ActivationDataAbstract>& data,
-                                           const Eigen::Ref<const Eigen::VectorXd>& r, const bool& recalc) {
+                                           const Eigen::Ref<const Eigen::VectorXd>& r, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/diff-action-base.cpp
```diff
@@ -29,8 +29,8 @@ void DifferentialActionModelAbstract::calc(const boost::shared_ptr<DifferentialA
 }
 
 void DifferentialActionModelAbstract::calcDiff(const boost::shared_ptr<DifferentialActionDataAbstract>& data,
-                                               const Eigen::Ref<const Eigen::VectorXd>& x, const bool& recalc) {
-  calcDiff(data, x, unone_, recalc);
+                                               const Eigen::Ref<const Eigen::VectorXd>& x, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/activations/quadratic-barrier.cpp
```diff
@@ -31,14 +31,11 @@ void ActivationModelQuadraticBarrier::calc(const boost::shared_ptr<ActivationDat
 }
 
 void ActivationModelQuadraticBarrier::calcDiff(const boost::shared_ptr<ActivationDataAbstract>& data,
-                                               const Eigen::Ref<const Eigen::VectorXd>& r, const bool& recalc) {
+                                               const Eigen::Ref<const Eigen::VectorXd>& r, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/integrator/euler.cpp
```diff
@@ -74,7 +74,7 @@ void IntegratedActionModelEuler::calc(const boost::shared_ptr<ActionDataAbstract
 
 void IntegratedActionModelEuler::calcDiff(const boost::shared_ptr<ActionDataAbstract>& data,
                                           const Eigen::Ref<const Eigen::VectorXd>& x,
-                                          const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
+                                          const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/numdiff/action.cpp
```diff
@@ -37,7 +37,7 @@ void ActionModelNumDiff::calc(const boost::shared_ptr<ActionDataAbstract>& data,
 
 void ActionModelNumDiff::calcDiff(const boost::shared_ptr<ActionDataAbstract>& data,
                                   const Eigen::Ref<const Eigen::VectorXd>& x,
-                                  const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
+                                  const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/numdiff/activation.cpp
```diff
@@ -31,16 +31,13 @@ void ActivationModelNumDiff::calc(const boost::shared_ptr<ActivationDataAbstract
 }
 
 void ActivationModelNumDiff::calcDiff(const boost::shared_ptr<ActivationDataAbstract>& data,
-                                      const Eigen::Ref<const Eigen::VectorXd>& r, const bool& recalc) {
+                                      const Eigen::Ref<const Eigen::VectorXd>& r, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/core/numdiff/diff-action.cpp
```diff
@@ -41,7 +41,7 @@ void DifferentialActionModelNumDiff::calc(const boost::shared_ptr<DifferentialAc
 
 void DifferentialActionModelNumDiff::calcDiff(const boost::shared_ptr<DifferentialActionDataAbstract>& data,
                                               const Eigen::Ref<const Eigen::VectorXd>& x,
-                                              const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
+                                              const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/actions/contact-fwddyn.cpp
```diff
@@ -97,8 +97,7 @@ void DifferentialActionModelContactFwdDynamics::calc(const boost::shared_ptr<Dif
 
 void DifferentialActionModelContactFwdDynamics::calcDiff(const boost::shared_ptr<DifferentialActionDataAbstract>& data,
                                                          const Eigen::Ref<const Eigen::VectorXd>& x,
-                                                         const Eigen::Ref<const Eigen::VectorXd>& u,
-                                                         const bool& recalc) {
+                                                         const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/actions/free-fwddyn.cpp
```diff
@@ -75,7 +75,7 @@ void DifferentialActionModelFreeFwdDynamics::calc(const boost::shared_ptr<Differ
 
 void DifferentialActionModelFreeFwdDynamics::calcDiff(const boost::shared_ptr<DifferentialActionDataAbstract>& data,
                                                       const Eigen::Ref<const Eigen::VectorXd>& x,
-                                                      const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
+                                                      const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/actions/free-fwddyn.cpp
```diff
@@ -90,12 +90,8 @@ void DifferentialActionModelFreeFwdDynamics::calcDiff(const boost::shared_ptr<Di
   const Eigen::VectorBlock<const Eigen::Ref<const Eigen::VectorXd>, Eigen::Dynamic> v = x.tail(nv);
 
   DifferentialActionDataFreeFwdDynamics* d = static_cast<DifferentialActionDataFreeFwdDynamics*>(data.get());
-  if (recalc) {
-    calc(data, x, u);
-    pinocchio::computeJointJacobians(pinocchio_, d->pinocchio, q);
```
cmastalli: I am not sure if removing `computeJoinJacobians` have an implication. It seems the frame costs require this computation for getting the frame Jacobians.

Did you check the log files?

src/multibody/actions/impulse-fwddyn.cpp
```diff
@@ -91,7 +91,7 @@ void ActionModelImpulseFwdDynamics::calc(const boost::shared_ptr<ActionDataAbstr
 
 void ActionModelImpulseFwdDynamics::calcDiff(const boost::shared_ptr<ActionDataAbstract>& data,
                                              const Eigen::Ref<const Eigen::VectorXd>& x,
-                                             const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
+                                             const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/actuations/floating-base.cpp
```diff
@@ -34,10 +34,7 @@ void ActuationModelFloatingBase::calc(const boost::shared_ptr<ActuationDataAbstr
 
 void ActuationModelFloatingBase::calcDiff(const boost::shared_ptr<ActuationDataAbstract>& data,
                                           const Eigen::Ref<const Eigen::VectorXd>& x,
-                                          const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
+                                          const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/actuations/full.cpp
```diff
@@ -33,10 +33,7 @@ void ActuationModelFull::calc(const boost::shared_ptr<ActuationDataAbstract>& da
 
 void ActuationModelFull::calcDiff(const boost::shared_ptr<ActuationDataAbstract>& data,
                                   const Eigen::Ref<const Eigen::VectorXd>& x,
-                                  const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
+                                  const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/contacts/contact-3d.cpp
```diff
@@ -46,11 +46,7 @@ void ContactModel3D::calc(const boost::shared_ptr<ContactDataAbstract>& data,
 }
 
 void ContactModel3D::calcDiff(const boost::shared_ptr<ContactDataAbstract>& data,
-                              const Eigen::Ref<const Eigen::VectorXd>& x, const bool& recalc) {
-  if (recalc) {
-    calc(data, x);
-  }
-
+                              const Eigen::Ref<const Eigen::VectorXd>& x, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/contacts/contact-6d.cpp
```diff
@@ -44,11 +44,7 @@ void ContactModel6D::calc(const boost::shared_ptr<ContactDataAbstract>& data,
 }
 
 void ContactModel6D::calcDiff(const boost::shared_ptr<ContactDataAbstract>& data,
-                              const Eigen::Ref<const Eigen::VectorXd>& x, const bool& recalc) {
-  if (recalc) {
-    calc(data, x);
-  }
-
+                              const Eigen::Ref<const Eigen::VectorXd>& x, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/contacts/multiple-contacts.cpp
```diff
@@ -69,14 +69,11 @@ void ContactModelMultiple::calc(const boost::shared_ptr<ContactDataMultiple>& da
 }
 
 void ContactModelMultiple::calcDiff(const boost::shared_ptr<ContactDataMultiple>& data,
-                                    const Eigen::Ref<const Eigen::VectorXd>& x, const bool& recalc) {
+                                    const Eigen::Ref<const Eigen::VectorXd>& x, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/centroidal-momentum.cpp
```diff
@@ -54,17 +54,13 @@ void CostModelCentroidalMomentum::calc(const boost::shared_ptr<CostDataAbstract>
 
 void CostModelCentroidalMomentum::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                                            const Eigen::Ref<const Eigen::VectorXd>& x,
-                                           const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
-
+                                           const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/com-position.cpp
```diff
@@ -53,16 +53,12 @@ void CostModelCoMPosition::calc(const boost::shared_ptr<CostDataAbstract>& data,
 
 void CostModelCoMPosition::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                                     const Eigen::Ref<const Eigen::VectorXd>& x,
-                                    const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
-
+                                    const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/contact-force.cpp
```diff
@@ -55,17 +55,13 @@ void CostModelContactForce::calc(const boost::shared_ptr<CostDataAbstract>& data
 
 void CostModelContactForce::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                                      const Eigen::Ref<const Eigen::VectorXd>& x,
-                                     const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
-
+                                     const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/contact-friction-cone.cpp
```diff
@@ -59,18 +59,14 @@ void CostModelContactFrictionCone::calc(const boost::shared_ptr<CostDataAbstract
 
 void CostModelContactFrictionCone::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                                             const Eigen::Ref<const Eigen::VectorXd>& x,
-                                            const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
-
+                                            const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/control.cpp
```diff
@@ -63,7 +63,7 @@ void CostModelControl::calc(const boost::shared_ptr<CostDataAbstract>& data, con
 
 void CostModelControl::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                                 const Eigen::Ref<const Eigen::VectorXd>& x, const Eigen::Ref<const Eigen::VectorXd>& u,
-                                const bool& recalc) {
+                                const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/cost-sum.cpp
```diff
@@ -78,7 +78,7 @@ void CostModelSum::calc(const boost::shared_ptr<CostDataSum>& data, const Eigen:
 }
 
 void CostModelSum::calcDiff(const boost::shared_ptr<CostDataSum>& data, const Eigen::Ref<const Eigen::VectorXd>& x,
-                            const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
+                            const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/frame-placement.cpp
```diff
@@ -59,10 +59,7 @@ void CostModelFramePlacement::calc(const boost::shared_ptr<CostDataAbstract>& da
 
 void CostModelFramePlacement::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                                        const Eigen::Ref<const Eigen::VectorXd>& x,
-                                       const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
+                                       const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/frame-rotation.cpp
```diff
@@ -58,10 +58,7 @@ void CostModelFrameRotation::calc(const boost::shared_ptr<CostDataAbstract>& dat
 
 void CostModelFrameRotation::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                                       const Eigen::Ref<const Eigen::VectorXd>& x,
-                                      const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
+                                      const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/frame-translation.cpp
```diff
@@ -57,10 +57,7 @@ void CostModelFrameTranslation::calc(const boost::shared_ptr<CostDataAbstract>&
 
 void CostModelFrameTranslation::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                                          const Eigen::Ref<const Eigen::VectorXd>& x,
-                                         const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
+                                         const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/frame-velocity.cpp
```diff
@@ -57,19 +57,15 @@ void CostModelFrameVelocity::calc(const boost::shared_ptr<CostDataAbstract>& dat
 
 void CostModelFrameVelocity::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                                       const Eigen::Ref<const Eigen::VectorXd>& x,
-                                      const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
-
+                                      const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/impulse-com.cpp
```diff
@@ -46,17 +46,13 @@ void CostModelImpulseCoM::calc(const boost::shared_ptr<CostDataAbstract>& data,
 
 void CostModelImpulseCoM::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                                    const Eigen::Ref<const Eigen::VectorXd>& x,
-                                   const Eigen::Ref<const Eigen::VectorXd>& u, const bool& recalc) {
-  if (recalc) {
-    calc(data, x, u);
-  }
-
+                                   const Eigen::Ref<const Eigen::VectorXd>& u, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/costs/state.cpp
```diff
@@ -129,18 +129,15 @@ void CostModelState::calc(const boost::shared_ptr<CostDataAbstract>& data, const
 
 void CostModelState::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
                               const Eigen::Ref<const Eigen::VectorXd>& x, const Eigen::Ref<const Eigen::VectorXd>& u,
-                              const bool& recalc) {
+                              const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/impulses/impulse-3d.cpp
```diff
@@ -27,11 +27,7 @@ void ImpulseModel3D::calc(const boost::shared_ptr<ImpulseDataAbstract>& data,
 }
 
 void ImpulseModel3D::calcDiff(const boost::shared_ptr<ImpulseDataAbstract>& data,
-                              const Eigen::Ref<const Eigen::VectorXd>& x, const bool& recalc) {
-  if (recalc) {
-    calc(data, x);
-  }
-
+                              const Eigen::Ref<const Eigen::VectorXd>& x, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/impulses/impulse-6d.cpp
```diff
@@ -26,11 +26,7 @@ void ImpulseModel6D::calc(const boost::shared_ptr<ImpulseDataAbstract>& data,
 }
 
 void ImpulseModel6D::calcDiff(const boost::shared_ptr<ImpulseDataAbstract>& data,
-                              const Eigen::Ref<const Eigen::VectorXd>& x, const bool& recalc) {
-  if (recalc) {
-    calc(data, x);
-  }
-
+                              const Eigen::Ref<const Eigen::VectorXd>& x, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

src/multibody/impulses/multiple-impulses.cpp
```diff
@@ -60,14 +60,11 @@ void ImpulseModelMultiple::calc(const boost::shared_ptr<ImpulseDataMultiple>& da
 }
 
 void ImpulseModelMultiple::calcDiff(const boost::shared_ptr<ImpulseDataMultiple>& data,
-                                    const Eigen::Ref<const Eigen::VectorXd>& x, const bool& recalc) {
+                                    const Eigen::Ref<const Eigen::VectorXd>& x, const bool&) {
```
cmastalli: Remove unnecessary `const bool&` and update Python bindings code

unittest/bindings/test_shooting.py
```diff
@@ -41,8 +41,8 @@ def test_calc(self):
 
     def test_calcDiff(self):
         # Running calc functions
-        cost = self.PROBLEM.calcDiff(self.xs, self.us)
-        costDer = self.PROBLEM_DER.calcDiff(self.xs, self.us)
+        cost = self.PROBLEM.calcDiff(self.xs, self.us, True)
```
cmastalli: Remove `recalc` option. Let's follow the same patter for the shooting problem too.
Call `calc` first

src/core/optctrl/shooting.cpp
```diff
@@ -66,23 +68,36 @@ double ShootingProblem::calcDiff(const std::vector<Eigen::VectorXd>& xs, const s
...
   }
 
-  cost_ = 0;
   std::size_t i;
 
 #ifdef WITH_MULTITHREADING
   omp_set_num_threads(NUM_THREADS);
+#endif
+
+  if (recalc) {
```
cmastalli: Remove the `recalc` option. Let's follow the same patter to this class too.
Move the `calc` parallelization to the `ShootingProblem::calc`.

unittest/bindings/testutils.py
```diff
@@ -0,0 +1,60 @@
+import numpy as np
```
cmastalli: Remove this file. These functions are in `test_utils.py`, excluding `df_dx` and `df_dq` because they are not needed.

bindings/python/crocoddyl/utils/__init__.py
```diff
@@ -307,14 +307,15 @@ def calcDiff(self, data, x, u=None, recalc=True):
         self.costsData.shareMemory(data)
         nq, nv = self.state.nq, self.state.nv
         q, v = x[:nq], x[-nv:]
-        self.actuation.calcDiff(self.actuationData, x, u)
-        tau = self.actuationData.tau
 
         if u is None:
             u = self.unone
         if recalc:
```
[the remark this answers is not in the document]
proyan: computejointjacobian was not needed. The rest is done.

unittest/bindings/test_shooting.py
proyan: Let's discuss this part separately. There is a faulty logic in the defintions of the derived classes in python.

src/multibody/actions/free-fwddyn.cpp
proyan: Yes I did.

src/core/optctrl/shooting.cpp
proyan: This recalc actually affects the code. And I need to think on how to do this properly, since this is the part being called in the solvers.
I'll handle it in a separate PR. I don't want to push this change with this big PR.

proyan:
@nim65s The CI is failing with the following error:

```
 File "/opt/openrobots/lib/python2.7/site-packages/example_robot_data/robots_loader.py", line 4, in <module>

    import numpy as np

ImportError: No module named 'numpy'
```

Is this a example-robot-data issue?

nim65s:
@proyan : you might consider configuring your text editor to format automatically your code when you save files, instead of adding many reformat commits :)

wxmerkt:
Regarding the CI failure, apparently the `robotpkg` packages no longer depend on `python-numpy` (?) so it won't be automatically installed on CI. Has there been a change to the debian dependency keys?

nim65s:
@wxmerkt : `apt depends robotpkg-py36-crocoddyl` shows `robotpkg-py36-eigenpy`, and `apt depends robotpkg-py36-eigenpy` shows `python3-numpy`, so the dependency is still here, isn't it ?

wxmerkt:
@nim65s Does this also apply to the py27 version? (I don't have robotpkg on my current workstation) Here is what is installed on CI:
```bash
sudo apt install -y -qq libeigen3-dev doxygen robotpkg-py27-eigenpy robotpkg-py27-pinocchio robotpkg-py27-example-robot-data python-scipy robotpkg-py27-quadprog
```

nim65s:
@wxmerkt : yep:

```bash
╰─>$ apt depends robotpkg-py27-crocoddyl | grep eigenpy
  Dépend: robotpkg-py27-eigenpy (= 1.6.12)
╰─>$ apt depends robotpkg-py27-eigenpy | grep numpy
  Dépend: python-numpy (>= 1:1)
```

unittest/bindings/test_shooting.py
cmastalli: Are you suggesting to do not remove `recalc` for shooting problem?

cmastalli:
> @nim65s Does this also apply to the py27 version? (I don't have robotpkg on my current workstation) Here is what is installed on CI:
> 
> ```shell
> sudo apt install -y -qq libeigen3-dev doxygen robotpkg-py27-eigenpy robotpkg-py27-pinocchio robotpkg-py27-example-robot-data python-scipy robotpkg-py27-quadprog
> ```

I also noticed that scipy is not found during compilation.
@nim65s @wxmerkt what are the possible solutions to it?

On the other hand, we should start to test our code against Python 3.

unittest/bindings/test_shooting.py
proyan: Already removed it. There is no recalc anymore.

proyan:
This PR is ready to merge if the CI passes.

There is no recalc anymore in DDP, so calcDiff is completely independent from calc. This means that moving forward, if you wish to obtain the derivatives via calcDiff, you would need to explicitly call the calc function. This might break some existing code, though it won't affect the user specific API for solver and ShootingProblem classes.

wxmerkt:
I dug deeper in the CI failure and the reason is that the CMake modules is now finding and configuring `python3` by default (when it should not on LTS systems!). Of course `python-numpy` installs the python2 version. So this issue was inherited from the CMake module upstream.

I thought the agreement was not to change default behaviour until the python interpreter changes by default on a LTS platform?

(And making the CI python3 is not a solution - the default CMake behaviour should not have changed)

cmastalli:
@proyan I have cleaned up the code:  removed warnings and updated doc.
Additionally, I have updated the quadrotor script (only missed part).

Everything is OK to be merged. I will proceed with it, when the CI passes.

Thanks for this great effort.

Issue #693, opened 2020-02-17
Assertion from pinocchio explog triggered in examples


wxmerkt:
I am on 18.04 and  have the latest Pinocchio devel compiled with `CMAKE_BUILD_TYPE=MinSizeRel -Os` and Crocoddyl compiled with `CMAKE_BUILD_TYPE=RelWIthDebInfo -O3`. Running any of the examples (e.g. `boxddp_vs_boxfddp` or `quadrupedal_walk_ubound`) results in the following assertion:

> include/pinocchio/spatial/explog.hpp:94: Eigen::Matrix<typename Matrix3Like::Scalar, 3, 1, typename Eigen::internal::plain_matrix_type<typename pinocchio::helper::argument_type<void(D)>::type, typename Eigen::internal::traits<typename pinocchio::helper::argument_type<void(D)>::type>::StorageKind>::type:: Options> pinocchio::log3(const Eigen::MatrixBase<Derived>&, typename Matrix3Like::Scalar&) [with Matrix3Like = Eigen::Matrix<double, 3, 3>; typename Eigen::internal::plain_matrix_type<typename pinocchio::helper::argument_type<void(D)>::type, typename Eigen::internal::traits<typename pinocchio::helper::argument_type<void(D)>::type>::StorageKind>::type = Eigen::Matrix<double, 3, 3>; typename Matrix3Like::Scalar = double]: Assertion `theta == theta && "theta contains some NaN"' failed.

This is on an Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz which has been freshly installed. I did not see these issues on my previous configurations (i7-8565U or i9-9900K) - albeit with older commits for both Pinocchio and Crocoddyl back then. 

Have you seen these before or any insights why these might be triggered?

wxmerkt:
I cannot reproduce this on another machine (i7-8565U) that should be set up in the same way. Any ideas why this could happen though? I saw it a few times in the past. Clean builds did not resolve it.

cmastalli:
I found the problem. I will make a PR that solves it.
You will see more details there.

cmastalli:
Please see the PR.

wxmerkt:
Thank you for the quick turn-around and fix :-) I did not expect it to be in the `quadruped.py`.

cmastalli:
> Thank you for the quick turn-around and fix :-) I did not expect it to be in the `quadruped.py`.

You're welcome, and thank you for the report.
[...]

PR #694, opened 2020-02-18
Malformation of quaternon in quadrupedal formulation

cmastalli:
This error appears only for Debug mode, since the quaternion malformation actives a Pinocchio assert (in `pinocchio::diff`).

It solves the issue #693.

### DIFF
--- a/bindings/python/crocoddyl/utils/quadruped.py
+++ b/bindings/python/crocoddyl/utils/quadruped.py
@@ -447,8 +447,8 @@ class SimpleQuadrupedalGaitProblem:
         costModel.addCost("stateReg", stateReg, 1e1)
         costModel.addCost("ctrlReg", ctrlReg, 1e-1)
 
-        lb = self.state.diff(0 * self.state.lb, self.state.lb)
-        ub = self.state.diff(0 * self.state.ub, self.state.ub)
+        lb = np.vstack([ self.state.lb[1:self.state.nv+1], self.state.lb[-self.state.nv:] ])
+        ub = np.vstack([ self.state.ub[1:self.state.nv+1], self.state.ub[-self.state.nv:] ])
         stateBounds = crocoddyl.CostModelState(
             self.state, crocoddyl.ActivationModelQuadraticBarrier(crocoddyl.ActivationBounds(lb, ub)),
             0 * self.rmodel.defaultState, self.actuation.nu)
--- a/bindings/python/crocoddyl/utils/quadruped.py
+++ b/bindings/python/crocoddyl/utils/quadruped.py
@@ -447,8 +447,8 @@ class SimpleQuadrupedalGaitProblem:
         costModel.addCost("stateReg", stateReg, 1e1)
         costModel.addCost("ctrlReg", ctrlReg, 1e-1)
 
-        lb = np.vstack([ self.state.lb[1:self.state.nv+1], self.state.lb[-self.state.nv:] ])
-        ub = np.vstack([ self.state.ub[1:self.state.nv+1], self.state.ub[-self.state.nv:] ])
+        lb = np.vstack([self.state.lb[1:self.state.nv + 1], self.state.lb[-self.state.nv:]])
+        ub = np.vstack([self.state.ub[1:self.state.nv + 1], self.state.ub[-self.state.nv:]])
         stateBounds = crocoddyl.CostModelState(
             self.state, crocoddyl.ActivationModelQuadraticBarrier(crocoddyl.ActivationBounds(lb, ub)),
             0 * self.rmodel.defaultState, self.actuation.nu)


cmastalli:
@nim65s @wxmerkt 
The CI is broken, this is the message:

```
The following packages have unmet dependencies:

 robotpkg-py27-example-robot-data : Depends: robotpkg-py27-eigenpy (= 1.6.12) but 2.0.2 is to be installed

                                    Depends: robotpkg-hpp-fcl (= 1.2.1) or

                                             robotpkg-hpp-fcl+doc (= 1.2.1) but it is not going to be installed or

                                             robotpkg-hpp-fcl (= 1.2.2) or

                                             robotpkg-hpp-fcl+doc (= 1.2.2) but it is not going to be installed

 robotpkg-py27-pinocchio : Depends: robotpkg-py27-hpp-fcl (= 1.2.1) or

                                    robotpkg-py27-hpp-fcl (= 1.2.0) or

                                    robotpkg-py27-hpp-fcl (= 1.2.2) or

                                    robotpkg-py27-hpp-fcl+doc (= 1.2.0) but it is not going to be installed

                           Depends: robotpkg-py27-eigenpy (= 1.6.12) but 2.0.2 is to be installed

                           Depends: robotpkg-hpp-fcl (= 1.2.1) or

                                    robotpkg-hpp-fcl+doc (= 1.2.1) but it is not going to be installed or

                                    robotpkg-hpp-fcl (= 1.2.2) or

                                    robotpkg-hpp-fcl+doc (= 1.2.2) but it is not going to be installed

E: Unable to correct problems, you have held broken packages.

The command "/bin/bash .travis/install-prerequisites.sh" failed and exited with 100 during .
```

do you know the reason?

nim65s:
An upgrade of hpp-fcl & eigenpy is currently running on robotpkg buildfarm.

ref:
https://git.openrobots.org/issues/214
https://github.com/stack-of-tasks/pinocchio/issues/884

cmastalli:
> An upgrade of hpp-fcl & eigenpy is currently running on robotpkg buildfarm.
> 
> ref:
> https://git.openrobots.org/issues/214
> [stack-of-tasks/pinocchio#884](https://github.com/stack-of-tasks/pinocchio/issues/884)
> 
> Please just wait a bit for http://robotpkg.openrobots.org/rbulk/rbulk-dispatchd.html to complete and start again :)

Thanks for the info :)

wxmerkt:
Is there an ETA of when robotpkg will enable staging? This happens pretty frequently for us these days :-(. Otherwise happy to move to ROS Industrial CI for build + unit tests

nim65s:
There is no ETA that I know of. 

This is way too far in robotpkg's internals for me, and the main developer did not answer on the issue yet.

wxmerkt:
Actually, rather than a band-aid added to the problem formulation for the quadruped, shouldn't this be fixed inside the `state.diff`? Or in other places in Crocoddyl as well?

cmastalli:
> Actually, rather than a band-aid added to the problem formulation for the quadruped, shouldn't this be fixed inside the `state.diff`? Or in other places in Crocoddyl as well?

It is not a bug in the `diff` function (or `pinocchio::difference`). If the user passes an unsymmetric rotation matrix to a hypothetical function then he/she is making a wrong mistake. This is exactly what it was happening but with a quaternion.

A common software engineering solution is to include an assert to your code, and we follow this approach.

A proper way to handle this issue will be to devise an API that cannot allow users to construct such kind of problem. However, this is not trivial as we might have system without floating base, or states that can be anything else. Additionally it could unnecessary work since asserts are there, and human resources are very limited.

cmastalli:
@nim65s thanks for the support. I understand the limitations that we have with robotpkg.
However, we could create a team policy to mitigate this inconvenient as suggested in riot.

I am at your disposal for any further discussion in the riot channel.

cmastalli:
> Is there an ETA of when robotpkg will enable staging? This happens pretty frequently for us these days :-(. Otherwise happy to move to ROS Industrial CI for build + unit tests

Could we support both?

wxmerkt:
I wasn't suggesting it's a bug but in the input arguments - and sadly easy enough to make and miss when Pinocchio is in release mode (or from binaries).

Perhaps adding a check and exception (only to check if the multibody-state has a free-flyer, and to check if it has a normalised quaternion) is a workable solution? We should try to protect users from these errors when experimenting and formulating problems (with Pinocchio in release mode they'd be almost impossible to debug "why the DDP doesn't converge/exhibits failure mode XYZ").

We could support both CI. I won't have time for that before IROS though.

Sorry I may miss the point here. What team policy could mitigate an issue that is outwith our control as Guilhem mentioned as it is fully upstream with robotpkg?

nim65s:
@wxmerkt : my CI system on gepgitlab is based on regularly building docker images with everything already automatically installed. This mean that at runtime, we don't have to waste time installing anything, which implies not having errors in the CI in case of network issues or this kind of unfinished job in the robotpkg's buildfarm. 

This system will be automatically enabled on this repo in the week.

Now, if you want to build a fully independent CI system on Travis, I think your suggestion to use builds from the ROS buildfarm should do the trick. But please note that I have no bandwidth to support the 2 systems.

wxmerkt:
> This system will be automatically enabled on this repo in the week.

That's great! Can you please add some detail in an issue? I was not aware of this development

wxmerkt:
And - I think this CI should be additional and not replace the Travis (which should stay required for a PR to be merged) as it's testing one of the userflows that is recommended in the README

nim65s:
It will be additional, don't worry :)

About the details, it will be available in another repo, as the rest of my CI system works with all our packages at once, and there is nothing particular for crocoddyl.

cmastalli:
> I wasn't suggesting it's a bug but in the input arguments - and sadly easy enough to make and miss when Pinocchio is in release mode (or from binaries).

I agree with you. My concerns are always about the balance between user-friendly vs high-performance software. I am happy to adapt any solution that does not affect the later. Give some time for thinking about it :)

PR #696, opened 2020-02-22
Big difference in examples log for Box ddp solver

proyan:
When comparing the logs to make sure that modifications don't produce a divergence in the output of DDP, it seems normal to expect some difference in numerical outputs because of floating point operations. However, when comparing the logs for the current devel in my system, there is a big difference in the cost outputs of the BoxDDP, which should not be there.

My system is `Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz`, and I have `Ubuntu 16.04.6 LTS` and all the system dependencies that come with it.

This PR is not to be merged, this is simply to discuss this issue and to provide a diff between the two computations.

### DIFF
--- a/examples/bipedal_walk.py
+++ b/examples/bipedal_walk.py
@@ -86,7 +86,7 @@ if WITHPLOT:
     plotSolution(ddp, bounds=False, figIndex=1, show=False)
 
     for i, phase in enumerate(GAITPHASES):
-        title = phase.keys()[0] + " (phase " + str(i) + ")"
+        title = list(phase.keys())[0] + " (phase " + str(i) + ")"
         log = ddp[i].getCallbacks()[0]
         crocoddyl.plotConvergence(log.costs,
                                   log.u_regs,
--- a/examples/bipedal_walk_ubound.py
+++ b/examples/bipedal_walk_ubound.py
@@ -89,7 +89,7 @@ if WITHPLOT:
     plotSolution(ddp, bounds=False, figIndex=1, show=False)
 
     for i, phase in enumerate(GAITPHASES):
-        title = phase.keys()[0] + " (phase " + str(i) + ")"
+        title = list(phase.keys())[0] + " (phase " + str(i) + ")"
         log = ddp[i].getCallbacks()[0]
         crocoddyl.plotConvergence(log.costs,
                                   log.u_regs,
--- a/examples/quadrupedal_gaits.py
+++ b/examples/quadrupedal_gaits.py
@@ -138,7 +138,7 @@ if WITHPLOT:
     plotSolution(ddp, figIndex=1, show=False)
 
     for i, phase in enumerate(GAITPHASES):
-        title = phase.keys()[0] + " (phase " + str(i) + ")"
+        title = list(phase.keys())[0] + " (phase " + str(i) + ")"
         log = ddp[i].getCallbacks()[0]
         crocoddyl.plotConvergence(log.costs,
                                   log.u_regs,


examples/log/boxfddp_vs_boxddp.log
```diff
@@ -13,78 +13,78 @@ iter 	 cost 	      stop 	    grad 	  xreg 	      ureg 	 step 	 feas
...
-   1  2.23361e+07  6.70272e+10  9.81907e+07  1.00000e-09  1.00000e-09   0.0312     1
-   2  1.86846e+07  2.60223e+11  6.80717e+07  1.00000e-09  1.00000e-09   0.0156     1
-   3  1.74778e+07  9.57254e+09  5.50219e+07  1.00000e-09  1.00000e-09   0.1250     1
-   4  1.70684e+07  9.05376e+10  6.05775e+07  1.00000e-08  1.00000e-08   0.0020     1
-   5  1.58771e+07  9.04287e+10  6.03229e+07  1.00000e-07  1.00000e-07   0.0039     1
-   6  1.41584e+07  7.75480e+10  5.34607e+07  1.00000e-06  1.00000e-06   0.0039     1
-   7  1.29359e+07  1.85569e+11  4.29009e+07  1.00000e-05  1.00000e-05   0.0020     1
-   8  1.28508e+07  2.70894e+11  3.79058e+07  1.00000e-04  1.00000e-04   0.0039     1
-   9  1.20118e+07  2.50503e+11  3.44535e+07  1.00000e-04  1.00000e-04   0.0156     1
+   1  2.23353e+07  6.70272e+10  9.81907e+07  1.00000e-09  1.00000e-09   0.0312     1
```
proyan: This is a big difference in the computation of cost. It should not be here.

examples/log/boxfddp_vs_boxddp.log
```diff
@@ -13,78 +13,78 @@ iter 	 cost 	      stop 	    grad 	  xreg 	      ureg 	 step 	 feas
...
-  21  1.71141e+05  2.00479e+08  5.90291e+05  1.00000e-01  1.00000e-01   0.2500     0
-  22  1.27199e+05  1.26571e+08  4.09531e+05  1.00000e-01  1.00000e-01   0.5000     0
-  23  9.40734e+04  4.90223e+07  2.47952e+05  1.00000e-01  1.00000e-01   0.5000     0
-  24  8.17370e+04  1.52939e+07  1.56901e+05  1.00000e-01  1.00000e-01   0.5000     0
-  25  5.99645e+04  9.77329e+06  1.19667e+05  1.00000e-02  1.00000e-02   1.0000     1
-  26  5.76511e+04  1.39711e+06  1.23761e+05  1.00000e-02  1.00000e-02   0.0625     1
-  27  4.99673e+04  2.13840e+06  8.06700e+04  1.00000e-02  1.00000e-02   0.1250     1
-  28  4.30063e+04  3.64565e+06  5.90217e+04  1.00000e-02  1.00000e-02   0.1250     1
-  29  4.18444e+04  1.35353e+06  4.46167e+04  1.00000e-02  1.00000e-02   0.1250     1
+  20  2.34231e+05  1.70003e+08  1.14113e+06  1.00000e-01  1.00000e-01   0.5000     0
```
proyan: Same here.

examples/log/boxfddp_vs_boxddp.log
jcarpent: You should check which variables are used and not initialized with valgrind.

cmastalli:
@proyan I am not sure if this might be connected, but I merged recently a PR that solves an issue, in quadruped examples, reported by @wxmerkt.

This is the PR:
https://github.com/loco-3d/crocoddyl/pull/694

So my question is did you use this commit when you generated these log files?

proyan:
> @proyan I am not sure if this might be connected, but I merged recently a PR that solves an issue, in quadruped examples, reported by @wxmerkt.
> 
> This is the PR:
> #694
> 
> So my question is did you use this commit when you generated these log files?

Yes, this is the logs from the commit fc20ca74b1674c88bf83cb1c4fa2d05ab85d5e42

wxmerkt:
@proyan Are these diffs with multi-threading? Can you try to run the same with a clean build with a single thread build and check the logs again? I had a similar occurrence back in autumn when using multi-threading, however, I thought you had it fixed back then?

proyan:
@wxmerkt These logs are with the standard cmake install that we have with crocoddyl (i.e without Multi-threading and without vectorization)

examples/log/boxfddp_vs_boxddp.log
proyan: Ok, Thanks. I'll have a look

wxmerkt:
Okay, thank you for the clarification. That's a good insight

cmastalli:
@proyan I don't know if useful, but could you report the version of examples-robot-data, pinocchio, eigen(py) and numpy?

examples/log/boxfddp_vs_boxddp.log
cmastalli: This difference might be related to your hardware / setup.
Remember that these files are generated with an specific PC and setup.

Unfortunately, it is hard to make an statement here.

I wonder if we should keep this discussion. I am OK to keep it if @proyan wants to investigate deeper. I am afraid that I don't have time for it.

proyan:
I'll close this PR since this problem has already been documented in separate issues.

PR #695, opened 2020-02-20
Code Refactoring on Template Scalar

proyan:
This PR refactors the code based on the Scalar template.
1) The tests are passing and are unchanged.
2) The examples are passing and are unchanged.
3) The user API is unchanged.
4) The solvers, the utils and the bindings are unchanged.

This PR follows the issue in #691. This is a big change in the current code structure, and thus I want to make sure that all our partners are well aware of this.

There is no new feature being added in this PR, and this PR only focuses on the code refactoring.

PR #698, opened 2020-02-24
Improving compilation of Python bindings

cmastalli:
This PR reorganizes entirely the compilation of the python bindings.
Now it compiles each single file which reduces the memory used by the PC.

There are no additional modification in this PR.


wxmerkt:
CI sadly says that compilation time _increased by up to 50%_:

## Current devel
image

## This PR
image

cmastalli:
> CI sadly says that compilation time _increased by up to 50%_:
> 
> ## Current devel
> image
> 
> ## This PR
> image

Now the bindings are compiled in different files.
I was expecting a decrease in the compilation time.
How can we check the memory usage?

To be honest the big problem is the memory needed to compile the bindings.

cmastalli:
Btw could we trust on these results? I mean it's deterministic the behaviour of Travis

cmastalli:
Some header files also include its implementation, I will see if I can improve that.
I will also double check if there are useless header inclusions.
A part of them, I don't know which other things might be improved.

wxmerkt:
I don't think Travis is fully reliable but that's quite a noticeable difference for build times that were more or less steady. [Unless Travis recently changes something with build resources in the past week]

Did you run fully clean builds (no ccache) before and after on your workstation and how much is the improvement roughly?

cmastalli:
@wxmerkt 

I have monitored the compilation time and memory usage for both cases (current and proposed compilation).  This is the setup that I used:
 - `time make -j12` (compilation time).
 - `make -j1` (memory usage).
 These are the results

|   | compilation time (real)  | compilation time (real) | memory (RES) |
|---|---|---|---|
| current  |  1m57.516s | 10m27.200s   |  max. 3400MB (often around 3000MB) |
| proposed  |  1m57.880s | 15m56.646s  |  max. 1000MB (often around 500MB) |

Note that the memory usage reported here are only for the compilation of the Python bindings.

Additionally, the proposed method uses less memory than current compilation of the Crocoddyl library. The highest peaks of used memory are during the compilation of `free-fwddyn`, `contact-fwddyn` and `impulse-fwddyn`. In this case, the memory used is max. 2400MB. Therefore, we could argue that Crocoddyl Python bindings are not the bottleneck with the proposed method.

Furthermore, the compilation time increases significantly (around 50%) when we don't know use multiple-threads.

wxmerkt:
Thank you for the detailed analysis. Since I don't know Boost.Python too well, I'd defer to @jcarpent / @proyan to comment.

Crocoddyl CI does not have memory exhaustion right now, it was/is the ROS buildfarm for Pinocchio that did (and a laptop).

cmastalli:
> Crocoddyl CI does not have memory exhaustion right now, it was/is the ROS buildfarm for Pinocchio that did (and a laptop).

I apologize if I wasn't explicitly enougth. The memory exhaustion appears with the templatization of Crocoddyl (for more details see https://github.com/loco-3d/crocoddyl/pull/695#issuecomment-589441205)

proyan:
It looks okay to me. I cherry-picked these commits on #695 to see if we still have memory exhaustion.

proyan:
I did a diff between the files and noticed a lot of renamings from hpp to cpp. I've tried to compare them, but it is still possible that I missed something. I'm trusting that @cmastalli didn't introduce any other changes in this PR. 

If that is the case and #695 passes, we can directly merge #695 and close this PR.

cmastalli:
> I did a diff between the files and noticed a lot of renamings from hpp to cpp. I've tried to compare them, but it is still possible that I missed something. I'm trusting that @cmastalli didn't introduce any other changes in this PR.
> 
> If that is the case and #695 passes, we can directly merge #695 and close this PR.

Yes, this is terrible for reviewing. I am so sorry for that but it was needed to make it work.

cmastalli:
> I did a diff between the files and noticed a lot of renamings from hpp to cpp. I've tried to compare them, but it is still possible that I missed something. I'm trusting that @cmastalli didn't introduce any other changes in this PR.
> 
> If that is the case and #695 passes, we can directly merge #695 and close this PR.

I prefer to merge it now, I want to have some extra time to check and or push modifications of the code, anything fundamental of course.

PR #700, opened 2020-02-27
FDDP: Fix expectedImprovement for terminalModel

wxmerkt:
I believe this should be `terminalModel` as the running models all get updated with the for-loop below.

I checked with the biped and quadruped examples and there was no difference in the logs for convergence. In practice, I don't believe this to have too much of an impact but this should be the correct behaviour.


cmastalli:
@wxmerkt thanks for taking care of this.

You're right! this doesn't affect any of our example because we use the same state for all nodes. However, it is not correct to use the last running model.

PR #712, opened 2020-03-11
CostModelContainer: Make CostItem a shared_ptr

wxmerkt:
Related to #672

### DIFF
--- a/bindings/python/crocoddyl/multibody/costs/cost-sum.cpp
+++ b/bindings/python/crocoddyl/multibody/costs/cost-sum.cpp
@@ -26,6 +26,8 @@ void exposeCostSum() {
   // Register custom converters between std::map and Python dict
   typedef boost::shared_ptr<CostItem> CostItemPtr;
   typedef boost::shared_ptr<CostDataAbstract> CostDataPtr;
+
+  bp::register_ptr_to_python<boost::shared_ptr<CostItem> >();
   bp::to_python_converter<std::map<std::string, CostItemPtr, std::less<std::string>,
                                    std::allocator<std::pair<const std::string, CostItemPtr> > >,
                           map_to_dict<std::string, CostItemPtr> >();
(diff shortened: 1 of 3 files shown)


wxmerkt:
I need to fix the Python bindings. Unfortunately, I am not that experienced with Boost.Python as I mostly use Pybind11 that does these translations automatically. I will provide a fix later this afternoon.

wxmerkt:
I still get this error:

```
Traceback (most recent call last):
  File "unittest/bindings/test_momemtum_cost.py", line 68, in <module>
    cmodel = model.costs.costs['momentum'].cost
TypeError: No Python class registered for C++ class boost::shared_ptr<crocoddyl::CostItemTpl<double> >
```

Alas, I don't know what I need to do in Boost.Python beyond `register_ptr_to_python`. Can you please advise? :-)

bindings/python/crocoddyl/multibody/costs/cost-sum.cpp
```diff
@@ -24,14 +24,17 @@ BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS(CostModelSum_calc_wraps, CostModelSum::ca
...
+  typedef boost::shared_ptr<CostItem> CostItemPtr;
   typedef boost::shared_ptr<CostDataAbstract> CostDataPtr;
-  bp::to_python_converter<std::map<std::string, CostItem, std::less<std::string>,
-                                   std::allocator<std::pair<const std::string, CostItem> > >,
-                          map_to_dict<std::string, CostItem> >();
+
+  bp::register_ptr_to_python<boost::shared_ptr<CostItem> >();
+  bp::to_python_converter<std::map<std::string, CostItemPtr, std::less<std::string>,
+                                   std::allocator<std::pair<const std::string, CostItemPtr> > >,
+                          map_to_dict<std::string, CostItemPtr> >();
```
cmastalli: I think you need to add the `false` flag, i.e.
```cpp
  bp::to_python_converter<std::map<std::string, CostItemPtr, std::less<std::string>,
                                   std::allocator<std::pair<const std::string, CostItemPtr> > >,
                          map_to_dict<std::string, CostItemPtr, false> >();
```

This should fix the issue. Let me know if this is not the case

wxmerkt: Thank you, this fixed it! I pushed the changes now and it should pass the buildserver soon.

wxmerkt:
Thank you, this is now passing CI and should be good to merge once Travis goes green :-).

cmastalli:
> Thank you, this is now passing CI and should be good to merge once Travis goes green :-).

Cool. Thanks for making this effort :)

Issue #184, opened 2019-11-18
Enabling writing of internal data in Eigen object for Python bindings


wxmerkt:
In GitLab by @cmastalli on Jun 28, 2019, 16:04

For more information about it see: https://github.com/stack-of-tasks/eigenpy/issues/65.

wxmerkt:
In GitLab by @cmastalli on Jun 28, 2019, 16:05

changed the description

wxmerkt:
In GitLab by @cmastalli on Jun 28, 2019, 16:05

@ MaximilienNaveau and @hammoudbilal feel free to tackle this issue in eigenpy, just let me know if that the case.

Otherwise I will try to solve it in my free time.

wxmerkt:
In GitLab by @cmastalli on Jun 28, 2019, 16:05

mentioned in merge request !171

cmastalli:
Solved in #715

PR #715, opened 2020-03-12
Allowed Python users to write internal data of NumPy objects

cmastalli:
This PR tackles the issue #184 which was first solved in https://github.com/stack-of-tasks/eigenpy/pull/160.

Concretely, it allows the Python users to write values in internal segments of NumPy data. For instance, it is possible to do the follows:
```python
class MyDerivatedModel(crocoddyl.ActionModelAbstract):
    def __init__(self):
        nx, nu, nr = 3, 2, 5
        crocoddyl.ActionModelAbstract.__init__(self, crocoddyl.StateVector(nx), nu, nr)
        self.dt = .1
        self.costWeights = [10., 1.]

    def calc(self, data, x, u):
        # it allows to do this for data.r (and all data members of course):
        data.r[:3] = self.costWeights[0] * x 
        data.r[3:] = self.costWeights[1] * u
        # instead of being constraint to do the follows
        # data.r = np.vstack([self.costWeights[0] * x, self.costWeights[1] * u])
```

On the other hand, there are plenty of getter functions that return `const ref`. For those functions, this PR proposes to return the `const ref` for efficiency reason. Nnote that you can't write internally as it is `const ref`.

### DIFF
--- a/bindings/python/crocoddyl/core/action-base.cpp
+++ b/bindings/python/crocoddyl/core/action-base.cpp
@@ -79,11 +79,11 @@ void exposeActionAbstract() {
                     "indicates whether problem has finite control limits")
       .add_property(
           "u_lb",
-          bp::make_function(&ActionModelAbstract_wrap::get_u_lb, bp::return_value_policy<bp::return_by_value>()),
+          bp::make_function(&ActionModelAbstract_wrap::get_u_lb, bp::return_internal_reference<>()),
           &ActionModelAbstract_wrap::set_u_lb, "lower control limits")
       .add_property(
           "u_ub",
-          bp::make_function(&ActionModelAbstract_wrap::get_u_ub, bp::return_value_policy<bp::return_by_value>()),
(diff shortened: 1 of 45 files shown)


wxmerkt:
> Nnote that you can't write internally as it is const ref.

Will it throw an error/exception in Python or just silently not do it?

cmastalli:
@wxmerkt yes, it does. But this eigenpy feature

cmastalli:
@nim65s this PR requires at least eigenpy v2.1.0.

Could you let me know when it's scheduled the release of this version?

nim65s:
next week

jcarpent:
@nim65s Just wait a bit. I will add the possibility to not share the memory but returns a copy in EigenPy to allow backward compatibility. This will be in EigenPy 2.2.0.

jcarpent:
That's also why I did not put this feature in Pinocchio yet.

cmastalli:
> next week

@nim65s what is the status of this release?

Issue #278, opened 2019-11-18
Add proper virtual keyword


wxmerkt:
In GitLab by @wxmerkt on Oct 23, 2019, 12:45

Currently, the implementations of (pure) virtual functions do not carry the `virtual` keyword. This is prone to introduce bugs and also requires copy-paste/duplication in the bindings. We should add the `virtual` (C++98) or `override/final` (C++11, once switch confirmed) keywords.

wxmerkt:
This just came back to bite me badly with overrides in Python and C++. We should make fixing the polymorphism and cleaning up the API a priority. There's a lot of duplication that can be removed with proper inheritance.

cmastalli:
> This just came back to bite me badly with overrides in Python and C++. We should make fixing the polymorphism and cleaning up the API a priority. There's a lot of duplication that can be removed with proper inheritance.

Could you handle this task?
Unfortunately, I don't have any bandwidth to work on it before beginning of February.

wxmerkt:
I won't have bandwidth before March. New action + contact model takes priority for me.

PR #717, opened 2020-03-13
Add virtual keyword to implemented virtual functions

cmastalli:
This PR solves the issue #278.


wxmerkt:
Awesome, thank you for fixing this important issue. Since we implicitly have a C++11 requirement (URDF parser, also see CI requiring C++11 on Xenial), we may want to use `override` when overriden to indicate it's not the first declaration

cmastalli:
> Awesome, thank you for fixing this important issue. Since we implicitly have a C++11 requirement (URDF parser, also see CI requiring C++11 on Xenial), we may want to use `override` when overriden to indicate it's not the first declaration

It sounds good for me.
Maybe we need to discuss when it coudl happen.
@nim65s could you let us know if we can pursue c++11 standard soon? or when?

nim65s:
I think for crocoddyl, we can decide to require c++11. And for almost all other softwares, but not on the very generic ones like eigenpy, hpp-fcl & pinocchio.

jcarpent:
I don't think that you need to use any C++11 feature here. robotpkg is forcing C++11, but this is a mistake and it will be fixed soon. 
And Xenial still relies on urdf 0.4.2 which is using boost::shared_ptr

wxmerkt:
Is there a requirement to still support C++ < C++14? Even on Xenial the default compilers support the features.

jcarpent:
We would prefer to not impose any restriction if it is not required. In the current context, adding or not "override" won't change anything, except adding C++11 as required feature, while until now, the code is compatible with all the standards, meaning that extra-lib can still use it.

jcarpent:
In addition, for C++14, not all the compilers are enforcing correctly all the rules, making hard to use this feature on all the platforms even if they claim it.

PR #723, opened 2020-03-23
Function to update a model for a given node

cmastalli:
This PR proposes a function that
 - updates the model of a i node, and
 - allocates data for that node.

This additional function is useful for MPC application, when we might want to replace the action model.


wxmerkt:
The CI failure is due to robotpkg not staging packages...

cmastalli:
> The CI failure is due to robotpkg not staging packages...

Thank you for noticing it.

PR #724, opened 2020-03-23
Squashing classes

PepMS:
This PR addresses the addition of squashing function classes.


include/crocoddyl/core/squashing-base.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+  virtual void calc(const boost::shared_ptr<SquashingDataAbstract>& data, const Eigen::Ref<const VectorXs>& u) = 0;
+  virtual void calcDiff(const boost::shared_ptr<SquashingDataAbstract>& data, const Eigen::Ref<const VectorXs>& u,
+                        const bool& recalc = true) = 0;
+  virtual boost::shared_ptr<SquashingDataAbstract> createData() {
+    return boost::make_shared<SquashingDataAbstract>(this);
+  }
+
+  const std::size_t& get_ns() const;
+  const VectorXs& get_in_lb() const { return in_lb_; };
```
cmastalli: Let's rename this function as:
```cpp
const VectorXs& get_u_lb() const { return u_lb_; };
```
and apply the same rule for the other functions.

include/crocoddyl/core/squashing/smooth-sat.hpp
```diff
@@ -0,0 +1,85 @@
...
+template <typename _Scalar>
+class SquashingModelSmoothSatTpl : public SquashingModelAbstractTpl<_Scalar> {
+ public:
+  typedef _Scalar Scalar;
+  typedef MathBaseTpl<Scalar> MathBase;
+  typedef SquashingModelAbstractTpl<Scalar> Base;
+  typedef SquashingDataAbstractTpl<Scalar> SquashingDataAbstract;
+  typedef typename MathBase::VectorXs VectorXs;
+
+  explicit SquashingModelSmoothSatTpl(const Eigen::Ref<const VectorXs>& out_lb,
```
cmastalli: Remove the keyword `explicit`.

include/crocoddyl/core/squashing/smooth-sat.hpp
```diff
@@ -0,0 +1,85 @@
...
+    // Squashing function used: "Smooth abs":
+    // s(u) = 0.5*(lb + ub + sqrt(smooth + (u - lb)^2) - sqrt(smooth + (u - ub)^2))
+    data->s =
+        0.5 * (Eigen::sqrt(Eigen::pow((u - out_lb_).array(), 2) + a_.array()) -
+               Eigen::sqrt(Eigen::pow((u - out_ub_).array(), 2) + a_.array()) + out_lb_.array() + out_ub_.array());
+  }
+
+  void calcDiff(const boost::shared_ptr<SquashingDataAbstract>& data, const Eigen::Ref<const VectorXs>& u,
+                const bool& recalc = true) {
+    if (recalc) {
```
cmastalli: Remove the recalc if-loop, i.e. these lines of code:
```cpp
if (recalc) {
  calc(data, u);
}
```

Note that we have changed all the `recalc` option.
You know will need to run manually `calc` for computing `calcDiff`.

include/crocoddyl/core/squashing/smooth-sat.hpp
```diff
@@ -0,0 +1,85 @@
...
+  void calc(const boost::shared_ptr<SquashingDataAbstract>& data, const Eigen::Ref<const VectorXs>& u) {
+    // Squashing function used: "Smooth abs":
+    // s(u) = 0.5*(lb + ub + sqrt(smooth + (u - lb)^2) - sqrt(smooth + (u - ub)^2))
+    data->s =
+        0.5 * (Eigen::sqrt(Eigen::pow((u - out_lb_).array(), 2) + a_.array()) -
+               Eigen::sqrt(Eigen::pow((u - out_ub_).array(), 2) + a_.array()) + out_lb_.array() + out_ub_.array());
+  }
+
+  void calcDiff(const boost::shared_ptr<SquashingDataAbstract>& data, const Eigen::Ref<const VectorXs>& u,
+                const bool& recalc = true) {
```
cmastalli: Remove `recalc` argument

include/crocoddyl/core/squashing/smooth-sat.hpp
```diff
@@ -0,0 +1,85 @@
...
+ protected:
+  using Base::out_lb_;
+  using Base::out_ub_;
+
+  using Base::in_lb_;
+  using Base::in_ub_;
+};
+}  // namespace crocoddyl
+
+#endif  // CROCODDYL_CORE_SQUASHING_SMOOTH_SAT_HPP_
```
cmastalli: Add an end line or better run the script for formatting the code. It is explained in the wiki.

src/core/squashing-base.cpp
```diff
@@ -0,0 +1,9 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2018-2020, LAAS-CNRS, University of Edinburgh, Universitat Politècinca de Catalunya
```
cmastalli: Remove `LAAS-CNRS`, keep the other affiliations

include/crocoddyl/core/squashing-base.hpp
PepMS: In the paper we called the input to the squashing function "s".
I understand that we could call it "u" since it is our decision variables.

How should we call the output of the squashing, which is now indicated by "out"? Independently from the paper notation, we could use "s" (this is how it is defined in the SquashingData struct).

include/crocoddyl/core/squashing/smooth-sat.hpp
PepMS: Already run. But it does not add the bottom line.

include/crocoddyl/core/squashing-base.hpp
cmastalli: I was confused indeed.

Then, I think we should call them as:
  - input: `s`
  - output: `u`

Do you agree with me?

PepMS: It makes more sense, yes. I will change how these variable are defined in the SquashingData struct too.

bindings/python/crocoddyl/core/actuation/squashing-base.cpp
```diff
@@ -0,0 +1,74 @@
...
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#include "python/crocoddyl/core/core.hpp"
+#include "python/crocoddyl/core/actuation/squashing-base.hpp"
+
+namespace crocoddyl {
+namespace python {
+
+// BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS(SquashingModel_calcDiff_wraps, SquashingModelAbstract::calcDiff_wrap, 2, 3)
```
cmastalli: Remove this line, we don't need it.

bindings/python/crocoddyl/core/actuation/squashing-base.cpp
```diff
@@ -0,0 +1,74 @@
...
+           ":param s: squashing input")
+      .def("calcDiff", pure_virtual(&SquashingModelAbstract_wrap::calcDiff), bp::args("self", "data", "s"),
+           "Compute the derivative of the squashing function.\n\n"
+           ":param data: squashing data\n"
+           ":param u: squashing input")
+      .def("createData", &SquashingModelAbstract_wrap::createData, bp::args("self"), "Create the squashing data.\n\n")
+      .add_property(
+          "ns",
+          bp::make_function(&SquashingModelAbstract_wrap::get_ns, bp::return_value_policy<bp::return_by_value>()),
+          "dimension of control vector")
```
cmastalli: Maybe it is better to say "dimension of the squashing vector"

However, I am OK if you like it as it is.

bindings/python/crocoddyl/core/actuation/squashing-base.cpp
```diff
@@ -0,0 +1,74 @@
...
+      "SquashingDataAbstract",
+      "Abstract class for squashing datas. \n\n"
+      "In crocoddyl, an squashing data contains all the required information for processing a \n"
+      "user-defined squashing model. The squashing data is typically allocated once per running.\n"
+      "model.createData().",
+      bp::init<SquashingModelAbstract*>(bp::args("self", "model"),
+                                        "Create common data shared between squashing models. \n\n"
+                                        "The squashing data uses the model in order to first process it. \n"
+                                        ":param model: squashing model"))
+      .add_property("u", bp::make_getter(&SquashingDataAbstract::u, bp::return_value_policy<bp::return_by_value>()),
```
cmastalli: With the latest (and recently) merged code #715, we can return by reference Eigen objects. This is possible due to recent improvements in eigenpy (v2.2.0).

Therefore, as returned policy, it is better to use `bp::return_internal_reference<>()` instead of `bp::return_value_policy<bp::return_by_value>()`.

Therefore, you need to write the following code:
```cpp
.add_property("u", bp::make_getter(&SquashingDataAbstract::u, bp::return_internal_reference<>()),
                              bp::make_setter(&SquashingDataAbstract::u), "squashing-output")
```

bindings/python/crocoddyl/core/actuation/squashing-base.cpp
```diff
@@ -0,0 +1,74 @@
...
+                    bp::make_setter(&SquashingDataAbstract::u), "squashing-output")
+      .add_property("du_ds",
```
cmastalli: same here, i.e.
```cpp
      .add_property("du_ds",
                    bp::make_getter(&SquashingDataAbstract::du_ds, bp::return_internal_reference<>()),
                    bp::make_setter(&SquashingDataAbstract::du_ds), "Jacobian of the squashing function");
```

include/crocoddyl/core/fwd.hpp
```diff
@@ -86,6 +86,13 @@ class ActuationDataAbstractTpl;
 template <typename Scalar>
 class ActuationModelAbstractTpl;
 
+// Squashing
+template <typename Scalar>
+class SquashingDataAbstractTpl;
+
+template <typename Scalar>
+class SquashingModelAbstractTpl;
+
```
cmastalli: You need to forward declared the `SquashingModelSmoothSatTpl`

include/crocoddyl/core/fwd.hpp
```diff
@@ -164,6 +171,9 @@ typedef StateAbstractTpl<double> StateAbstract;
 typedef ActuationDataAbstractTpl<double> ActuationDataAbstract;
 typedef ActuationModelAbstractTpl<double> ActuationModelAbstract;
 
+typedef SquashingDataAbstractTpl<double> SquashingDataAbstract;
+typedef SquashingModelAbstractTpl<double> SquashingModelAbstract;
+
```
cmastalli: You need to create the typedef of `SquashingModelSmoothSatTpl`, i.e.:
```cpp
typedef SquashingModelSmoothSatTpl<double> SquashingModelSmoothSat;
```

include/crocoddyl/core/actuation/squashing/smooth-sat.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+    smooth_ = 0.1;
+
+    d_ = (u_ub_ - u_lb_) * smooth_;
+    a_ = d_.array() * d_.array();
+  }
+
+  ~SquashingModelSmoothSatTpl();
+
+  void calc(const boost::shared_ptr<SquashingDataAbstract>& data, const Eigen::Ref<const VectorXs>& s) {
```
cmastalli: I think you removed the `virtual` keyword, it is a good practice to add it as well in derived classes.
```cpp
virtual void calc(const boost::shared_ptr<SquashingDataAbstract>& data, const Eigen::Ref<const VectorXs>& s) {
```

include/crocoddyl/core/actuation/squashing/smooth-sat.hpp
```diff
@@ -0,0 +1,87 @@
...
+    // Squashing function used: "Smooth abs":
+    // s(u) = 0.5*(lb + ub + sqrt(smooth + (u - lb)^2) - sqrt(smooth + (u - ub)^2))
+    data->u = 0.5 * (Eigen::sqrt(Eigen::pow((s - u_lb_).array(), 2) + a_.array()) -
+                     Eigen::sqrt(Eigen::pow((s - u_ub_).array(), 2) + a_.array()) + u_lb_.array() + u_ub_.array());
+  }
+
+  void calcDiff(const boost::shared_ptr<SquashingDataAbstract>& data, const Eigen::Ref<const VectorXs>& s) {
```
cmastalli: same here
```cpp
virtual void calcDiff(const boost::shared_ptr<SquashingDataAbstract>& data, const Eigen::Ref<const VectorXs>& s) {
```

bindings/python/crocoddyl/utils/__init__.py
```diff
@@ -579,8 +598,8 @@ def calc(self, data, x):
         data.a0[:] = pinocchio.getFrameAcceleration(self.state.pinocchio, data.pinocchio,
                                                     self.xref.frame).linear + pinocchio.utils.cross(self.vw, self.vv)
         if self.gains[0] != 0.:
-            data.a0[:] += np.asscalar(
-                self.gains[0]) * (data.pinocchio.oMf[self.xref.frame].translation - self.xref.oxf)
+            data.a0[:] += np.asscalar(self.gains[0]) * (
```
cmastalli: I believe this change will produce a problem in the Python code format.

We are also checking the Python code format, please run `yafp -ri .` in the root folder. More explanations are in the wiki!

PR #725, opened 2020-03-24
Created abstract function for setting and getting cost references

cmastalli:
Currently, if we want to set a new reference value to a cost function (from an action model) we need to downcast cost to the specific one.
This creates a lot of code complexity for c++ developers due to tons of code dedicated to do downcasting.

With this PR, I proposed a solution to reduce this undesired complexity.
Now if you want to retrieve a reference from a cost abstraction, you could do:
```cpp
Eigen::VectorXd u_ref;
control_cost.get_reference(u_ref);

crocoddyl::FramePlacement M_ref;
frame_cost.get_reference(M_ref);
```
instead to set a reference value, you might do:
```cpp
control_cost.set_reference(u_ref);
frame_cost.set_reference(M_ref);
```

**Note that the old getter and setter functions are there**. So these modifications are back compatible and does not break the current API.

@wxmerkt and @proyan -- this features are for you. I will work in other aspects to make our life much easier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ APPLY_DEFAULT_APPLE_CONFIGURATION()
 
 # Add the different required and optional dependencies
 ADD_REQUIRED_DEPENDENCY("eigen3 >= 3.0.5")
-ADD_REQUIRED_DEPENDENCY("eigenpy")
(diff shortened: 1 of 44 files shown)


wxmerkt:
Thank you for working on this important feature. I haven't looked into the diff closely but expect this to simplify life significantly :-).

PR #726, opened 2020-03-24
Included status information of costs (active / inactive)

cmastalli:
This PR includes the cost status as proposed in #720.

As mentioned in previous pull request, the idea behind of this is to have a global mechanism allocation.

Additionally, this PR includes unittest code for cost sum.

**Note the changes of this PR are still back compatible with previous code.**

### DIFF
--- a/bindings/python/crocoddyl/multibody/costs/cost-sum.cpp
+++ b/bindings/python/crocoddyl/multibody/costs/cost-sum.cpp
@@ -116,6 +116,9 @@ void exposeCostSum() {
       .add_property("nu", bp::make_function(&CostModelSum::get_nu, bp::return_value_policy<bp::return_by_value>()),
                     "dimension of control vector")
       .add_property("nr", bp::make_function(&CostModelSum::get_nr, bp::return_value_policy<bp::return_by_value>()),
+                    "dimension of the residual vector of active cost")
+      .add_property("nr_total",
+                    bp::make_function(&CostModelSum::get_nr_total, bp::return_value_policy<bp::return_by_value>()),
                     "dimension of the total residual vector");
 
   bp::class_<CostDataSum, boost::shared_ptr<CostDataSum>, boost::noncopyable>(
--- a/bindings/python/crocoddyl/multibody/costs/cost-sum.cpp
+++ b/bindings/python/crocoddyl/multibody/costs/cost-sum.cpp
@@ -38,15 +38,14 @@ void exposeCostSum() {
 
   bp::register_ptr_to_python<boost::shared_ptr<CostItem> >();
 
-  bp::class_<CostItem, boost::noncopyable>(
-      "CostItem", "Describe a cost item.\n\n",
-      bp::init<std::string, boost::shared_ptr<CostModelAbstract>, double, bp::optional<bool> >(
-          bp::args("self", "name", "cost", "weight", "active"),
-          "Initialize the cost item.\n\n"
-          ":param name: cost name\n"
-          ":param cost: cost model\n"
-          ":param weight: cost weight\n"
-          ":param active: True if the cost is activated (default true)"))
+  bp::class_<CostItem>("CostItem", "Describe a cost item.\n\n",
+                       bp::init<std::string, boost::shared_ptr<CostModelAbstract>, double, bp::optional<bool> >(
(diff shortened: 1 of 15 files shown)


bindings/python/crocoddyl/multibody/costs/cost-sum.cpp
```diff
@@ -68,14 +73,20 @@ void exposeCostSum() {
...
+                                        "Add a cost item.\n\n"
+                                        ":param name: cost name\n"
+                                        ":param cost: cost model\n"
+                                        ":param weight: cost weight\n"
+                                        ":param active: True if the cost is activated (default true)"))
       .def("removeCost", &CostModelSum::removeCost, bp::args("self", "name"),
            "Remove a cost item.\n\n"
            ":param name: cost name")
+      .def("changeCostStatus", &CostModelSum::changeCostStatus, bp::args("self", "name", "active"),
+           "Change the cost status.\n\n"
```
wxmerkt: Can't we deduce `active` or `inactive` from whether the `weight==0`?

cmastalli: No, we can't.

We could still run computation (`active=true`) with a `weight=0.`.

It is possible to adopt the policy of disabling the computation (`active=false`) whenever we add a cost with `weight=0.`, however this might produce user errors if this later set a weight different to zero, e.g.
```cpp
sum_cost.addCost("my_cost", cost, 0., true); // if we adopt this policy, then `active=false` without any user acknowledgment
sum_cost.get_costs()->second("my_cost")->weight = 1.;
```

So I am not sure we should adopt this policy, what do you think?

wxmerkt: Okay, I see. So the question becomes what happens if `nr_` changes?

cmastalli: If the cost is disable, then the total `nr_` is decreased. The same for the activating a cost.

There is an unittest code that checks this. See this: https://github.com/cmastalli/crocoddyl/blob/topic/cost-status/unittest/test_cost_sum.cpp#L52

Note that `nr_` (for cost sum) represents the total `nr_`.

cmastalli: @wxmerkt do you have any further question or request?

wxmerkt: Yes, I saw that it is decreased. Does that change allocation or matrix sizes? Is another operation required after activating/deactivating to avoid data corruption or sth therelike?

cmastalli: This is a good point, let me add some relevant information about it.

`nr` is not use for data allocation inside cost sum, however, it is used for numdiff cost sum.

Therefore, we would expect an error if we change the cost status and use numdiff to retrieve calcDiff.

To solve this, we should have to kind of `nr`: one for the current `nr` another one for the total `nr`. This latter one is only used to allocate data for numdiff. Concretely, I proposed a new function call `get_nr_total()`.

**Further notes**
We might have a similar problem for the contact and impulse status. I will handle this issue in this PR too, and will also include other unittest code for checking it.

wxmerkt: Thank you for the detailed explanation and highlighting that this issue is (luckily) limited to the num-diff cases. I think we should use a more descriptive name for `nra`, otherwise your proposal sounds great and clarifies all my questions :-)

cmastalli: What about the above options?
 1. `get_nr_allocated`
 2. `get_nr_total`

any preference?

wxmerkt: (2) sounds good :-) `allocated` is a concept that may need further explanation

cmastalli:
The additional commits include:
 1. The concept of `nr`, `nc` and `ni` total for global allocation
 2. Cost sum data doesn't inherit from cost data abstract.
 3. Multiple contact / impulse datas don't inherit from contact / data data abstract.
 4. Unittest code for `nr`, `nc` and `ni` total.
 5. Allowed to register copy of cost sum, multiple contact and impulse classes (Python bindings)

Issue #728, opened 2020-03-27
Bad initialization of dx0 in the forward pass


edantec:
The gap f0 should be used to initialize the first dx0, rather than force it to be dx0 = 0. In practice, f0 should be equal to the difference of the initial constrained state x0 (provided during the creation of the shooting problem) and the first state of the warm state trajectory, provided during the ddp->solve.

cmastalli:
@edantec -- Nicolas explained the solution, could you provide a PR with this work?
If so, note that you will need to review:
  1. `SolverDDP`
  2. `SolverFDDP`
  3. `SolverKKT`
and check that the unit-tests are passing.

cmastalli:
@edantec, @proyan and @nmansard 

I review a second time the code for both: DDP and FDDP, and **I conclude that we don't have this issue. I mean the solvers are properly handling the initial condition**. Let me to explain it as easy as possible.

However, before starting the explanation, we need to say that `dx` has `T` length (not `T+1` length). In practice means that `dx0` isn't associated to the initial condition, instead it's associated to the first equality condition (dynamic condition). Note that we don't need to allocate another `dx` for the initial condition in any of the solvers.

**DDP solver***
In this case, we already know that the first forward pass will close the gaps. To ensure this, I have defined the `xs_try` with the initial state (https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solvers/ddp.cpp#L43). And this is needed for running the forward pass (i.e. `dx0 = x0 - xs[0]`) as you can see in https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solvers/ddp.cpp#L250.


***FDDP solver**
If the user provides an infeasible warm-start for the initial condition, then the FDDP solver accounts it during the calculation of the gap `f0` (https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solvers/ddp.cpp#L156) (as well the DDP solver). Then the forward pass will close the gaps as much as defined by the step length (https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solvers/box-fddp.cpp#L98).

---
Do you agree with me?

cmastalli:
I will close this issue.
Feel free to re-open if that the case.

PR #733, opened 2020-03-30
Add initialization for first state in solver

edantec:
dx0 should be equal to the difference between x0 constraint as defined in the problem, and x0 variable as provided in the guess.

### DIFF
--- a/src/core/solvers/ddp.cpp
+++ b/src/core/solvers/ddp.cpp
@@ -246,8 +246,13 @@ void SolverDDP::forwardPass(const double& steplength) {
   for (std::size_t t = 0; t < T; ++t) {
     const boost::shared_ptr<ActionModelAbstract>& m = problem_->get_runningModels()[t];
     const boost::shared_ptr<ActionDataAbstract>& d = problem_->get_runningDatas()[t];
-
-    m->get_state()->diff(xs_[t], xs_try_[t], dx_[t]);
+    
+    if (t == 0) {
+      m->get_state()->diff(problem_->get_x0(), xs_[0], dx_[0]);
+    }
+    else {
+	  m->get_state()->diff(xs_[t], xs_try_[t], dx_[t]);
+	}
     us_try_[t].noalias() = us_[t];
     us_try_[t].noalias() -= k_[t] * steplength;
     us_try_[t].noalias() -= K_[t] * dx_[t];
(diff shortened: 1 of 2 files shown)


src/core/solvers/ddp.cpp
```diff
@@ -246,8 +246,13 @@ void SolverDDP::forwardPass(const double& steplength) {
   for (std::size_t t = 0; t < T; ++t) {
     const boost::shared_ptr<ActionModelAbstract>& m = problem_->get_runningModels()[t];
     const boost::shared_ptr<ActionDataAbstract>& d = problem_->get_runningDatas()[t];
-
-    m->get_state()->diff(xs_[t], xs_try_[t], dx_[t]);
+    
+    if (t == 0) {
+      m->get_state()->diff(problem_->get_x0(), xs_[0], dx_[0]);
```
proyan: I think it is better if you don't specialize for t==0 here. Rather, you should look at 
1) https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solvers/ddp.cpp#L43
2) https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solvers/ddp.cpp#L155
3) https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solvers/ddp.cpp#L345
At all these places, we are assigining xs_try_ to be equal to get_x0(). Instead, we need to make sure that dx0 is properly accounted for in these assignments.

src/core/solvers/fddp.cpp
```diff
@@ -179,7 +179,12 @@ void SolverFDDP::forwardPass(const double& steplength) {
     } else {
       m->get_state()->integrate(xnext_, fs_[t] * (steplength - 1), xs_try_[t]);
     }
-    m->get_state()->diff(xs_[t], xs_try_[t], dx_[t]);
+    if (t == 0) {
```
proyan: same here.

cmastalli:
We are already properly handling the initial conditions in both solvers. Therefore, we don't need this PR (and most probably there is an error).

See my explanation here: **https://github.com/loco-3d/crocoddyl/issues/728#issuecomment-605958175**

Let me know if you agree.

edantec:
Ok, I agree with that: I didn't see that initial condition was already handled.

PR #735, opened 2020-03-31
Improved efficiency in forward passes + removed deprecated pinocchio signatures

cmastalli:
This PR removes the useless extra if-loop due to the feasibility or step-length status in FDDP and Box-FDDP. It also removes an useless line in the Box-DDP forward pass.

On the other hand, it removes the deprecated Pinocchio signatures in Python code. Finally, it fixes an issue in the frame trajectory display when there is not contact (fly phases).


wxmerkt:
> This PR removes the useless extra if-loop due to the feasibility or step-length status in FDDP and Box-FDDP.

It appears that the `if` is still there - now just outside the loop. I.e. before: if inside the for loop. Now two for loops in if-else. Correct? Or was there something else to it?

What was the gained efficiency improvement?

bindings/python/crocoddyl/__init__.py
```diff
@@ -9,12 +9,12 @@
 def rotationMatrixFromTwoVectors(a, b):
     a_copy = a / np.linalg.norm(a)
     b_copy = b / np.linalg.norm(b)
-    a_cross_b = pinocchio.utils.cross(a_copy, b_copy)
+    a_cross_b = np.cross(a_copy, b_copy, axis=0)
```
jcarpent, line 12: Use the Quaternion module of eigenpy here.

PR #736, opened 2020-03-31
Multicopter body actuation

PepMS:
The aim of this PR is to add an actuation model for aerial vehicles. Basically, any robot with its base-link actuated by propellers.


nim65s:
Do you need a release of example-robot-data for this ?

PepMS:
I don't think so. The examples including a quadcopter are using the hector model. Any use of the iris  will be local. Thanks anyway!

bindings/python/crocoddyl/multibody/actuations/multicopter-base.cpp
```diff
@@ -0,0 +1,52 @@
...
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#include "python/crocoddyl/multibody/multibody.hpp"
+#include "crocoddyl/multibody/actuations/multicopter-base.hpp"
+
+namespace crocoddyl {
+namespace python {
+
+void exposeActuationModelMCBase() {
```
wxmerkt: Can we please spell "MC" out to "MultiCopter"?

cmastalli:
@PepMS I will make a proper review asap. In the meantime you need to fix this problem: https://travis-ci.org/github/loco-3d/crocoddyl/jobs/669181053#L791. It seems you forgot to push that file.

bindings/python/crocoddyl/multibody/actuations/multicopter-base.cpp
PepMS: Sure, will do

include/crocoddyl/multibody/actuations/multicopter-base.hpp
```diff
@@ -0,0 +1,84 @@
...
+    boost::shared_ptr<ActuationDataAbstract> data = boost::make_shared<ActuationDataAbstract>(this);
+
+    data->dtau_du = tau_f_;
+
+    return data;
+  }
+
+  const std::size_t& get_nrotors() const { return n_rotors_; };
+  const MatrixXs& get_tauf() const { return tau_f_; };
+  void set_tauf(const Eigen::Ref<const MatrixXs>& tau_f) { tau_f_ = tau_f; }
```
cmastalli, line 71: You could rename it to 
```cpp
set_parameters(const double& d_cog, const double& cm, const double& cf) {
 // write tau_f from these parameters
```cpp

File: include/crocoddyl/multibody/actuations/multicopter-base.hpp [hunk 3]
```diff
@@ -0,0 +1,84 @@
...
+  boost::shared_ptr<ActuationDataAbstract> createData() {
+    boost::shared_ptr<ActuationDataAbstract> data = boost::make_shared<ActuationDataAbstract>(this);
+
+    data->dtau_du = tau_f_;
+
+    return data;
+  }
+
+  const std::size_t& get_nrotors() const { return n_rotors_; };
+  const MatrixXs& get_tauf() const { return tau_f_; };
```
cmastalli, line 70: You should remove this function and change for
```cpp
const double& get_d_cog() const { return d_cog_; }
const double& get_cm() const { return cm_; }
const double& get_cf() const { return cf_; }
```

include/crocoddyl/multibody/actuations/multicopter-base.hpp
```diff
@@ -0,0 +1,84 @@
...
+  void set_tauf(const Eigen::Ref<const MatrixXs>& tau_f) { tau_f_ = tau_f; }
+
+ protected:
+  // Specific of multicopter
+  MatrixXs tau_f_;  // Matrix from rotors thrust to body force/moments
+  std::size_t n_rotors_;
+
```
cmastalli, line 77: Add
```cpp
double d_cog_;
double cm_;
double cf_;
```
and include the description of each variable

src/multibody/actuations/multicopter-base.cpp
```diff
@@ -0,0 +1,9 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2018-2019, LAAS-CNRS, IRI: CSIC-UPC
```
cmastalli, line 4:
```cpp
// Copyright (C) 2018-2020, LAAS-CNRS, IRI: CSIC-UPC
```

examples/quadrotor.py
```diff
@@ -19,12 +18,15 @@
...
+d_cog = 0.1525
 cf = 6.6e-5
 cm = 1e-6
 u_lim = 5
 l_lim = 0.1
-actModel = ActuationModelUAM(state, '+', distance_rotor_COG, cm, cf, u_lim, l_lim)
+tau_f = np.matrix([[0.0, 0.0, 0.0, 0.0], [0.0, 0.0, 0.0, 0.0], [1.0, 1.0, 1.0, 1.0], [0.0, d_cog, 0.0, -d_cog],
+                   [-d_cog, 0.0, d_cog, 0.0], [-cm / cf, cm / cf, -cm / cf, cm / cf]])
+
+actModel = crocoddyl.ActuationModelMultiCopterBase(state, 4, tau_f)
```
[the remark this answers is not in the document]
PepMS: cf, cm are aerodynamic parameters of the propellers and d_cog is the distance of the respective propellers to the center of gravity.

To get the matrix tau_f from the stated parameters, it is not only necessary their values but you also need to know how the propellers are distributed around the center of gravity (number of propellers and rotation direction). It would be great to have this info in the urdf and you can extract it from there. However, I don't think this part has to be done in Crocoddyl. The essential part you need to know is the tau_f, that is why I only added this to the API.

In that sense, I have a separated library that parses a yaml file where you can specify a multirotor geometry and outputs the tau_f matrix that Crocoddyl needs.

include/crocoddyl/multibody/actuations/multicopter-base.hpp
PepMS: see answer above. If we do so, we are assuming a specific multirotor geometry, i.e. '+' quadrotor. What if we'd like to test an hexacopter or an octocopter?

include/crocoddyl/multibody/actuations/multicopter-base.hpp
PepMS: same above.

examples/quadrotor.py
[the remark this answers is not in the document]
cmastalli: Ok, I didn't think about different rotor geometries.

Then it makes sense to me, as well, to keep the API with `tau_f`. Let's keep it as it is --> no action here.

Thank you for the explanation.

include/crocoddyl/multibody/actuations/multicopter-base.hpp
```diff
@@ -0,0 +1,84 @@
...
+ public:
+  typedef _Scalar Scalar;
+  typedef MathBaseTpl<Scalar> MathBase;
+  typedef ActuationModelAbstractTpl<Scalar> Base;
+  typedef StateMultibodyTpl<Scalar> StateMultibody;
+  typedef ActuationDataAbstractTpl<Scalar> ActuationDataAbstract;
+  typedef typename MathBase::VectorXs VectorXs;
+  typedef typename MathBase::MatrixXs MatrixXs;
+
+  ActuationModelMultiCopterBaseTpl(boost::shared_ptr<StateMultibody> state, const std::size_t& n_rotors,
```
wxmerkt, line 29: It would be good to add some comments what kind of MultiCopter actuation model this is - e.g. a reference to a paper or at least inline code comments on how it works, what the assumptions and notation are, etc. 
[E.g. the Mellinger model is fairly common so if this were one of that kind, a reference would be sufficient with notes of modifications]

PepMS: I agree in explaining a little bit the model. I think that in the Mellinger's model the control inputs are considered to be the net force and moments applied to the body due to the force of the different propellers., i.e. `data->tau`.
Here, instead, we consider the force of each propeller.

wxmerkt: Thank you @PepMS, this is really helpful for anybody looking at the code at a later stage :-)

cmastalli: @PepMS and @wxmerkt I didn't notice this thread. I just merged this branch.

@PepMS could you create another PR tackling this issue?

PepMartiSaumell: Sure!

PR #751, opened 2020-04-09
Understanding the relinearization in differental manifolds (DO NOT MERGE)

cmastalli:
This PR intends to open a discussion regarding the relinearization perform in the DDP backward pass.
As discussed with @nmansard and @jcarpent, it seems to me that we need to apply a parallel transport of the `Vxx` during the backward relinearization (https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solvers/ddp.cpp#L182). Concretely speaking, I suspect that `Vxx_next` lies in a different tangent space to `Vx`.

@nmansard I commented the lines that are not passing. You could run it in your PC like this:
```cpp
cp ${crocoddyl_path}/unittest/python/
python test_solvers.py
```

Note that this is legacy code which we never modified.

### DIFF
--- a/unittest/python/test_solvers.py
+++ b/unittest/python/test_solvers.py
@@ -103,7 +103,7 @@ class SolverKKTTest(unittest.TestCase):
 # ---------------------------------------------------
 NX = 1
 NU = 1
-model = ActionModelUnicycle()
+model = ActionModelUnicycleVar()
 data = model.createData()
 LQR = isinstance(model, ActionModelLQR)
 
@@ -129,7 +129,8 @@ xs = [m.State.rand() for m in problem.runningModels + [problem.terminalModel]]
 us = [np.random.rand(m.nu) for m in problem.runningModels]
 kkt.setCandidate(xs, us)
 dxs, dus, ls = kkt.computeDirection()
-assert (np.linalg.norm(xs[0] + dxs[0] - problem.initialState) < 1e-9)
+print "Initial condition error: ", np.linalg.norm(model.State.integrate(xs[0], dxs[0]) - problem.initialState)
+#assert (np.linalg.norm(model.State.integrate(xs[0], dxs[0]) - problem.initialState) < 1e-9) #### This fails
 for i, _ in enumerate(dus):
     if LQR:
         assert (np.linalg.norm(model.calc(data, xs[i] + dxs[i], us[i] + dus[i])[0] - (xs[i + 1] + dxs[i + 1])) < 1e-9)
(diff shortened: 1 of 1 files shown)


cmastalli:
@jmirabel you might want to give your feedback.

jmirabel:
@cmastalli is it really my review you asked for ? I believe you made a mistake. If no, I need a bit of context.

cmastalli:
> @cmastalli is it really my review you asked for ? I believe you made a mistake. If no, I need a bit of context.

@jmirabel I want to confirm with you that the parallel transport in Lie algebra has the form:



where   transport a tangent vector  to the tangent space ,  is the configuration manifold, and  is the operator Jacobian computed using Pinocchio.

jcarpent:
This Jint that as to be used here.
More precisely, using Pinocchio's convention and code, what you're looking for:
```
Jint = dIntegrate(model,q1,difference(model,q1,q2),ARG0)
```
This operator transports the tangent space of T_q2 to the tangent space of T_q1.

jmirabel:
What is the operation you differentiate ?

cmastalli:
> This Jint that as to be used here.
> More precisely, using Pinocchio's convention and code, what you're looking for:
> 
> ```
> Jint = dIntegrate(model,q1,difference(model,q1,q2),ARG0)
> ```
> 
> This operator transports the tangent space of T_q2 to the tangent space of T_q1.

Indeed, my notation is equivalent to this Pinocchio code. When I wrote it, I used the Crocoddyl notation.



> What is the operation you differentiate ?
It's is the diff operator. @jcarpent wrote a code that retrieve that one from the Jacobian of the integrator.

jcarpent:
Jdiff is equivalent to dDifference ;)

jcarpent:
@cmastalli, you have to use the dIntegrate instead, giving:
`v_1 = Jint * v2`

proyan:
@cmastalli Could I close this PR?

cmastalli:
> @cmastalli Could I close this PR?

No, don't do it. I need to finish this work.

proyan:
Since there hasn't been any activity at all on this PR for almost a month and a half, could you instead open another PR when it is ready?

proyan:
@cmastalli 
Could you please close this PR? I think that it has served its purpose.

cmastalli:
>  Could you please close this PR? I think that it has served its purpose.

Let's close it because I don't have time.
@proyan note that we still need to investigate if it is needed to apply the parallel transport in the backward pass.

PR #759, opened 2020-04-21
Using active contact and impulses in actions + other minor improvements

cmastalli:
This PR parses the current active set of contacts (or impulses) to carry computations on the contact (or impulse) dynamics. This was the last piece of cake needed to follow the idea of global allocation of data.

Additionally, there are other general improvements:
 - sorted the active cost, contact and impulse vectors for easier reading.
 - used numpy array inside gepetto display for efficiency (and numpy matrix is tagged as deprecated)
 - created a display abstract for including other display methods
 - added meshcat displayer and integrated in couple of Jupyter notebooks (it needs to include force, frame trajectory and friction cone display).
 - used numpy array in all examples and notebooks.

### DIFF
--- a/include/crocoddyl/multibody/actions/contact-fwddyn.hxx
+++ b/include/crocoddyl/multibody/actions/contact-fwddyn.hxx
@@ -66,6 +66,7 @@ void DifferentialActionModelContactFwdDynamicsTpl<Scalar>::calc(
                  << "u has wrong dimension (it should be " + std::to_string(nu_) + ")");
   }
 
+  const std::size_t& nc = contacts_->get_nc();
   DifferentialActionDataContactFwdDynamicsTpl<Scalar>* d =
       static_cast<DifferentialActionDataContactFwdDynamicsTpl<Scalar>*>(data.get());
   const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> q = x.head(state_->get_nq());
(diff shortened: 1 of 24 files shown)


include/crocoddyl/multibody/actions/contact-fwddyn.hxx
```diff
@@ -82,15 +83,16 @@ void DifferentialActionModelContactFwdDynamicsTpl<Scalar>::calc(
   contacts_->calc(d->multibody.contacts, x);
 
 #ifndef NDEBUG
-  Eigen::FullPivLU<MatrixXs> Jc_lu(d->multibody.contacts->Jc);
+  Eigen::FullPivLU<MatrixXs> Jc_lu(d->multibody.contacts->Jc.topRows(nc));
```
proyan, line 86: Isn't the size of Jc already nc?

cmastalli: I have introduced, in a previous PR, the concept on active and inactive contacts. `nc_total` defines the dimension of total number of contacts, instead `nc` defines the dimension of active contacts. The same has been done for impulses and costs.

The idea is to allow user to allocate the memory only once. For instance, you can create a contact dynamics with all contact consider in your MPC, and then online to define which of them are active.

Saying that, the answer of your question is no. `Jc` is defined with `nc_total` but we fill the first `nc` rows (active contacts).

Let me know if you have more questions regarding this topic. Btw, I have tested in my MPC formulation. Later I will make it public.

cmastalli:
I will proceed with this PR.

PR #761, opened 2020-04-22
[CMake] fix target_link_libraries use

nim65s:
For recent CMake versions. Otherwise, we get errors like:

```
CMake Error at CMakeLists.txt:127 (TARGET_LINK_LIBRARIES):
  The keyword signature for target_link_libraries has already been used with
  the target "crocoddyl".  All uses of target_link_libraries with a target
  must be either all-keyword or all-plain.
```

And as `TARGET_LINK_BOOST_PYTHON` doesn't use a  keyword, we can only use the "all-plain" alternative for now.


proyan:
Thanks Guilhem.
So the permanent fix has to be in jrl-cmakeModules?

proyan:
Also, which version of CMake are you using?

nim65s:
yes, ultimately, `TARGET_LINK_BOOST_PYTHON` should let the user use the keyword or the plain signature.

I get this error on ArchLinux / CMake 3.17.1 and Ubuntu 20.04 / 3.16.3.

cmastalli:
@nim65s thank you for this contribution.

jcarpent:
TARGET_LINK_BOOST_PYTHON supports the keyword!!!

nim65s:
Reverting would not work, but we can switch to all-keyword. there are 10 `TARGET_LINK_LIBRARIES` around.

May I ask why you need the version with keywords ?

jcarpent:
It depends in fact if one's is also using a Python interpreter in its application or not.

nim65s:
Do you mean that choosing between all-plain or all-keyword in Crocoddyl's CMakeLists.txt changes anything in a package that would depend on crocoddyl and embed a python interpreter ?

jcarpent:
Yep.

nim65s:
Ok, with the cmake exports, it totally make sense, I just didn't experience any issue with this yet.

So to fix this, I would need to know exactly what is part of the public API and what is not, in all links, but I have no idea to do this unless digging in the whole codebase.

Another option would be to just put everything public, and make things private if they cause any trouble to anybody.

cmastalli:
@nim65s I encountered a problem with this modification when I tried to compile Crocoddyl with multithreading support. This is the message:
```bash
CMake Error at CMakeLists.txt:116 (TARGET_LINK_LIBRARIES):
  The keyword signature for target_link_libraries has already been used with
  the target "crocoddyl".  All uses of target_link_libraries with a target
  must be either all-keyword or all-plain.

  The uses of the keyword signature are here:

   * CMakeLists.txt:111 (TARGET_LINK_LIBRARIES)
   * CMakeLists.txt:112 (TARGET_LINK_LIBRARIES)
```
and the lines 111 and 112 are:
```
  TARGET_LINK_LIBRARIES(${PROJECT_NAME} PUBLIC pinocchio::pinocchio)
  TARGET_LINK_LIBRARIES(${PROJECT_NAME} PRIVATE ${Boost_FILESYSTEM_LIBRARY} ${Boost_SYSTEM_LIBRARY}
      ${Boost_SERIALIZATION_LIBRARY})
```

Could I revert these modifications (only regarding line 111 and 112)? Or do you want to handle this by your own?

nim65s:
Maybe I'm not fully awaken yet, but I think this PR is fixing the issue you are describing.

You say your likes contain PUBLIC and PRIVATE, and the whole point of this PR is to remove that to fix the error you have.

cmastalli:
> Maybe I'm not fully awaken yet, but I think this PR is fixing the issue you are describing.
> 
> You say your lines contain PUBLIC and PRIVATE, and the whole point of this PR is to remove that to fix the error you have.

Yes, those lines are creating a compilation issue. I remove `PUBLIC` and `PRIVATE` locally and Crocoddyl compiles.

nim65s:
Instead of removing PUBLIC and PRIVATE locally, you can APPLY this PR, by rebasing your local branch on devel, or merging devel on it, or cherry-picking the commit e4e02f1

cmastalli:
> Instead of removing PUBLIC and PRIVATE locally, you can apply this PR, eg. by rebasing your local branch on devel, or merging devel on it, or cherry-picking the commit [e4e02f1](https://github.com/loco-3d/crocoddyl/commit/e4e02f13bba5f7096a7ae7e16b71bf5b40e4963a)

Sorry I got confused, I confirm you that this PR fixes this things (as you told me). Thank you so much :)

Issue #291, opened 2019-11-18
Currently pinocchio::impulseDynamics allocates internally memory, do not use it for multithreading


wxmerkt:
In GitLab by @cmastalli on Nov 9, 2019, 11:14

I have been working on #271, and I noticed that part of the non-deterministic behavior of the solver is that `pinoccho::impulseDynamics` allocates internally memory, and I reported it few weeks ago: https://github.com/stack-of-tasks/pinocchio/issues/881.

@wxmerkt and @proyan be aware of it, and do not waste your time.

For more elaborate discussion, I suggest to use the Pinocchio issue.

wxmerkt:
In GitLab by @cmastalli on Nov 9, 2019, 11:15

changed title from **Currently {-impulse d-}ynamics creates interally memory, do not use for multithreading** to **Currently {+pinocchio::impulseD+}ynamics creates interally memory, do not use for multithreading**

wxmerkt:
In GitLab by @cmastalli on Nov 9, 2019, 11:16

changed title from **Currently pinocchio::impulseDynamics {-creates interally memory, do not use-} for multithreading** to **Currently pinocchio::impulseDynamics {+allocates internally memory, do not use it+} for multithreading**

wxmerkt:
In GitLab by @cmastalli on Nov 9, 2019, 11:16

changed the description

wxmerkt:
In GitLab by @wxmerkt on Nov 9, 2019, 11:22

Thank you for the note, I wasn't aware of the upstream issue. It did already waste quite a lot of time.

wxmerkt:
In GitLab by @proyan on Nov 9, 2019, 18:14

Cool! Good work Carlos

cmastalli:
solved in #765

Issue #743, opened 2020-04-02
Numerical impact of changes to BoxQP


wxmerkt:
In https://github.com/loco-3d/crocoddyl/pull/658, the BoxQP was changed substantially.

From the PR description:
> One of the main problem was due to a bug in the box QP, incorrect output of the active set. 

Can you please elaborate and provide an example / line numbers from the old solver?

> Additionally, our box QP is able to accept infeasible warm-start in contrast to the original proposed method by Tassa, this is done by a "cheap" projection to the feasible set. 

Correct, this step was omitted and is new.

> In conclusion, this PR proposes a completely new box QP solver, which is based on Cholesky factorization.

The old solver relied on the general inverse (`.inverse()`), which in our benchmarks wasn't too much of an overhead. In fact, with compiler optimisations I sometimes get opposite results (faster convergence using the old).

Unfortunately, no test cases or any further detailed description was included.

From a side-by-side inspection, two notable changes to the originally introduced BoxQP exist:

1) Change of the L2 norm to the infinity norm for gradient tolerance check. This has no influence on convergence or numerical behaviour (verified just now).
2) Change from a linear, backtracking line search (alpha=1.0, 0.9, ... 0.1) to an exponential line search. This has a large numerical impact (see below).

Additionally, I believe the analysis that there was a bug in the active set is correct. I have not been able to pinpoint it down _where_ yet. 

To figure out the different numerical behaviours, I am testing all different combinations on a quadcopter task (in Exotica "example_dynamic_time_indexed 07_..."). The only element that is being changed are the changes to the BoxQP.

![image](https://user-images.githubusercontent.com/1664508/78254301-9f7c0e00-74ed-11ea-8b91-76f3a482b371.png)

This plot shows:

1. The original BoxQP in blue. The convergence is identical to orange, which corresponds to changing the L2 to the Infinity norm for gradient tolerance checks.
2. The new BoxQP in purple. Convergence is initially faster, but then becomes linear and is not as good as the original BoxQP.
3. The original BoxQP, but changed to an exponential line search. Clearly, the performance is worse. This may be related to the the bug you identified but without further detail on where it was / how it manifested it's really hard to evaluate what's going on.

Any insight is highly appreciated

wxmerkt:
I found one bug in the old active-set for the special case that it converged during the first iteration. In this case `free_idx` was not set correctly, which lead to the feedback matrix `K` being incorrect. However, it does not have an effect on convergence (coincides with the orange line).

[...]

PR #765, opened 2020-04-29
Enable multithreading apps with impulse dynamics #291

cmastalli:
The current code was setting and resetting the gravity vector of the Pinocchio model. This clearly creates an issue in multithreading applications.

Instead the PR follows the approach to compute the `dgrad_dq` and cancelling from the RNEA derivatives computation. Note that this method is equivalent, with the collateral effect of computing twice `dtau_dq`.

This approach was suggested by @jcarpent here: https://github.com/stack-of-tasks/pinocchio/issues/881#issuecomment-621047646.

PD: The tiny numerical difference produces a bit difference log files.


examples/log/humanoid_taichi.log
```diff
@@ -156,5 +156,5 @@ iter 	 cost 	      stop 	    grad 	  xreg 	      ureg 	 step 	 feas
  140  6.81843e+00  6.62800e-08  1.57323e-07  1.00000e-09  1.00000e-09   0.5000     1
 ('Finally reached = ', array([3.99963367e-01, 2.37599941e-05, 1.19998324e+00]))
 ('Distance between hand and target = ', 4.676946718296189e-05)
-('Distance to default state = ', 32.524482279937715)
```
proyan, line 159: This is a bit weird. `Distance = 32.5`, and it is changed to `Distance = 1.6`? Were the logs wrong before?

examples/log/boxfddp_vs_boxddp.log
```diff
@@ -23,68 +23,68 @@ iter 	 cost 	      stop 	    grad 	  xreg 	      ureg 	 step 	 feas
   19  3.09833e+05  1.20330e+08  1.91734e+06  1.00000e-01  1.00000e-01   0.2500     0
 iter 	 cost 	      stop 	    grad 	  xreg 	      ureg 	 step 	 feas
   20  2.34230e+05  1.70003e+08  1.14113e+06  1.00000e-01  1.00000e-01   0.5000     0
-  21  1.71141e+05  2.00479e+08  5.90291e+05  1.00000e-01  1.00000e-01   0.2500     0
```
proyan, line 26: There is something in the way that we are using the box solvers that is causing this difference in numerical behaviour. The difference in cost here is quite high (order of 1.). This is in the same vein as the difference because of different architectures, as reported in https://github.com/loco-3d/crocoddyl/pull/696.
I'm inclined to say it is because of `box-qp::Hff_inv_llt_` , which is an Eigen::LLT object. But this is only my guess.

examples/log/humanoid_taichi.log
cmastalli: This is quiet weird indeed. I saw the same thing with codegen case. Do you think we should investigate this before merging it?

examples/log/boxfddp_vs_boxddp.log
cmastalli: You might be right. We were discussing about it.
@wxmerkt you might want to be involved in this comment.

wxmerkt: I have seen some issues with LLT, and particularly the pre-allocated Hff_inv_llt_ (`Hff.llt().solve()` is fine) in _some_ cases on a wider benchmark. Changing to a different decomposition solved it indeed.

cmastalli: I think this is an interesting topic, but I would prefer to handle it in a dedicate PR.

We need to evaluate carefully the proposition of using LDLT. Note that its limitation is the computational speed (LLT is faster). I have conducted a benchmark, you could  check more information here: https://github.com/loco-3d/crocoddyl/issues/743#issuecomment-621667401.

Let's continue this discussion in that thread :)

examples/log/humanoid_taichi.log
cmastalli: I cannot explain you the reason of this difference, however, I can tell you that this, and other humanoid examples, does not use the impulse dynamics. Therefore, the changes propose in this PR should not affect them.

Additionally, I check qualitatively the resulted motion using Gepetto viewer and it seems to be the same.

cmastalli:
thank you for the feedback.

PR #739, opened 2020-04-01
WIP: Release v1.2.0

nim65s:
Hi,

This PR is here to track the work on #731 
Please don't merge it :)

PR #766, opened 2020-04-30
[Cmake] use project exports

nim65s:
I totally forgot that in  #739, sorry.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ ENDIF()
 
 # Disable -Werror on Unix for now.
 SET(CXX_DISABLE_WERROR True)
+SET(PROJECT_USE_CMAKE_EXPORT True)
 
 # Create different building options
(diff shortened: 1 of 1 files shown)


proyan:
Does this export automatically the definitions that we use in crocoddyl? Do we still need to do `SET(PACKAGE_EXTRA_MACROS ...)`?

nim65s:
This doesn't export all definitions by default. This is required to create the `$CMAKE_INSTALL_PREFIX/lib/cmake/crocoddyl/*.cmake` files. They will include the `crocoddyl::crocoddyl` target, thanks to the  `EXPORT ${TARGETS_EXPORT_NAME}` in the `INSTALL` command, and all dependenciess that come from `ADD_PROJECT_DEPENDENCY`, but that's it, unless you add more stuff in this generated file with `SET(PACKAGE_EXTRA_MACROS …)`.

proyan:
Thanks Guilhem

PR #767, opened 2020-05-04
[Python] fix wrong hardcoded windowID for gepetto gui

nim65s:
found by @olivier-stasse : https://github.com/Gepetto/gepetto-viewer-corba/issues/144


nim65s:
A grep for `(0,` seems to indicate that this is the only occurence where this hardcoded value was used in this way.

Issue #255, opened 2019-11-18
Build with multi-threads OFF and vectorisation ON results in double free corruption


wxmerkt:
When building with multithreads OFF and vectorisation ON, the benchmarks segfault with a double free corruption.

The following is tested by running talos_legs:

*  Multithreads OFF, vectorisation OFF: OK (8.89 ms)
*  Multithreads ON, vectorisation OFF: OK (5.05 ms)
*  Multithreads ON, vectorisation ON: OK (5.13 ms)
*  **Multithreads OFF, vectorisation ON: double free or corruption (out)**

This is the backtrace for a crash:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff7a24801 in __GI_abort () at abort.c:79
#2  0x00007ffff7a6d897 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff7b9ab9a "%s\n")
    at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007ffff7a7490a in malloc_printerr (str=str@entry=0x7ffff7b9c870 "double free or corruption (out)")
    at malloc.c:5350
#4  0x00007ffff7a7be75 in _int_free (have_lock=0, p=0x555555ba2a70, av=0x7ffff7dcfc40 <main_arena>) at malloc.c:4278
#5  __GI___libc_free (mem=0x555555ba2a80) at malloc.c:3124
#6  0x00007fffdbe32543 in boost::python::objects::caller_py_function_impl<boost::python::detail::caller<boost::python::detail::member<Eigen::Matrix<double, -1, 1, 0, -1, 1>, pinocchio::ModelTpl<double, 0, pinocchio::JointCollectionDefaultTpl> >, boost::python::return_value_policy<boost::python::return_by_value, boost::python::default_call_policies>, boost::mpl::vector3<void, pinocchio::ModelTpl<double, 0, pinocchio::JointCollectionDefaultTpl>&, Eigen::Matrix<double, -1, 1, 0, -1, 1> const&> > >::operator()(_object*, _object*) ()
   from /opt/openrobots/lib/python2.7/site-packages/pinocchio/libpinocchio_pywrap.so
#7  0x00007fffded76e55 in boost::python::objects::function::call(_object*, _object*) const ()
   from /usr/lib/x86_64-linux-gnu/libboost_python-py27.so.1.65.1
[...]

PR #768, opened 2020-05-05
Support of data alignment + fixed Scalar typecast of few doubles

cmastalli:
This PR brings data alignment support to Crocoddyl (issue #255). As you might know this is an important feature for boosting the computation on your own hardware.

The main problem was with the data create. We use shared pointers to spawn data in each model. The data creation was highly depended on `make_shared`, however it doesn't honor the data alignment request. To enable this request, we need to use `allocate_shared` with the `Eigen::aligned_allocator`.

Additionally, I have reviewed entirely the code to add proper typecasting of `double` to `Scalar`. @proyan you might want to review these changes carefully.


unittest/test_boxqp.cpp
```diff
@@ -31,7 +31,7 @@ void test_unconstrained_qp_with_identity_hessian() {
   Eigen::VectorXd gradient = Eigen::VectorXd::Random(nx);
   Eigen::VectorXd lb = -std::numeric_limits<double>::infinity() * Eigen::VectorXd::Ones(nx);
   Eigen::VectorXd ub = std::numeric_limits<double>::infinity() * Eigen::VectorXd::Ones(nx);
-  Eigen::VectorXd xinit(nx);
+  Eigen::VectorXd xinit = Eigen::VectorXd::Random(nx);
```
wxmerkt, line 34: It would have been good to highlight this change in the PR summary... looking over 96 files takes a while and this is quite an important fix to the unit test.

cmastalli: The main changes are described in the description, I believe.

This modification provides a proper method to initialize the vector. Note that before we're not initializing the values.

wxmerkt: Yes, I understand the changes well. The summary should serve as a summary though in case someone does not read every PR in full length (it's a lot of files...). This is an important catch-and-fix of a bug of a unit test that is likely to not have tested something useful before, and now works well. So it would be good to include these notes in a summary :-).

include/crocoddyl/core/actions/lqr.hxx
```diff
@@ -44,7 +44,7 @@ void ActionModelLQRTpl<Scalar>::calc(const boost::shared_ptr<ActionDataAbstract>
   } else {
     data->xnext.noalias() = Fx_ * x + Fu_ * u + f0_;
   }
-  data->cost = 0.5 * x.dot(Lxx_ * x) + 0.5 * u.dot(Luu_ * u) + x.dot(Lxu_ * u) + lx_.dot(x) + lu_.dot(u);
+  data->cost = Scalar(0.5) * x.dot(Lxx_ * x) + 0.5 * u.dot(Luu_ * u) + x.dot(Lxu_ * u) + lx_.dot(x) + lu_.dot(u);
```
proyan, line 47: You missed the scalar casting here.

unittest/test_boxqp.cpp
cmastalli: I didn't think it was significant to point out this specific modification.

I will try to do my best in the next time :)

It would be great if you can try these modifications yourself. Remember that you need to match the same compilation flag (i.e. `-march=native`) in each dependency of Crocoddyl that dependents on Eigen.

include/crocoddyl/core/actions/lqr.hxx
cmastalli: Nice catch! I will create another PR tomorrow with all your annotations :)

include/crocoddyl/core/solvers/box-qp.hpp
```diff
@@ -29,10 +31,10 @@ struct BoxQPSolution {
   std::vector<size_t> clamped_idx;
 };
 
-// Based on Yuval Tassa's BoxQP
```
proyan, line 32: Why did you remove the doc from here? I found it useful to have this doc to understand what was being developed!

cmastalli: @wxmerkt implemented using this code. Then I re-wrote it from scratch without checking or reviewing that code. I had to removed earlier, but I forgot it.

cmastalli: I could add citation to Tassa paper (or better the Projected Newton QP paper that Tassa is citing)

cmastalli: @wxmerkt what is your opinion about it?

I removed it because this is not a copy-translation of Tassa's code, and it seems to me that we should remove this comment.

wxmerkt: I compared the implementations a few weeks back and there is little difference: polynomial vs linear line-search (makes no difference in convergence) and LpNorm vs L2Norm for convergence check (no difference in convergence). I think there is a minor change in the Armijo criterion for accepting a step. And of course there's the change from inverse() to LLT. 

We should definitely include a reference (also in general add more comments to code - it's often quite an effort to read Crocoddyl code and follow what's going on) - if you worked off the Projected Newton QP cited in Tassa, that should be included. Otherwise we should leave the Tassa reference.

cmastalli: I have included a preliminary doxygen documentation of this class. See my PR.

include/crocoddyl/core/actions/lqr.hxx
cmastalli: Solved this issue and other in diff-lqr in my ongoing PR.

PR #769, opened 2020-05-06
Minor changes regarding Scalar + Added simple doxygen doc

cmastalli:
This PR includes few extra Scalar type casting plus the first actions regarding the doxygen documentation. The Doxygen documentation includes:
- A description of the Crocoddyl with example and citation reference.
- Initial documentation of `Box-QP`

@nim65s could you review the doxygen documentation?
@proyan and @wxmerkt could you help me with the documentation of abstract classes?


proyan:
I'll look at the PR tomorrow.

Just looking at the list of files, why are you updating again the log 
files in this PR?

examples/log/boxfddp_vs_boxddp.log
```diff
@@ -17,74 +17,74 @@ iter 	 cost 	      stop 	    grad 	  xreg 	      ureg 	 step 	 feas
...
+  20  2.17520e+06  1.24754e+09  6.09848e+06  1.00000e+00  1.00000e+00   0.0625     1
+  21  1.79919e+06  5.33754e+08  3.76591e+06  1.00000e+00  1.00000e+00   0.1250     1
+  22  1.13695e+06  9.05971e+08  2.59785e+06  1.00000e+00  1.00000e+00   0.5000     1
+  23  8.30290e+05  1.41761e+08  1.93369e+06  1.00000e+00  1.00000e+00   0.5000     1
+  24  5.77495e+05  6.77960e+07  1.13559e+06  1.00000e+00  1.00000e+00   0.2500     1
+  25  4.27336e+05  2.77789e+07  4.46809e+05  1.00000e+00  1.00000e+00   0.5000     1
+  26  3.91060e+05  2.39396e+07  2.19975e+05  1.00000e+00  1.00000e+00   0.2500     1
+  27  3.72508e+05  1.91335e+07  1.63448e+05  1.00000e+00  1.00000e+00   0.1250     1
+  28  3.45084e+05  1.43382e+07  1.30711e+05  1.00000e+00  1.00000e+00   0.5000     1
+  29  3.27129e+05  1.71809e+07  1.18675e+05  1.00000e+00  1.00000e+00   0.5000     1
```
wxmerkt, line 90: The other log file changes are minor. This is quite significant. Any insights into why it disimproved this much? I.e. the accepted step lengths are quite a few steps smaller than before

cmastalli: I forgot to mention this, thank you for bringing the topic.

There are two changes that produces different log files, none of them coming from this PR. The first one was regarding the changes of the impulse dynamics, and I corrected it. I did the same with the changes of the codegen in the @proyan branch. The former action was merged first. The second was done later and it has to be resolved conflict which values that doesn't reflect both changes.

I am constantly changing these values, that's why I have a perfect track of it. However we really need an unittest for this.

Btw, those values are for the Box-DDP, which doesn't solve the problem.

wxmerkt: I understand those are the values for the regular Box-DDP; what we see here is that it gets a lot more difficult for a regular Box-DDP to solve this problem: So it's either a disimprovement of the Box-DDP in one of the recent PRs or that the problem is much harder to solve? From following the Codegen PR, I did not see any changes that directly explain this numerical change - but I didn't read it too closely. Do we know this log file change is then due to the impulse dynamics update? Or another recent update and we forgot to look into the logs?

cmastalli: In that PR, @proyan has improved the efficiency of many matrix multiplications (e.g.in the activation models and euler integrator). They produce tiny numerical differences.

Tomorrow I will tell you what happens precisely. I don't remember now what of the two changes (impulse or codegen) is the source.

cmastalli:
I have added quite descriptive documentation of the state abstraction and improved the QP one.

I won't continue to work on this branch since I need to move to other tasks.

If you want to contribute in the documentation of the classes. Please let me know and add the task in the Github.
Our first goal should be to document the abstract classes and solvers.

examples/log/boxfddp_vs_boxddp.log
proyan: Why was this marked resolved? Did you find the source of these changes to the log?

proyan:
I haven't followed the changes in the doc, I trust you on that. 

The changes in box-ddp are quite concerning to me. There is an open comparison from February #696, and this issue seems to have persisted. We should move this to a priority.

examples/log/boxfddp_vs_boxddp.log
cmastalli: > Why was this marked resolved?

My mistake, I didn't notice that my message were not posted (my home internet is not that great)




>  Did you find the source of these changes to the log?

Yeap. You merged the impulse dynamics into your `codegen2` branch. It happens after my modification of the log files in this branch. Therefore, the impulse dynamics changes overwrote the ones relate to the codegen.

Unfortunately, we forgot to re-do the log file generation after that. To be honest, I was aware of it, but those days very busy. Additionally, I didn't want you to ask it, because it could produce other differences relates to your hardware.

Let say that for the moment this is my own unittest, and I am not sure if we could get robust values to be tested in the CI.

cmastalli: @nim65s and I do not know why there are differences between hardware, or probably software versions. Who knows?

cmastalli:
> The changes in box-ddp are quite concerning to me. There is an open comparison from February #696, and this issue seems to have persisted. We should move this to a priority.

I agree, it is important to understand this phenomena. @wxmerkt did some work, and I hope we can figure out soon.

Note that I have tried the LDLT solver in both: QP and Box-QP solver as described in https://github.com/loco-3d/crocoddyl/issues/743#issuecomment-621713612. I suggest to keep this discussion in that thread :)

wxmerkt:
Just as a note, before moving further discussion to #743, my evaluations where between the two different BoxQP and the open variables/differences in their implementations - and how that affects BoxDDP and BoxFDDP. However, this analysis was with all else being equal. 

#696 was _after_ the new QP was added, so shouldn't have shown a difference between January when the QP was added and logs updated, and February when Rohan opened the issue. 

My analysis showed differences between the QPs, but those would not account for something that we see here in this PR (since the QP hasn't been changed since January).

The disimprovement shown in #696 is significant (on the fifth decimal place or so), but within the same order of magnitude. The disimprovement in performance in the logs included in this PR is ~2x.

PR #772, opened 2020-05-15
Functionalities needed for cyclic MPC formulations

cmastalli:
This PR proposes to use `boost::circular_buffer` to store action models and datas in the shooting problem class. It simplifies the cyclic MPC formulation when the user might need to update the first action model and move on the remaining models. It could be easily do it through a `push_back` method implementing in this PR as well (i.e. `pushBackRunningNode`). I have bench the use of `boost::circular_buffer` against `std::vectors`, and the computation time remains the same.

Additionally, this PR includes the `updateNode` function and extra constructor that allows user to pass the datas. It is quite useful since users can have the control of data allocation. Again it is very important for any MPC application.

Finally, the PR includes a `checkData` method for all action models (and diff-action models) to allow the shooting problem does data checking which protects us against user mistakes. These new functionality is tested in our unittest!

**Important note**
I want to include few other things here:
 - a `pushBackRunningModel` function similar to `pushBackRunningNode` but we internal data allocation.
 - doxygen documentation of all the shooting-problem methods.

However, I wanted to share the code for discussion: @proyan and @nmansard.

### DIFF
--- a/benchmark/arm-manipulation-codegen.cpp
+++ b/benchmark/arm-manipulation-codegen.cpp
@@ -52,8 +52,8 @@ int main(int argc, char* argv[]) {
   x0 << q0, Eigen::VectorXd::Random(state->get_nv());
 
   // Defining the shooting problem for both cases: with and without code generation
-  std::vector<boost::shared_ptr<crocoddyl::ActionModelAbstract> > runningModels(N, runningModel);
-  std::vector<boost::shared_ptr<crocoddyl::ActionModelAbstract> > cg_runningModels(N, cg_runningModel);
+  boost::circular_buffer<boost::shared_ptr<crocoddyl::ActionModelAbstract> > runningModels(N, runningModel);
+  boost::circular_buffer<boost::shared_ptr<crocoddyl::ActionModelAbstract> > cg_runningModels(N, cg_runningModel);
   boost::shared_ptr<crocoddyl::ShootingProblem> problem =
       boost::make_shared<crocoddyl::ShootingProblem>(x0, runningModels, terminalModel);
   boost::shared_ptr<crocoddyl::ShootingProblem> cg_problem =
--- a/benchmark/arm-manipulation-optctrl.cpp
+++ b/benchmark/arm-manipulation-optctrl.cpp
@@ -35,7 +35,7 @@ int main(int argc, char* argv[]) {
 
   // For this optimal control problem, we define 100 knots (or running action
   // models) plus a terminal knot
-  std::vector<boost::shared_ptr<crocoddyl::ActionModelAbstract> > runningModels(N, runningModel);
+  boost::circular_buffer<boost::shared_ptr<crocoddyl::ActionModelAbstract> > runningModels(N, runningModel);
   boost::shared_ptr<crocoddyl::ShootingProblem> problem =
       boost::make_shared<crocoddyl::ShootingProblem>(x0, runningModels, terminalModel);
   std::vector<Eigen::VectorXd> xs(N + 1, x0);
--- a/benchmark/arm-manipulation-timings.cpp
+++ b/benchmark/arm-manipulation-timings.cpp
@@ -99,7 +99,7 @@ int main(int argc, char* argv[]) {
   boost::shared_ptr<crocoddyl::ActionModelAbstract> terminalModel =
       boost::make_shared<crocoddyl::IntegratedActionModelEuler>(terminalDAM, 1e-3);
 
-  std::vector<boost::shared_ptr<crocoddyl::ActionModelAbstract> > runningModels(N, runningModel);
+  boost::circular_buffer<boost::shared_ptr<crocoddyl::ActionModelAbstract> > runningModels(N, runningModel);
 
   boost::shared_ptr<crocoddyl::ShootingProblem> problem =
       boost::make_shared<crocoddyl::ShootingProblem>(x0, runningModels, terminalModel);
--- a/benchmark/bipedal-timings.cpp
+++ b/benchmark/bipedal-timings.cpp
@@ -119,7 +119,7 @@ int main(int argc, char* argv[]) {
   boost::shared_ptr<crocoddyl::ActionModelAbstract> terminalModel =
       boost::make_shared<crocoddyl::IntegratedActionModelEuler>(terminalDAM, 1e-3);
 
-  std::vector<boost::shared_ptr<crocoddyl::ActionModelAbstract> > runningModels(N, runningModel);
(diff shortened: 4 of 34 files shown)


cmastalli:
I have benched the computation time of using `std::vector` against `boost::circular_buffer`. For the following test, I have defined 108 nodes using the quadrupedal action models. These are the results:
 - `std::vector`: 4.7865 usecs
 - `boost::circular_buffer`: 0.6916 usecs

Clearly, the `boost::circular_buffer` is better since there is no needed extra copy of the action model items, and there is no memory overhead.

On the other hand, `boost::circular_buffer` introduces an API change, which it is not hard to deal for the users. The question is "is this result significance for making an API change?" @proyan, @nmansard and @jcarpent.

If you want to inspect the details of the `std::vector` approach, please see this code: https://github.com/cmastalli/crocoddyl/blob/topic/shooting-extra/include/crocoddyl/core/optctrl/shooting.hxx#L175

wxmerkt:
@cmastalli Can you please add some detail on the test that these numbers are represented? E.g. is this using a `reserve` with the vector, what is the change that is timed - pop_front and push_back - or something else? It is really unclear what the timed code is.

cmastalli:
> @cmastalli Can you please add some detail on the test that these numbers are represented? E.g. is this using a `reserve` with the vector, what is the change that is timed - pop_front and push_back - or something else? It is really unclear what the timed code is.

It doesn't use `reserve`, it uses `push_back` plus copying data. See more details: https://github.com/loco-3d/crocoddyl/pull/772#issuecomment-629380073.

I also though to use `insert` and then `pop_front`. Do you think is this better approach? I could bench it as well.

cmastalli:
@wxmerkt or do you have a better way to code it? I could bench it as well.

jcarpent:
> I have benched the computation time of using `std::vector` against `boost::circular_buffer`. For the following test, I have defined 108 nodes using the quadrupedal action models.

Could you provide the bench you did?

wxmerkt:
I assume most of the time is due to allocating `copy_models` and `copy_data`. I don't think you actually need the copy: At i=0, you are assigning 1 to location 0, at i=1, you are assigning 2 to 1... why do you need a copy for that? As long as this for-loop is _sequential_ and not parallel, you don't need `copy_data`, `copy_model` at all. Another variant would be to switch `vector` to `deque`.

On another node, it wasn't immediately obvious to me from the name `pushBackRunningNode` that it uses a fixed size, drops the first, and pushes back at the end (as in MPC). The name could also suggest that the vector is appended by one. Alas, I don't have a suggestion for a better name right now.

cmastalli:
> Could you provide the bench you did?

I run the quadrupedal_gait problem, and then I run the following script:
```python
import time
problem = ddp[0].problem
model = problem.runningModels[0]
data = problem.runningDatas[0]
start = time.time()
T = int(1e5)
for i in range(T):
  problem.pushBackRunningNode(model, data)
end = time.time()
print 1e3 * (end - start) / T
```

I also have a much better PC that the ones that we could get on a robot.

cmastalli:
> I assume most of the time is due to allocating `copy_models` and `copy_data`. I don't think you actually need the copy: At i=0, you are assigning 1 to location 0, at i=1, you are assigning 2 to 1... why do you need a copy for that? As long as this for-loop is _sequential_ and not parallel, you don't need `copy_data`, `copy_model` at all. Another variant would be to switch `vector` to `deque`.
> 
> On another node, it wasn't immediately obvious to me from the name `pushBackRunningNode` that it uses a fixed size, drops the first, and pushes back at the end (as in MPC). The name could also suggest that the vector is appended by one. Alas, I don't have a suggestion for a better name right now.
> 
> Edit: So my suggestion is this:
> 
> ```c++
> template <typename Scalar>
> void ShootingProblemTpl<Scalar>::pushBackRunningNode(boost::shared_ptr<ActionModelAbstract> model,
>                                                      boost::shared_ptr<ActionDataAbstract> data) {
>   if (!model->checkData(data)) {
>     throw_pretty("Invalid argument: "
>                  << "action data is not consistent with the action model")
>   }
> 
>   for (std::size_t i = 0; i < T_ - 1; ++i) {
>     running_models_[i] = running_models[i + 1];
>     running_datas_[i] = running_datas[i + 1];
>   }
>   running_models_[T_ - 1] = model;
>   running_datas_[T_ - 1] = data;
> }
> ```
> 
> Could you perhaps test that as well please?

@wxmerkt thank you for noting an obvious unnecessary copy.
I have benched your suggested code. The computation time is `2.6238` usecs.

I agree with you that the function name is terrible, it used to have more sense with `boost::circular_buffer`. Do you have any naming suggestion?

proyan:
Hi @cmastalli, 

One iteration of the solver, for 100 nodes and 4dof robot, takes about 400us on my system. I don't think we should change the API for an advantage of 2us.

wxmerkt:
Thanks for the quick test :-). The other option is [`deque`](https://en.cppreference.com/w/cpp/container/deque) where it would be `pop_front()`, followed by `insert`/`push_back`. I am not sure how much that affects memory locality in practice.

Regarding naming perhaps something with "cycleAndAppend" (but that isn't great either) - unfortunately I don't have a good suggestion for naming if the buffer size isn't fixed.

cmastalli:
> Hi @cmastalli,
> 
> One iteration of the solver, for 100 nodes and 4dof robot, takes about 400us on my system. I don't think we should change the API for an advantage of 2us.

Yes, it could be an undesired optimization. @wxmerkt, @jcarpent and @nmansard.
Do you agree with this?



> Thanks for the quick test :-). The other option is [`deque`](https://en.cppreference.com/w/cpp/container/deque) where it would be `pop_front()`, followed by `insert`/`push_back`. I am not sure how much that affects memory locality in practice.
> 
> Regarding naming perhaps something with "cycleAndAppend" (but that isn't great either) - unfortunately I don't have a good suggestion for naming if the buffer size isn't fixed.

Deque is worse than boost circular buffer. It doesn't have a constant complexity for accessing data. Boost circular buffer is indeed designed as std vectors but with deque functionalities.

wxmerkt:
Rohan's point regarding the comparatively minimal advantage of 2us in the grand scheme of things is important: These 2us - in relative terms - will be even less for higher dimensional systems (at the moment 0.5% of iteration time; likely much less for >4 DoF).

As thus, I don't think the change to circular buffer gives us enough benefit to deviate from a standard STL container. The STL container also has the benefit that when not used in MPC it change size. As thus, I vote for maintaining `vector`.

[I follow the theory behind deque, but we haven't benchmarked those implications here in practice, I assume]

cmastalli:
@wxmerkt and @proyan. I also agree with you, and thank you for your feedback.
I will create another PR with that approach.

__

Issue #774, opened 2020-05-15
gepetto-gui core dumps on Ubuntu 18.04


GlenHenshaw:
robotpkg *wants* to install gepetto-viewer=4.4, but gepetto-viewer-corab requires gepetto-viewer=4.2.

Manually specifying gepetto-viewer=4.2 installs fine, but trying to run gepetto-gui results in:

    Gtk-Message: 15:47:42.823: Failed to load module "canberra-gtk-module"
    libpng warning: iCCP: known incorrect sRGB profile
    terminate called after throwing an instance of 'CORBA::TRANSIENT'
    Aborted (core dumped)

nim65s:
Hi @GlenHenshaw 
gepetto-viewer 4.4 is really old.
I think you are looking for `sudo apt install robotpkg-py27-qt4-gepetto-viewer-corba`
You can adapt your python version, but for Qt, we are still bound to ubuntu's packaging of OSG with only qt4.

GlenHenshaw:
I'm just barely starting to learn about robotpkg, but I have a fresh Ubuntu 18.04 install and I installed the current robotpkg from their apt repository. And 4.4 is what it wants to install.

Building everything from source -- all the urdfdom dependencies, then gepetto -- seems to work. But the instructions indicate you can install via robotpkg, and that doesn't appear to work at least for me.

I'm not trying to do anything exotic, just get the examples running...

GlenHenshaw:
On the other hand, installing from source and then trying to run one of the crocoddyl exaples gives me this:

    python -m crocoddyl.examples.quadrupedal_gaits "display" "plot"
    *** SOLVE walking ***
    Traceback (most recent call last):
      File "/home/glenhenshaw/anaconda3/envs/crocoddyl/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
      File "/home/glenhenshaw/anaconda3/envs/crocoddyl/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
      File "/opt/openrobots/lib/python3.6/site-packages/crocoddyl/examples/quadrupedal_gaits.py", line 104, in <module>
    display = crocoddyl.GepettoDisplay(anymal, 4, 4, cameraTF, frameNames=[lfFoot, rfFoot, lhFoot, rhFoot])
      File "/opt/openrobots/lib/python3.6/site-packages/crocoddyl/__init__.py", line 57, in __init__
    self.addRobot()
[...]

PR #778, opened 2020-05-20
keys doesn't return a list in python3

proyan:


proyan:
Problem raised in #774

cmastalli:
Thank you for fixing this. 

I was thinking that it could be convenient to create a CI job for Python 3 in 18.04 environment with Release mode compilation.
If you prefer to do it here, then I could wait to merge the code.
Otherwise it is OK for me to do it in another PR.

proyan:
I'll merge this for now. I'm not that confident in my ability to setup a CI, and I don't know how much that would affect he CI time. I let you handle that.

PR #779, opened 2020-05-20
Extra functions in the ShootingProblem class to easily update models and datas

cmastalli:
This PR includes few extra functions for easily updating models and datas that are convenient for MPC applications:
 -  extra constructor that allows user to pass the datas
 - `updateNode`: update the node information (two methods, one of them allocates data)
 - `circularAppend`: emulate a `push_back` routine of a circular_buffer. This solution was discussed in #772.

Additionally, this PR includes:
 - a function to check that the data is consistent with the model,
 - doxygen documentation of the `ShootingProblem` members,
 - a clean-up of the `ActionModelAbstractTpl` class, and
 - a missed virtual keywords


proyan:
Thanks for the PR carlos

Issue #780, opened 2020-05-20
Data allocation in solvers doesn't consider contact-phase MPC applications


cmastalli:
Hi guys,

Let me first explain you quickly the problem.
I have a MPC formulation that smartly updates cost, contacts and impulse given an externally-defined contact-phase.
As you might see, when the motion evolves is imperatively to update the running nodes without allocating memory.
During this procedure, we introduce impulse models in different nodes with the original (or first) optimal control to be solved.

On the other hand, our solvers use the original formulation to allocate internally its data. However, if we update impulse models, then the dimension of some data changes (e.g. `Qu`, `k`, `K`, etc) due to `nu=0`. With this issue, I introduce two complementary options that I am willing to develop soon.

**Solution 1: simpler and faster**
We use the system `nu` to allocate the data, then internally the solvers turn-off useless computations when `nu=0`. With this solution we assume that there is a unique system `nu` with some autonomous action models (i.e. impulse dynamics).

**Solution 2: general**
We create a function that resize the corresponding datas which is triggered by the user. This option is general because we could accept any number of `nu`.
In solution 2, I don't plan to resize `ndx` since we assume that the state dimension never changes. This follows our current assumption.

@proyan any feedback about it? Note that we could discard solution 2 is we assume that there is a unique `nu` with some impulse dynamics. However, the assumption could be restrictive for future developments.

PR #782, opened 2020-05-22
Data allocation in solvers for contact-phase MPC applications

cmastalli:
This PR solves the issue #780 by applying the solution 1. It consists of:
 1. do not performances useless matrix-matrix multiplications when `nu=0`
 2. store a unique `nx`, `ndx` and `nu` in the shooting problem
 3. use 2 for allocating the solver data

Additionally, there are other tasks that this PR includes:
 1. Added doxygen documentation of  `SolverAbstract`, `SolverDDP` and `SolverFDDP`
 2. Tiny improvement in the doxygen documentation of `StateAbstractTpl`
 3. Configure doxygen to do not sort alphabetically the functions
 4. Updated the REAME file (cherry-pick from `release/v1.3.0` branch)

### DIFF
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -21,11 +21,47 @@ ShootingProblemTpl<Scalar>::ShootingProblemTpl(
       T_(running_models.size()),
       x0_(x0),
       terminal_model_(terminal_model),
-      running_models_(running_models) {
-  if (static_cast<std::size_t>(x0.size()) != running_models_[0]->get_state()->get_nx()) {
+      running_models_(running_models),
+      nx_(running_models[0]->get_state()->get_nx()),
+      ndx_(running_models[0]->get_state()->get_ndx()),
+      nu_(running_models[0]->get_nu()) {
+  // This routine is needed if we define an autonomous action model in the first node
+  if (nu_ == 0) {
+    for (std::size_t i = 1; i < T_; ++i) {
+      const boost::shared_ptr<ActionModelAbstract>& model = running_models_[i];
+      const std::size_t& nu = model->get_nu();
+      if (nu > 0) {
+        nu_ = nu;
+        break;
+      }
+    }
+  }
+  if (static_cast<std::size_t>(x0.size()) != nx_) {
     throw_pretty("Invalid argument: "
-                 << "x0 has wrong dimension (it should be " +
-                        std::to_string(running_models_[0]->get_state()->get_nx()) + ")");
+                 << "x0 has wrong dimension (it should be " + std::to_string(nx_) + ")");
+  }
+  for (std::size_t i = 1; i < T_; ++i) {
+    const boost::shared_ptr<ActionModelAbstract>& model = running_models_[i];
+    if (model->get_state()->get_nx() != nx_) {
+      throw_pretty("Invalid argument: "
+                   << "nx in " << i << " node is not consistent with the other nodes")
+    }
+    if (model->get_state()->get_ndx() != ndx_) {
+      throw_pretty("Invalid argument: "
+                   << "ndx in " << i << " node is not consistent with the other nodes")
+    }
+    if (model->get_nu() != 0 && model->get_nu() != nu_) {
+      throw_pretty("Invalid argument: "
+                   << "nu in " << i << " node is not consistent with the other nodes")
+    }
+  }
+  if (terminal_model_->get_state()->get_nx() != nx_) {
+    throw_pretty("Invalid argument: "
+                 << "nx in terminal node is not consistent with the other nodes")
+  }
+  if (terminal_model_->get_state()->get_ndx() != ndx_) {
+    throw_pretty("Invalid argument: "
+                 << "ndx in terminal node is not consistent with the other nodes")
   }
   allocateData();
 }
@@ -42,11 +78,13 @@ ShootingProblemTpl<Scalar>::ShootingProblemTpl(
       terminal_model_(terminal_model),
       terminal_data_(terminal_data),
       running_models_(running_models),
-      running_datas_(running_datas) {
-  if (static_cast<std::size_t>(x0.size()) != running_models_[0]->get_state()->get_nx()) {
+      running_datas_(running_datas),
+      nx_(running_models[0]->get_state()->get_nx()),
+      ndx_(running_models[0]->get_state()->get_ndx()),
+      nu_(running_models[0]->get_nu()) {
+  if (static_cast<std::size_t>(x0.size()) != nx_) {
     throw_pretty("Invalid argument: "
-                 << "x0 has wrong dimension (it should be " +
-                        std::to_string(running_models_[0]->get_state()->get_nx()) + ")");
+                 << "x0 has wrong dimension (it should be " + std::to_string(nx_) + ")");
   }
   std::size_t Td = running_datas.size();
   if (Td != T_) {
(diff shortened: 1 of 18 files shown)


cmastalli:
I have included an extra commit that allocated `xs_try`,`us_try` with zero values.
This is convenient for contact-phase MPC with impulse dynamics as well.
The reason if we want to use the previous solution for warm-starting, however, the values for impulse nodes remain NAN.

include/crocoddyl/core/optctrl/shooting.hxx
```diff
@@ -189,6 +263,19 @@ void ShootingProblemTpl<Scalar>::circularAppend(boost::shared_ptr<ActionModelAbs
...
 void ShootingProblemTpl<Scalar>::circularAppend(boost::shared_ptr<ActionModelAbstract> model) {
+  if (model->get_state()->get_nx() != nx_) {
+    throw_pretty("Invalid argument: "
+                 << "nx is not consistent with the other nodes")
+  }
+  if (model->get_state()->get_ndx() != ndx_) {
+    throw_pretty("Invalid argument: "
+                 << "ndx node is not consistent with the other nodes")
+  }
+  if (model->get_nu() != 0 && model->get_nu() != nu_) {
```
proyan, line 274: Please remove this change here as well.

src/core/solver-base.cpp
```diff
@@ -30,10 +30,9 @@ SolverAbstract::SolverAbstract(boost::shared_ptr<ShootingProblem> problem)
   us_.resize(T);
   for (std::size_t t = 0; t < T; ++t) {
     const boost::shared_ptr<ActionModelAbstract>& model = problem_->get_runningModels()[t];
-    const std::size_t& nu = model->get_nu();
 
     xs_[t] = model->get_state()->zero();
-    us_[t] = Eigen::VectorXd::Zero(nu);
+    us_[t] = Eigen::VectorXd::Zero(problem_->get_nu());
```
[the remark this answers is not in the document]
proyan: Actuation size is a property of the model, not the shooting problem

src/core/solvers/box-ddp.cpp
```diff
@@ -89,16 +82,23 @@ void SolverBoxDDP::forwardPass(const double& steplength) {
...
     m->get_state()->diff(xs_[t], xs_try_[t], dx_[t]);
-    us_try_[t].noalias() = us_[t] - k_[t] * steplength - K_[t] * dx_[t];
-    if (m->get_has_control_limits()) {  // clamp control
-      us_try_[t] = us_try_[t].cwiseMax(m->get_u_lb()).cwiseMin(m->get_u_ub());
+    if (m->get_nu() != 0) {
+      us_try_[t].noalias() = us_[t] - k_[t] * steplength - K_[t] * dx_[t];
+      if (m->get_has_control_limits()) {  // clamp control
+        us_try_[t] = us_try_[t].cwiseMax(m->get_u_lb()).cwiseMin(m->get_u_ub());
+      }
+      m->calc(d, xs_try_[t], us_try_[t]);
```
proyan, line 98: again, the solver should not care about what the dimension of actuation is. The API should remain the same here for both cases.

src/core/solvers/ddp.cpp
```diff
@@ -182,40 +190,48 @@ void SolverDDP::backwardPass() {
...
-    FuTVxx_p_[t].noalias() = d->Fu.transpose() * Vxx_p;
     Qxx_[t].noalias() += FxTVxx_p_ * d->Fx;
-    Qxu_[t].noalias() += FxTVxx_p_ * d->Fu;
-    Quu_[t].noalias() += FuTVxx_p_[t] * d->Fu;
     Qx_[t].noalias() += d->Fx.transpose() * Vx_p;
-    Qu_[t].noalias() += d->Fu.transpose() * Vx_p;
-
-    if (!std::isnan(ureg_)) {
-      Quu_[t].diagonal().array() += ureg_;
+    if (nu != 0) {
```
proyan, line 207: same here!

cmastalli:
@proyan I write a general comment here.

This PR works for impulse models! and indeed, the previous code wasn't working for impulse models if you were changing the impulse models in the shooting nodes.

The solvers allocate data with the real system `nu`. Then when we define an impulse model, the computations are ignored (those regarding `nu`). Furthermore, `nx`, `ndx` and `nu` in the shooting model are here to check that the user doesn't combine action models with wrong dimensions. They've been used to allocate data.

Note that we don't have c++ unittest for impulse models, however, we run impulse models in the quadrupedal examples (i.e. we are running the impulse models in the CI).

include/crocoddyl/core/optctrl/shooting.hxx
```diff
@@ -96,7 +146,11 @@ Scalar ShootingProblemTpl<Scalar>::calc(const std::vector<VectorXs>& xs, const s
 #pragma omp parallel for
 #endif
   for (std::size_t i = 0; i < T_; ++i) {
-    running_models_[i]->calc(running_datas_[i], xs[i], us[i]);
+    if (running_models_[i]->get_nu() != 0) {
+      running_models_[i]->calc(running_datas_[i], xs[i], us[i]);
```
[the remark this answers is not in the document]
cmastalli: We need it because `us` is a vector with the system dimension `nu`. Otherwise it would triggers an error for a impulse action (i.e. `nu != 0`)

proyan:
Carlos, `nu` is a property of the action model. You are constricting the shooting problem to only accept two possible actuations, 0 or the actuation size of the first action model. Theoretically, the same would be true for `nx` and `ndx`. But the problem with `nu` is more urgent

cmastalli:
> You are constricting the shooting problem to only accept two possible actuations, 0 or the actuation size of the first action model.

I agree with this. It could possible to work around this limitation if we want to exploit different `nu` values.
Let me think how we could do it, however, feel free to propose a solution by writing an new issue. 


> Theoretically, the same would be true for `nx` and `ndx`. But the problem with `nu` is more urgent

Note that all action models need to share the same `ndx` (`nx` doesn't play any role here). Otherwise, it is not possible to integrate the system dynamics

cmastalli:
For instance, we could combine this PR with solution 2 (described in #780). But I would like to reflect a bit about it.

proyan:
"I agree with this. It could possible to work around this limitation if we want to exploit different nu values."

We were already exploiting different nu values before you changed in this PR!

cmastalli:
> We were already exploiting different nu values before you changed in this PR!

Could you share a simplified Python script with this?
So, I could test the solution quickly.

Otherwise, I would need to spend a quite among of timing writing unittest before.

cmastalli:
Btw, have you tried this PR with your code?

proyan:
"Could you share a simplified Python script with this?"

What I mean is, the crocoddyl implementation we had already handled `nu` in different action models. It doesn't do that now.

cmastalli:
> What I mean is, the crocoddyl implementation we had already handled `nu` in different action models. It doesn't do that now.

I see. I believe by removing the `nu` checking might work as before. However, I need to write an unittest code to check it.

I will work on it, I promise to have a solution asap (and before the next release).
Is this OK for you?

proyan:
Again, we should define `nu`, `ndx`, and `nx` in the action models. This is not a property of a _ShootingProblem_.

cmastalli:
> Again, we should define `nu`, `ndx`, and `nx` in the action models. This is not a property of a _ShootingProblem_.

They are defined in the action model (as before). The shooting problem checks that they're the same and store it for easy retrieve.

PR #783, opened 2020-05-27
benchmark update

proyan:
Provides more detailed information on the benchmarks. specifically, maximum and minimum times, and std-deviation of the timings.


proyan:
Thanks for the review

PR #788, opened 2020-06-10
Contact force cost for 3d cases + added contact impulse cost mode + some documentation

cmastalli:
This PR extends the contact force cost model for 3d contact cases, and create a contact impulse cost. To validate the method, it is modified the unittest to check contact 3d as well. Finally, the PR includes doxygen documentation of the contact force/impulse cost model, and multiple-contact/impulse classes.


bindings/python/crocoddyl/multibody/costs/contact-impulse.cpp
```diff
@@ -0,0 +1,93 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2020-2022, University of Edinburgh
```
jcarpent, line 4: Why this 2022?

cmastalli: I assumed we define two years ahead copyright, is this not the correct way of doing?

nim65s: It is not. This line is about authorship: who authored, and when. You can't author something from the future. If **you** wrote this in **2020**, you should put 2020 - (your employer).

If you did not write anything in a file, you can't claim authorship on this file, and if you didn't write anything in 2022 in that file, you can't claim authorship on this file for that year.

cmastalli: Thank you for the clarification.
I fixed it.

PR #789, opened 2020-06-16
Minor fixes

proyan:
1) use inherited calcdiff for solver/fddp
2) fix cppad support for activation model quadratic barrier


proyan:
We are not currently testing the quadratic barrier. you are right, it needs to be added in test_codegen, I was waiting for the pinocchio support to be released first.

PR #792, opened 2020-06-29
CoP cost

julesser:
The intention of this PR is to implement a cost function for the Center of Pressure (CoP) of a single contact surface (i.e. one cost for each foot in contact). 

Motivation: Preliminary investigations for the trajectory generation of a humanoid robot have shown that the computed CoP for various motions (balancing, slow walking) leaves the convex hull around the contact points by far. This leads to inherently unbalanced trajectories and consequently unstable motions when utilizing a simple control architecture.  

**The basic idea:** 
1. The constraints for the CoP to lie inside the convex hull of the foot (see [eq.(18-19) of this paper](https://hal.archives-ouvertes.fr/hal-02108449/document)) can be written as:
`$\begin{align}
\begin{split}
\tau^x &\leq Yf^z \\
-\tau^x &\leq Yf^z \\
\tau^y &\leq Yf^z \\
-\tau^y &\leq Yf^z 
\end{split}
\end{align}$`
2. Implement these inequality constraints as: 
`$r = A \cdot f \geq 0$`
with the according A matrix as
`$\begin{equation}
\begin{bmatrix} 0 & 0 & X/2 & 0 & -1 & 0 \\
0 & 0 & X/2 & 0 & 1 & 0 \\
0 & 0 & Y/2 & 1 & 0 & 0 \\
0 & 0 & Y/2 & -1 & 0 & 0 \end{bmatrix} 
\end{equation}$`

### DIFF
--- a/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
+++ b/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
@@ -17,14 +17,29 @@ void exposeCostContactCoPPosition() {
   bp::class_<CostModelContactCoPPosition, bp::bases<CostModelAbstract> >(
       "CostModelContactCoPPosition",
       bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, FrameCoPSupport, int>(
-          bp::args("self", "state", "activation",
-                   "cop_support"
-                   "nu"),
+          bp::args("self", "state", "activation", "cop_support", "nu"),
           "Initialize the contact CoP position cost model.\n\n"
           ":param state: state of the multibody system\n"
-          ":param activation: activation model\n"
-          ":param cop_support: contact frame ID and cop support region"
+          ":param activation: activation model (default ActivationModelQuadraticBarrier)\n"
+          ":param cop_support: contact frame ID and cop support region\n"
           ":param nu: dimension of control vector"))
+      .def(bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, FrameCoPSupport>(
+          bp::args("self", "state", "activation", "cop_support"),
+          "Initialize the contact CoP position cost model.\n\n"
+          ":param state: state of the multibody system\n"
+          ":param activation: activation model (default ActivationModelQuadraticBarrier)\n"
+          ":param cop_support: contact frame ID and cop support region"))
+      .def(bp::init<boost::shared_ptr<StateMultibody>, FrameCoPSupport, int>(
+          bp::args("self", "state", "cop_support", "nu"),
+          "Initialize the contact CoP position cost model.\n\n"
+          ":param state: state of the multibody system\n"
+          ":param cop_support: contact frame ID and cop support region\n"
+          ":param nu: dimension of control vector"))
+      .def(bp::init<boost::shared_ptr<StateMultibody>, FrameCoPSupport>(
+          bp::args("self", "state", "cop_support"),
+          "Initialize the contact CoP position cost model.\n\n"
+          ":param state: state of the multibody system\n"
+          ":param cop_support: contact frame ID and cop support region"))
       .def<void (CostModelContactCoPPosition::*)(const boost::shared_ptr<CostDataAbstract>&,
                                                  const Eigen::Ref<const Eigen::VectorXd>&,
                                                  const Eigen::Ref<const Eigen::VectorXd>&)>(
--- a/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
+++ b/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
@@ -16,14 +16,15 @@ namespace python {
 void exposeCostContactCoPPosition() {
   bp::class_<CostModelContactCoPPosition, bp::bases<CostModelAbstract> >(
       "CostModelContactCoPPosition",
-      bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, FrameFootGeometry, 
-      const Eigen::Ref<const Eigen::VectorXd>&>(
-          bp::args("self", "state", "activation", "foot_geom", "normal"),
+      bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, FrameCoPSupport, 
+      const Eigen::Ref<const Eigen::VectorXd>&, int>(
+          bp::args("self", "state", "activation", "cop_support", "normal", "nu"),
           "Initialize the contact CoP position cost model.\n\n"
           ":param state: state of the multibody system\n"
           ":param activation: activation model\n"
-          ":param foot_geom: 2d geometry of the contact surface"
-          ":param normal: vector normal to the contact surface"))
+          ":param cop_support: contact frame ID and cop support region"
+          ":param normal: vector normal to the contact surface"
+          ":param nu: dimension of control vector"))
       .def<void (CostModelContactCoPPosition::*)(const boost::shared_ptr<CostDataAbstract>&,
                                                   const Eigen::Ref<const Eigen::VectorXd>&,
                                                   const Eigen::Ref<const Eigen::VectorXd>&)>(
@@ -48,8 +49,8 @@ void exposeCostContactCoPPosition() {
            ":param data: shared data\n"
            ":return cost data.")
       .add_property("reference",
-                    bp::make_function(&CostModelContactCoPPosition::get_footGeom, bp::return_internal_reference<>()),
-                    &CostModelContactCoPPosition::set_reference<FrameFootGeometry>, "reference foot geometry");
+                    bp::make_function(&CostModelContactCoPPosition::get_copSupport, bp::return_internal_reference<>()),
+                    &CostModelContactCoPPosition::set_reference<FrameCoPSupport>, "reference foot geometry");
 
   bp::register_ptr_to_python<boost::shared_ptr<CostDataContactCoPPosition> >();
 
--- a/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
+++ b/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
@@ -16,12 +16,14 @@ namespace python {
 void exposeCostContactCoPPosition() {
   bp::class_<CostModelContactCoPPosition, bp::bases<CostModelAbstract> >(
       "CostModelContactCoPPosition",
-      bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, FrameFootGeometry>(
-          bp::args("self", "state", "activation", "foot_geom"),
+      bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, FrameFootGeometry, 
+      const Eigen::Ref<const Eigen::VectorXd>&>(
+          bp::args("self", "state", "activation", "foot_geom", "normal"),
           "Initialize the contact CoP position cost model.\n\n"
           ":param state: state of the multibody system\n"
           ":param activation: activation model\n"
-          ":param foot_geom: 2d geometry of the contact surface"))
+          ":param foot_geom: 2d geometry of the contact surface"
+          ":param normal: vector normal to the contact surface"))
       .def<void (CostModelContactCoPPosition::*)(const boost::shared_ptr<CostDataAbstract>&,
                                                   const Eigen::Ref<const Eigen::VectorXd>&,
                                                   const Eigen::Ref<const Eigen::VectorXd>&)>(
--- a/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
+++ b/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
@@ -17,13 +17,12 @@ void exposeCostContactCoPPosition() {
   bp::class_<CostModelContactCoPPosition, bp::bases<CostModelAbstract> >(
       "CostModelContactCoPPosition",
       bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, FrameCoPSupport, 
-      const Eigen::Ref<const Eigen::VectorXd>&, int>(
-          bp::args("self", "state", "activation", "cop_support", "normal", "nu"),
+      int>(
+          bp::args("self", "state", "activation", "cop_support" "nu"),
           "Initialize the contact CoP position cost model.\n\n"
           ":param state: state of the multibody system\n"
           ":param activation: activation model\n"
           ":param cop_support: contact frame ID and cop support region"
-          ":param normal: vector normal to the contact surface"
           ":param nu: dimension of control vector"))
       .def<void (CostModelContactCoPPosition::*)(const boost::shared_ptr<CostDataAbstract>&,
                                                   const Eigen::Ref<const Eigen::VectorXd>&,
--- a/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
+++ b/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
@@ -1,7 +1,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2018-2020, University of Edinburgh
+// Copyright (C) 2020, University of Duisburg-Essen, University of Edinburgh
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
@@ -28,16 +28,16 @@ void exposeCostContactCoPPosition() {
           "calc", &CostModelContactCoPPosition::calc, bp::args("self", "data", "x", "u"),
           "Compute the contact CoP position cost.\n\n"
           ":param data: cost data\n"
-          ":param x: time-discrete state vector\n"
-          ":param u: time-discrete control input")
+          ":param x: state vector\n"
+          ":param u: control input")
       .def<void (CostModelContactCoPPosition::*)(const boost::shared_ptr<CostDataAbstract>&,
                                                   const Eigen::Ref<const Eigen::VectorXd>&,
                                                   const Eigen::Ref<const Eigen::VectorXd>&)>(
           "calcDiff", &CostModelContactCoPPosition::calcDiff, bp::args("self", "data", "x", "u"),
           "Compute the derivatives of the contact CoP position cost.\n\n"
           ":param data: action data\n"
-          ":param x: time-discrete state vector\n"
-          ":param u: time-discrete control input\n")
+          ":param x: state vector\n"
+          ":param u: control input\n")
       .def("createData", &CostModelContactCoPPosition::createData, bp::with_custodian_and_ward_postcall<0, 2>(),
            bp::args("self", "data"),
            "Create the contact CoP position cost data.\n\n"
@@ -47,7 +47,7 @@ void exposeCostContactCoPPosition() {
            ":return cost data.")
       .add_property("reference",
                     bp::make_function(&CostModelContactCoPPosition::get_footGeom, bp::return_internal_reference<>()),
-                    &CostModelContactCoPPosition::set_reference<FrameFootGeometry>, "reference frame friction cone");
+                    &CostModelContactCoPPosition::set_reference<FrameFootGeometry>, "reference foot geometry");
 
   bp::register_ptr_to_python<boost::shared_ptr<CostDataContactCoPPosition> >();
 
--- a/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
+++ b/bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
@@ -1,0 +1,70 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2018-2020, University of Edinburgh
+// Copyright note valid unless otherwise stated in individual files.
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#include "python/crocoddyl/multibody/multibody.hpp"
+#include "python/crocoddyl/multibody/cost-base.hpp"
+#include "crocoddyl/multibody/costs/contact-cop-position.hpp"
+
+namespace crocoddyl {
+namespace python {
+
+void exposeCostContactCoPPosition() {
+  bp::class_<CostModelContactCoPPosition, bp::bases<CostModelAbstract> >(
+      "CostModelContactCoPPosition",
+      bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, FrameFootGeometry>(
+          bp::args("self", "state", "activation", "foot_geom"),
+          "Initialize the contact CoP position cost model.\n\n"
+          ":param state: state of the multibody system\n"
+          ":param activation: activation model\n"
+          ":param foot_geom: 2d geometry of the contact surface"))
+      .def<void (CostModelContactCoPPosition::*)(const boost::shared_ptr<CostDataAbstract>&,
+                                                  const Eigen::Ref<const Eigen::VectorXd>&,
+                                                  const Eigen::Ref<const Eigen::VectorXd>&)>(
+          "calc", &CostModelContactCoPPosition::calc, bp::args("self", "data", "x", "u"),
+          "Compute the contact CoP position cost.\n\n"
+          ":param data: cost data\n"
+          ":param x: time-discrete state vector\n"
+          ":param u: time-discrete control input")
+      .def<void (CostModelContactCoPPosition::*)(const boost::shared_ptr<CostDataAbstract>&,
+                                                  const Eigen::Ref<const Eigen::VectorXd>&,
+                                                  const Eigen::Ref<const Eigen::VectorXd>&)>(
+          "calcDiff", &CostModelContactCoPPosition::calcDiff, bp::args("self", "data", "x", "u"),
+          "Compute the derivatives of the contact CoP position cost.\n\n"
+          ":param data: action data\n"
+          ":param x: time-discrete state vector\n"
+          ":param u: time-discrete control input\n")
+      .def("createData", &CostModelContactCoPPosition::createData, bp::with_custodian_and_ward_postcall<0, 2>(),
+           bp::args("self", "data"),
+           "Create the contact CoP position cost data.\n\n"
+           "Each cost model has its own data that needs to be allocated. This function\n"
+           "returns the allocated data for a predefined cost.\n"
+           ":param data: shared data\n"
+           ":return cost data.")
+      .add_property("reference",
+                    bp::make_function(&CostModelContactCoPPosition::get_footGeom, bp::return_internal_reference<>()),
+                    &CostModelContactCoPPosition::set_reference<FrameFootGeometry>, "reference frame friction cone");
+
+  bp::register_ptr_to_python<boost::shared_ptr<CostDataContactCoPPosition> >();
+
+  bp::class_<CostDataContactCoPPosition, bp::bases<CostDataAbstract> >(
+      "CostDataContactCoPPosition", "Data for contact CoP position cost.\n\n",
+      bp::init<CostModelContactCoPPosition*, DataCollectorAbstract*>(
+          bp::args("self", "model", "data"),
+          "Create contact CoP position cost data.\n\n"
+          ":param model: contact CoP position cost model\n"
+          ":param data: shared data")[bp::with_custodian_and_ward<1, 2, bp::with_custodian_and_ward<1, 3> >()])
+      .add_property("Arr_Ru", bp::make_getter(&CostDataContactCoPPosition::Arr_Ru, bp::return_internal_reference<>()),
+                    "Intermediate product of Arr (2nd deriv of Activation) with Ru (deriv of residue)")
+      .add_property(
+          "contact",
+          bp::make_getter(&CostDataContactCoPPosition::contact, bp::return_value_policy<bp::return_by_value>()),
+          bp::make_setter(&CostDataContactCoPPosition::contact), "contact data associated with the current cost");
+}
+
+}  // namespace python
+}  // namespace crocoddyl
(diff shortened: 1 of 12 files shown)


include/crocoddyl/multibody/costs/contact-cop-position.hxx
```diff
@@ -0,0 +1,94 @@
...
+
+  // Define the inequality matrix as A * f <= 0 compare eq.(18-19) in https://hal.archives-ouvertes.fr/hal-02108449/document
+  //Matrix3s c_R_o = Quaternions::FromTwoVectors(nsurf_, Vector3s::UnitZ()).toRotationMatrix(); TODO: Rotation necessary for each row of A?
+  d->A << 0, 0, -foot_geom_.dim[1] / 2, 1, 0, 0,
+             0, 0, -foot_geom_.dim[1] / 2, -1, 0, 0,
+             0, 0, -foot_geom_.dim[0] / 2, 0, 1, 0,
+             0, 0, -foot_geom_.dim[0] / 2, 0, -1, 0;
+
+  // Compute the cost residual   
+  data->r.noalias() = d->A * d->f; //TODO: Debug error: no match for ‘operator*’
```
cmastalli, line 47: You need to get the vector of `d->f`, i.e. `d->f.toVector()`

include/crocoddyl/multibody/costs/contact-cop-position.hxx
```diff
@@ -0,0 +1,94 @@
...
+
+template <typename Scalar>
+void CostModelContactCoPPositionTpl<Scalar>::calcDiff(const boost::shared_ptr<CostDataAbstract>& data, //TODO: Correct assignment of the derivatives?
+                                                      const Eigen::Ref<const VectorXs>&, 
+                                                      const Eigen::Ref<const VectorXs>&) {
+  // Update all data
+  Data* d = static_cast<Data*>(data.get());
+
+  // Get the derivatives of the contact wrench
+  const MatrixXs& df_dx = d->f->df_dx; //TODO: Debug error: base operand of ‘->’ has non-pointer type
```
cmastalli, line 62: Change to:
```cpp
  const MatrixXs& df_dx = d->contact->df_dx;
  const MatrixXs& df_du = d->contact->df_du;
```

include/crocoddyl/multibody/costs/contact-cop-position.hxx
```diff
@@ -0,0 +1,94 @@
...
+  const MatrixXs& df_du = d->f->df_du; //TODO: Same here
+
+  //Compute the derivatives of the activation function
+  activation_->calcDiff(data->activation, data->r);
+
+  //Compute the derivatives of the cost residual
+  data->Rx.noalias() = d->A * df_dx;
```
cmastalli, line 69: Get A matrix from `footGeom`

include/crocoddyl/multibody/frames.hpp
```diff
@@ -136,6 +136,25 @@ struct FrameFrictionConeTpl {
...
+  typedef _Scalar Scalar;
+  typedef typename MathBaseTpl<Scalar>::Vector2s Vector2s;
+
+  explicit FrameFootGeometryTpl() : frame(0), dim(Vector2s::Zero()) {}
+  FrameFootGeometryTpl(const FrameFootGeometryTpl<Scalar>& value) : frame(value.frame), dim(value.dim) {}
+  FrameFootGeometryTpl(const FrameIndex& frame, const Vector2s& dim) : frame(frame), dim(dim) {}
+  friend std::ostream& operator<<(std::ostream& os, const FrameTranslationTpl<Scalar>& X) {
+    os << "      frame: " << X.frame << std::endl << "foot dimensions: " << std::endl << X.dim << std::endl;
+    return os;
+  }
```
cmastalli, line 152: Create `get_A` and `update_A` functions

unittest/bindings/test_cop_cost.py
```diff
@@ -0,0 +1,55 @@
...
+from test_utils import NUMDIFF_MODIFIER, assertNumDiff
+
+# Create robot model, data, state and actuation
+ROBOT_MODEL = example_robot_data.loadICub().model
+ROBOT_DATA = ROBOT_MODEL.createData()
+ROBOT_STATE = crocoddyl.StateMultibody(ROBOT_MODEL)
+ACTUATION = crocoddyl.ActuationModelFloatingBase(ROBOT_STATE)
+
+# Create friction cone and its activation
+frictionCone = crocoddyl.FrictionCone(np.matrix([0., 0., 1.]).T, 0.7, 4, False)
```
cmastalli, line 15: These lines are not needed since we are not testing the friction cone cost.

include/crocoddyl/multibody/costs/contact-cop-position.hpp
```diff
@@ -0,0 +1,154 @@
...
+  typedef typename MathBase::Matrix3xs Matrix3xs;
+  typedef typename MathBase::MatrixX3s MatrixX3s;
+  typedef typename MathBase::Matrix6xs Matrix6xs;
+  typedef typename MathBase::Matrix6s Matrix6s;
+  typedef typename MathBase::Vector6s Vector6s;
+
+  template <template <typename Scalar> class Model>
+  CostDataContactCoPPositionTpl(Model<Scalar>* const model, DataCollectorAbstract* const data)
+      : Base(model, data), Arr_Ru(model->get_activation()->get_nr(), model->get_state()->get_nv()) {
+    Arr_Ru.setZero();
```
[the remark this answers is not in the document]
julesser: Good idea.

include/crocoddyl/multibody/costs/contact-cop-position.hxx
julesser: That did the trick, thanks for the hint Carlos!

include/crocoddyl/multibody/costs/contact-cop-position.hxx
julesser: I'm not shure about this. Shouldn't we calculate the derivatives of the contact wrench in the desired frame (i.e. `d->f`)? Otherwise I guess we would need a transformation afterwards again.

include/crocoddyl/multibody/costs/contact-cop-position.hpp
```diff
@@ -0,0 +1,154 @@
...
+      throw_pretty("Domain error: there isn't defined contact data for " + frame_name);
+    }
+  }
+
+  pinocchio::DataTpl<Scalar>* pinocchio;
+  MatrixXs Arr_Ru;
+  boost::shared_ptr<ContactDataAbstractTpl<Scalar> > contact; // spatial force expressed in world coordinates
+  pinocchio::SE3Tpl<Scalar> fiMo; // SE3 object with origin at contact point and rotation aligned with the world frame
+  pinocchio::ForceTpl<Scalar> f; // cartesian force expressed in world coordinates 
+  MatrixX3s A; // inequality matrix
```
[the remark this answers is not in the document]
julesser: I like this idea.

include/crocoddyl/multibody/costs/contact-cop-position.hxx
```diff
@@ -0,0 +1,94 @@
...
+                                           const Eigen::Ref<const VectorXs>&, const Eigen::Ref<const VectorXs>&) {
+  Data* d = static_cast<Data*>(data.get());
+
+  // Transform the spatial force to a cartesian force expressed in world coordinates  
+  d->fiMo = d->pinocchio.SE3(d->pinocchio->oMi[d->contact->joint].rotation(), d->contact->jMf.translation()); //TODO: Debug error: request for member ‘SE3’
+  d->f = d->fiMo.actInv(d->contact->f);
+  
+  // Compute the CoP (for evaluation)
+  // OC = (tau_0^p x n) / (n * f^p) compare eq.(13) in https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.138.8014&rep=rep1&type=pdf
+  d->cop << -d->f.angular()[1] / d->f.linear()[2], d->f.angular()[0] / d->f.linear()[2], 0.0;
```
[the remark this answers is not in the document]
julesser: I've extended the equation for the general case now, including the normal vector. However, this equation here is **only used for evaluation purpose, not for the computation of the cost**! So we should to discuss this separately.

julesser:
Thanks for your helpful feedback Carlos, I have pushed some changes accordingly. Please find the open issues in your remarks to discuss them separately.

proyan:
Hi @julesser 

Sorry for the late reply. I just saw your code, and you've done good work. 
1) It looks like you are creating a bound on the cop. I would prefer if you name this cost that reflects the same (e.g., CostModelCoPBound).
2) Also, please don't create another new class FootGeometry for just one cost. The A matrix and the frame index can be stored inside the cost model itself.
3) In order to make sure that the default activationquadratic is not used here, please change the default activation to inequality.

For reference, I used a similar cost in python for my simulations, you could use that:

```python
class CostModelZmpBound(CostModelPinocchio):
    '''
    The class implements a model for keeping ZMP with a given bound.
    The residual is given by tau_x-z*fz, z*fz-tau_x, tau_y-z*fz, z*fz-tau_y.
    For the zmp to be within the bounds, residual should always be positive.
    '''

    def __init__(self, pinocchioModel, contactModel, ncost=4, zmpRadius=0.01, nu=None, activation=None):
        self.CostDataType = CostDataZmpBound
        CostModelPinocchio.__init__(self, pinocchioModel, ncost=ncost, nu=nu)
        self.zmpRadius = zmpRadius
        self.contact = contactModel

        self.boundMatrix = np.zeros((4,6))
        self.boundMatrix[:,2] = self.zmpRadius
        self.boundMatrix[0,3] = 1.0; self.boundMatrix[1,3] = -1.0;
        self.boundMatrix[2,4] = 1.0; self.boundMatrix[3,4] = -1.0;
        assert (activation is None and "defining your own activation model is not possible here.")
        self.activation = ActivationModelInequality(np.array([0.]*ncost),
                                                    np.array([np.inf]*ncost))

    def calc(self, data, x, u):
        if data.contact is None:
            raise RuntimeError('''The CostForce data should be specifically initialized from the
            contact data ... no automatic way of doing that yet ...''')
        data.f = data.contact.f
        zfz = self.zmpRadius*data.f[2]
        tx = data.f[3]
        ty = data.f[4]
        data.residuals = np.array([tx+zfz, zfz-tx, ty+zfz, zfz-ty])
        data.cost = sum(self.activation.calc(data.activation, data.residuals))
        return data.cost

    def calcDiff(self, data, x, u, recalc=True):
        if recalc:
            self.calc(data, x, u)
        assert (self.nu == len(u) and self.contact.nu == self.nu)
        df_dx, df_du = data.contact.df_dx, data.contact.df_du
        Ax, Axx = self.activation.calcDiff(data.activation, data.residuals, recalc=recalc)
        sel = Axx.astype(bool)[:, 0]
        #print sel, data.cost, data.residuals, data.f
        A = self.boundMatrix[sel, :]
        A2 = np.dot(A.T, A)

        data.Rx[:, :] = np.dot(self.boundMatrix, df_dx)
        data.Ru[:, :] = np.dot(self.boundMatrix, df_du)

        data.Lx[:] = np.dot(data.Rx[sel, :].T, Ax[sel])
        data.Lu[:] = np.dot(data.Ru[sel, :].T, Ax[sel])

        data.Lxx[:, :] = np.dot(df_dx.T, np.dot(A2, df_dx))
        data.Lxu[:, :] = np.dot(df_dx.T, np.dot(A2, df_du))
        data.Luu[:, :] = np.dot(df_du.T, np.dot(A2, df_du))

        return data.cost
```

proyan:
@olivier-stasse I've added the cost model for your reference

julesser:
Hi @proyan and thanks a lot for your feedback and advice. Good to hear that we are on the right track with this :)

> 1. It looks like you are creating a bound on the cop. I would prefer if you name this cost that reflects the same (e.g., CostModelCoPBound).
> 2. Also, please don't create another new class FootGeometry for just one cost. The A matrix and the frame index can be stored inside the cost model itself.

This is a design choice and I can modify according to your needs. @cmastalli What do you think?

> 3. In order to make sure that the default activationquadratic is not used here, please change the default activation to inequality.

100% agreed. 

> You don't need to calculate and save the CoP explicitly. Avoid extra calculations :). Just like the eq 18-19 of the paper https://hal.archives-ouvertes.fr/hal-02108449/document that you mentioned, this cost is an bounded cost on the contact force. So only calculate the eq 18 and 19.

You are right. The purpose right now is just to have the resulting CoP position for evaluation purpose, I will remove it before merging! Btw, [eq (18-19) ](https://hal.archives-ouvertes.fr/hal-02108449/document) are only valid for the flat floor case I guess. So I would derive the general inequality conditions as depicted with #792, including X and Y, correct?

bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
```diff
@@ -16,14 +16,16 @@ namespace python {
 void exposeCostContactCoPPosition() {
   bp::class_<CostModelContactCoPPosition, bp::bases<CostModelAbstract> >(
       "CostModelContactCoPPosition",
-      bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, FrameFootGeometry, 
-      const Eigen::Ref<const Eigen::VectorXd>&>(
-          bp::args("self", "state", "activation", "foot_geom", "normal"),
+      bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, 
+      const std::size_t&, const Eigen::Ref<const Eigen::VectorXd>&, const Eigen::Ref<const Eigen::VectorXd>&, int>(
```
cmastalli, line 20: 1. Use Vector2 and Vector3, instead of VectorXd.
2. Do not use `Eigen::Ref`

include/crocoddyl/multibody/costs/contact-cop-position.hpp
```diff
@@ -53,17 +53,20 @@ class CostModelContactCoPPositionTpl : public CostModelAbstractTpl<_Scalar> {
...
   using Base::nu_;
   using Base::state_;
   using Base::unone_;
 
-  protected: 
-    FootGeometry foot_geom_; //!< frame name and geometrical dimension of the contact foot
-    const Vector3s normal_; //!< vector normal to the contact surface 
-    Vector3s foot_pos_; //!< position of the foot w.r.t. the ground plane
+ private: 
+  const std::size_t frame_id_; //!< contact frame id
```
cmastalli, line 67: Do use `const` for all private objects. We might want to change its values at runtime

include/crocoddyl/multibody/costs/contact-cop-position.hpp
```diff
@@ -127,10 +136,12 @@ struct CostDataContactCoPPositionTpl : public CostDataAbstractTpl<_Scalar> {
 
   pinocchio::DataTpl<Scalar>* pinocchio;
   MatrixXs Arr_Ru;
-  boost::shared_ptr<ContactDataAbstractTpl<Scalar> > contact; //!< spatial force expressed in world coordinates
-  pinocchio::SE3Tpl<Scalar> fiMo; //!< SE3 object with origin at contact point and rotation aligned with the world frame
-  pinocchio::ForceTpl<Scalar> f; //!< cartesian force expressed in world coordinates 
-  Vector3s cop; //!< center of pressure
+  boost::shared_ptr<ContactDataAbstractTpl<Scalar> > contact; //!< contact force
+  pinocchio::ForceTpl<Scalar> f; //!< transformed contact force
+  std::size_t frame_id; //!< contact frame id
```
cmastalli, line 141: Do not store frame_id, cop_region and normal in the data.
These objects belong to the model.

include/crocoddyl/core/mathbase.hpp
```diff
@@ -25,6 +25,7 @@ struct MathBaseTpl {
   typedef Eigen::Matrix<Scalar, 6, 1> Vector6s;
   typedef Eigen::Matrix<Scalar, 3, 3> Matrix3s;
   typedef Eigen::Matrix<Scalar, 6, 6> Matrix6s;
+  typedef Eigen::Matrix<Scalar, 4, 6> Matrix46s;
```
proyan, line 28: Don't add this here. Use the typedef inside the cost itself. This is too specific, and this naming is a bit weird.

include/crocoddyl/multibody/costs/contact-cop-position.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+ protected:
+  using Base::activation_;
+  using Base::nu_;
+  using Base::state_;
+  using Base::unone_;
+
+ private: 
+  const std::size_t frame_id_; //!< contact frame id
+  const Vector2s cop_region_; //!< desired 2d area (length, width) of the CoP defined w.r.t. the contact frame
```
proyan, line 68: Name it support_area

include/crocoddyl/multibody/costs/contact-cop-position.hpp
```diff
@@ -0,0 +1,165 @@
...
+      throw_pretty("Invalid argument: the shared data should be derived from DataCollectorContact");
+    }
+
+    //Get model parameters
+    frame_id = model->get_frame_id();
+    cop_region = model->get_cop_region();
+    normal = model->get_normal();
+
+    //Compute the inequality matrix
+    A << 0, 0, cop_region[0] / 2, 0, -1, 0,
```
proyan, line 107: replace with Scalar(0) Scalar(2) Scalar(-1)...

include/crocoddyl/multibody/costs/contact-cop-position.hxx
```diff
@@ -0,0 +1,92 @@
...
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#include "crocoddyl/core/utils/exception.hpp"
+#include "crocoddyl/multibody/costs/contact-cop-position.hpp"
+
+namespace crocoddyl {
+
+template<typename _Scalar>
+CostModelContactCoPPositionTpl<_Scalar>::CostModelContactCoPPositionTpl(boost::shared_ptr<StateMultibody> state,
```
proyan, line 15: Can you please make sure here that the default activation is quadraticbarrier?

include/crocoddyl/multibody/costs/contact-cop-position.hxx
```diff
@@ -0,0 +1,92 @@
...
+  return boost::allocate_shared<Data>(Eigen::aligned_allocator<Data>(), this, data);
+}
+
+template<typename Scalar>
+const std::size_t& CostModelContactCoPPositionTpl<Scalar>::get_frame_id() const {
+  return frame_id_;
+}
+
+template<typename Scalar>
+const typename MathBaseTpl<Scalar>::Vector2s& CostModelContactCoPPositionTpl<Scalar>::get_cop_region() const {
```
proyan, line 83: change the name here to support_area as well

bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
julesser: Shall we directly access the typedef `Vector2s&`, or what do you suggest instead?

include/crocoddyl/multibody/costs/contact-cop-position.hpp
julesser: Also considered this. However, wouldn't it be a bit confusing since the ideal CoP region can be a subregion of the actual support area of the foot?

include/crocoddyl/multibody/costs/contact-cop-position.hxx
julesser: See above comment

include/crocoddyl/multibody/costs/contact-cop-position.hxx
julesser: I've added `typedef ActivationModelQuadraticBarrierTpl<Scalar> ActivationModelQuadraticBarrier;` and replaced the activation everywhere. Are there additionally requirements for the `cost-base.hpp` (related compilation error occurs)?

cmastalli:
@proyan as discussed internally, we need to have a `frameCoPSupport` class to be able to modify the reference of both: `frame_id` and `support_region` through `set_reference`.

Let me explain you a bit more how it works, contrary to what you might think, this method makes the API much simpler, especially for c++ code, for updating references (of any kind, not just vectors). There are two functions `set_reference` (template function) and `set_referenceImpl` (virtual function). The former uses a template to easily define any reference, while the later allows us to do proper checking of the type and its conversion. An example of this, you could find it for the frame-velocity cost:
  - `set_reference`: https://github.com/loco-3d/crocoddyl/blob/master/include/crocoddyl/multibody/cost-base.hpp#L60
  - `set_referenceImpl`: https://github.com/loco-3d/crocoddyl/blob/master/include/crocoddyl/multibody/costs/frame-velocity.hxx#L92

These are the benefits:
 1. **No down-casting**: This method simplifies the process to users (especially c++ users) because you don't need to downcast the abstract cost class into the expected one. Otherwise, the users need to add extra lines for setting or getting a reference.
 2. **Reference are more than vector**: The structures `FrameXXX` are done to define the references. Reference are desired values of our cost function, they are not just vectors because you might want to choose an specific frame id or any kind of reference structure such as a friction cone.

This method has been applied for all the cost functions and release a few months ago. Frankly speaking, this is not the right PR to debate about changing these stuffs. If you have a better idea, then write an issue that explains it. (Before making any decision / proposition, take your time to play with it)

@julesser please reverted the changes regarding the `FrameCoPSupport`. See the explanations to understand why we need this. You will also need to implement `set_referenceImpl`. I offer you my help since I understand that going back & forth is unpleasant.

include/crocoddyl/multibody/costs/contact-cop-position.hxx
julesser: I've changed it to `support_region` as suggested by Carlos.

include/crocoddyl/multibody/costs/contact-cop-position.hpp
julesser: I've changed the name to `support_region` as suggested by Carlos.

bindings/python/crocoddyl/multibody/frames.cpp
```diff
@@ -90,6 +90,18 @@ void exposeFrames() {
...
+
+  bp::class_<FrameCoPSupport>("FrameCoPSupport",
+                                "Frame foot geometry.\n\n"
+                                "It defines the geometry of the contact surface for a given frame ID",
+                                bp::init<FrameIndex, Eigen::Vector2d>(bp::args("self", "frame", "support_region"),
+                                                                    "Initialize the frame foot geometry.\n\n"
+                                                                    ":param frame: frame ID\n"
+                                                                    ":param support_region: 2d-vector containg the length and width of the foot"))
+      .def(bp::init<>(bp::args("self"), "Default initialization of the frame friction cone."))
+      .def_readwrite("frame", &FrameCoPSupport::frame, "frame ID")
```
cmastalli, line 102: Name the normal as `normal`

include/crocoddyl/multibody/costs/contact-cop-position.hpp
```diff
@@ -0,0 +1,148 @@
...
+                          const Vector3s& normal, const std::size_t& nu);
+  virtual ~CostModelContactCoPPositionTpl();
+
+  virtual void calc(const boost::shared_ptr<CostDataAbstract>& data, const Eigen::Ref<const VectorXs>& x,
+                    const Eigen::Ref<const VectorXs>& u);
+  virtual void calcDiff(const boost::shared_ptr<CostDataAbstract>& data, const Eigen::Ref<const VectorXs>& x,
+                        const Eigen::Ref<const VectorXs>& u);
+  virtual boost::shared_ptr<CostDataAbstract> createData(DataCollectorAbstract* const data);
+
+  const CoPSupport& get_copSupport() const;
```
cmastalli, line 57: Remove this function. Instead we need to define `set_referenceImpl`. See this example: https://github.com/loco-3d/crocoddyl/blob/master/include/crocoddyl/multibody/costs/contact-friction-cone.hxx#L100

bindings/python/crocoddyl/multibody/costs/contact-cop-position.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+void exposeCostContactCoPPosition() {
+  bp::class_<CostModelContactCoPPosition, bp::bases<CostModelAbstract> >(
+      "CostModelContactCoPPosition",
+      bp::init<boost::shared_ptr<StateMultibody>, boost::shared_ptr<ActivationModelAbstract>, FrameCoPSupport, 
+      const Eigen::Ref<const Eigen::VectorXd>&, int>(
+          bp::args("self", "state", "activation", "cop_support", "normal", "nu"),
+          "Initialize the contact CoP position cost model.\n\n"
+          ":param state: state of the multibody system\n"
+          ":param activation: activation model\n"
```
cmastalli, line 24: Document the default activation as quadratic barrier, i.e.
```c++
          ":param activation: activation model (default ActivationModelQuadraticBarrier) \n"
```

include/crocoddyl/multibody/costs/contact-cop-position.hxx
cmastalli: @julesser you need to create a constructor that doesn't include `activation` argument, and then to pass to `Base` your default one.

If you don't know how to do it, I could deal myself with it.

You will also need to declare both constructors for the Python bindings.

include/crocoddyl/multibody/frames.hpp
```diff
@@ -136,6 +136,40 @@ struct FrameFrictionConeTpl {
...
+  FrameCoPSupportTpl(const FrameIndex& frame, const Vector2s& support_region) : frame(frame), support_region(support_region) {}
+  friend std::ostream& operator<<(std::ostream& os, const FrameCoPSupportTpl<Scalar>& X) {
+    os << "      frame: " << X.frame << std::endl << "foot dimensions: " << std::endl << X.support_region << std::endl;
+    return os;
+  }
+
+  // Define the inequality matrix A to implement A * f <= 0 compare eq.(18-19) in https://hal.archives-ouvertes.fr/hal-02108449/document
+  //Matrix3s c_R_o = Quaternions::FromTwoVectors(nsurf_, Vector3s::UnitZ()).toRotationMatrix(); TODO: Rotation necessary for each row of A?
+  void update_A() {
+  A_ << 0, 0, support_region[0] / 2, 0, -1, 0,
```
cmastalli, line 158: You need to add the scalar thing that @proyan requested. This is important for code-generating this cost.

include/crocoddyl/multibody/frames.hpp
```diff
@@ -136,6 +136,40 @@ struct FrameFrictionConeTpl {
...
+  FrameCoPSupportTpl(const FrameCoPSupportTpl<Scalar>& value) : frame(value.frame), support_region(value.support_region) {}
+  FrameCoPSupportTpl(const FrameIndex& frame, const Vector2s& support_region) : frame(frame), support_region(support_region) {}
+  friend std::ostream& operator<<(std::ostream& os, const FrameCoPSupportTpl<Scalar>& X) {
+    os << "      frame: " << X.frame << std::endl << "foot dimensions: " << std::endl << X.support_region << std::endl;
+    return os;
+  }
+
+  // Define the inequality matrix A to implement A * f <= 0 compare eq.(18-19) in https://hal.archives-ouvertes.fr/hal-02108449/document
+  //Matrix3s c_R_o = Quaternions::FromTwoVectors(nsurf_, Vector3s::UnitZ()).toRotationMatrix(); TODO: Rotation necessary for each row of A?
+  void update_A() {
```
cmastalli, line 157: You forgot to bind this function.

include/crocoddyl/multibody/frames.hpp
```diff
@@ -136,6 +136,40 @@ struct FrameFrictionConeTpl {
...
+        0, 0, support_region[1] / 2, -1, 0, 0;
+  }
+
+  const Matrix46& get_A() const {
+    return A_;
+}
+
+  FrameIndex frame; //!< name of the contact frame 
+  Vector2s support_region; //!< dimension of the foot surface dim = (length, width)
+  Matrix46 A_; //!< inequality matrix
```
cmastalli, line 170: You also forgot to bind this functions. Additionally, it needs to be renamed as `A`, not `A_`.

Quick clarification about the code style. Structure's members do not use `_` as last character, the opposite is for the class's members. (this is Google code style that we follow in the project)

include/crocoddyl/multibody/costs/contact-cop-position.hpp
julesser: We need the frame id for getting the `contact`, currently we use `get_copSupport()` to have access to this id. How do you want to retrieve the frame id then?

cmastalli: Ok, I forgot to mention that we need to define `get_referenceImpl` as well.
Is this answer your question?

julesser: Actually not. I don't understand how `get_referenceImpl` allows for accessing the `FrameCoPSupport` values since there seem to be no return values. What do I miss here?

julesser:
@cmastalli  Thanks for your feedback, I have pushed the recent changes. Regarding the remaining style and reference issues (see unresolved conservations), I would appreciate your contribution which probably will be the more convenient way.

julesser:
Hi guys! I ran some tests with the CoP cost and can confirm it is working as expected. See e.g. this results of a bipedal walking application, where the cost is included as 
```
CoP = crocoddyl.CostModelContactCoPPosition(self.state, 
                crocoddyl.FrameCoPSupport(i, np.array([0.2, 0.08]), self.nsurf), self.actuation.nu)
costModel.addCost(self.rmodel.frames[i].name + "_CoP", CoP, 1e2
```
**Results for performing 2 steps with a humanoid (top view on the feet):**
As desired, the CoP now stays within the define `support_region`, which here equals the boundaries of the foot geometry. Two notes (as already dicussed internally): 
- In order to improve the robustness of the contact stability, the user simply can reduce `support_region` to a smaller area than the foot geometry itself, e.g. support_region = 0.5 * foot_geometry. 
- `support_region` always spans from the origin of the defined contact frame. In order to move this reference to say the ankle region, the user is suggested to define a new frame at the desired position (if the URDF model not already contains one at this position) and pass it to `FrameCoPSupport` **as well as** to the `ContactModel6D`. 
StabilityAnalysisCoPCostActive

In case you have any further proposals, let me know! Otherwise, I think we're ready to merge.

cmastalli:
@julesser thank you for this contribution and taking care for each feedback.

@proyan thank you to help me with the review of this PR.

PR #796, opened 2020-07-10
[Notebook] Fix display rate of meshcat visualizer

paLeziart:
A small fix to change the display rate of the meshcat visualizer at the end of the bipedal walking notebook.

With that setting you have enough time to scroll up and see the animation of the walk.


cmastalli:
Thank you for this contribution :)

PR #797, opened 2020-07-10
[Notebook] Fix dynamic equation and add analytical derivatives for cartpole

paLeziart:
Fix `thdot**2` into `thdot` in the dynamic equation for `xddot`.

Add the analytical derivatives of the cartpole as an alternative to the numerical approximation with NumDiff.

Need to be tested for the crocoddyl repo before merging. 

I've tested it in the notebook of the summer school but there seem to be some differences with the notebook of this repo. For instance I got deprecated function warnings when I run the crocoddyl repo's notebook and the display functions are different.

@nim65s Is it possible to propagate that to the summer school repo or should I make another PR ?


nim65s:
> @nim65s Is it possible to propagate that to the summer school repo or should I make another PR ?

The branch for the summer-school has been modified to strip some things that were not useful for the summer school, but which would break features and options that can be otherwise used by crocoddyl users.

eg. `A.dot(B)` was replaced by `A @ B`, as it is easier to read for students during the summer school, as they had Python 3, but this would break Python2 compatibility if we merged that in the main repo. Or another e.g., some code that lets user choose between gepetto-viewer and meshcat was removed, as we chose for them to use meshcat, and they didn't need to be cluttered by code which is out of topic.

Then, I don't know what @cmastalli wants to do with this branch:
- we can choose to merge it, and force all notebook user to use python 3 & meshcat
- we can maintain both branch in parallel, and try to ensure that each change on one gets reflected on the other
- we can declare that summer-school branch as archived
- we can just put a few fixes when we find those, but overall try not to modify it too much
- etc.

cmastalli:
> > @nim65s Is it possible to propagate that to the summer school repo or should I make another PR ?
> 
> The branch for the summer-school has been modified to strip some things that were not useful for the summer school, but which would break features and options that can be otherwise used by crocoddyl users.
> 
> eg. `A.dot(B)` was replaced by `A @ B`, as it is easier to read for students during the summer school, as they had Python 3, but this would break Python2 compatibility if we merged that in the main repo. Or another e.g., some code that lets user choose between gepetto-viewer and meshcat was removed, as we chose for them to use meshcat, and they didn't need to be cluttered by code which is out of topic.
> 
> Then, I don't know what @cmastalli wants to do with this branch:
> 
>     * we can choose to merge it, and force all notebook user to use python 3 & meshcat
> 
>     * we can maintain both branch in parallel, and try to ensure that each change on one gets reflected on the other
> 
>     * we can declare that summer-school branch as archived
> 
>     * we can just put a few fixes when we find those, but overall try not to modify it too much
> 
>     * etc.

What do you suggest? Probably the easiest thing to do is to first merge the summer-school branch, and then to rebase / to pull it into this branch.

nim65s:
I don't have any particular preference.

cmastalli:
@paLeziart could you format the code? To do it, you just need to run `yapf -ri .` in the root folder. Be aware that you will need v0.28.0 of yapf

paLeziart:
Ok I will try to format it.

cmastalli:
I will merge this code as it is since it just needed to format the code.

I will format it myself there.

Issue #795, opened 2020-07-02
Memory leak issue


gfadini:
I encountered this problem since I have to solve a great number of crocoddyl fddp problems in loop.
I found that after some iterations during an optimization routine the memory consumption increased significantly.

To test this I'm attaching some examples of the robot manipulation in a loop.
https://gist.github.com/gfadini/df00b6d8a245acf40d68771403946eb8
In one of them the problem is initialized and solved, in the other initialized and assigned to a variable and then solved.
In both case there seem to be the same unexpected behaviour.
To isolate the issue I put the pinocchio model initialization outside the loops and try to call the del() method on the ddp class.
This doesn't same to change the outcome that the memory allocation to the ddp problem is not freed.
If you can reproduce the examples, you should see the increase of memory consumption with top, htop, mprofile or similia.

I'm not sure if it's related to crocoddyl itself or the python bindings but I'd need to have this issue checked since it's quite limiting in my study case.

EDIT:
added the plots of memory vs progress
![memprof](https://user-images.githubusercontent.com/56651708/86442594-5450e180-bd0e-11ea-9a8b-3df416c85472.png)
![memprof2](https://user-images.githubusercontent.com/56651708/86442597-55820e80-bd0e-11ea-8abc-49438cc0e66b.png)

jcarpent:
This might be related to https://github.com/stack-of-tasks/eigenpy/issues/156. Please consider including `eigenpy/eigenpy.hpp` inside a `fwd.hpp` for the Python bindings.
The issue is coming from Boost.Python which has a specific way of destroying the memory of allocated objects. In the present context, we need to overload the destroying of `Eigen::MatrixBase` which is only done by including the mentioned file.

gfadini:
@jcarpent Hi, thanks for the quick reply. So to apply this fix I should compile eigenpy from scratch?
At the moment I was still working with the up-to-date binary version of crocoddyl for convenience.
I also mention that in my original problem I also need to continuously modify the pinocchio model in the loop so that too would be another possible source of leak if I don't apply the fix in the issue you linked https://github.com/stack-of-tasks/eigenpy/issues/156
Am I correct on this guess?

nim65s:
@gfadini : I can try to reproduce your issue and check @jcarpent fix if you prefer

jcarpent:
No, Python bindings of Croco should include the header that I've mentioned.

gfadini:
[...]

PR #800, opened 2020-07-17
Included eigenpy header to avoid memory leak

cmastalli:
It turns out that there was only a single place that we were not including the eigenpy header. This hopefully should fix the issue #795. @gfadini could you try this branch and report if this fixes your issue?

### DIFF
--- a/bindings/python/crocoddyl/core/actuation/actuation-squashing.cpp
+++ b/bindings/python/crocoddyl/core/actuation/actuation-squashing.cpp
@@ -7,7 +7,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
 #include "python/crocoddyl/core/core.hpp"
-
 #include "crocoddyl/core/actuation/actuation-squashing.hpp"
 #include "crocoddyl/core/utils/exception.hpp"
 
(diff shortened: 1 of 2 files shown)


cmastalli:
> For effectiveness and to avoid other memory leaks in the future, we need to add the eigenpy header in a fwd.hpp file and include it everywhere. This feature is also requested for Windows.

This is already done. The logic is the follows: 
  1. Include eigenpy header in main fwp.hpp (https://github.com/loco-3d/crocoddyl/blob/devel/bindings/python/crocoddyl/fwd.hpp#L12)
  2. Include fwp in core.hpp and multibody.hpp (https://github.com/loco-3d/crocoddyl/blob/devel/bindings/python/crocoddyl/core/core.hpp#L12  and https://github.com/loco-3d/crocoddyl/blob/devel/bindings/python/crocoddyl/multibody/multibody.hpp#L13)
 3. Include core.hpp (or multibody.hpp) in each core (or multibody) cpp file

This was followed in each file with the exception of the state cpp. Not sure how this happened.

cmastalli:
@gfadini let me know if this solved your issue.
In the meantime, I will merge this PR.

gfadini:
Hi, I tested the issue #795 with this modification, but it seems that there's no improvement, I've included more details in the issue

cmastalli:
> Hi, I tested the issue #795 with this modification, but it seems that there's no improvement, I've included more details in the issue

Thanks for the reply. To be honest, this is the result that I was expecting.

PR #803, opened 2020-07-23
Fixed equations in quadratic-barrier python doc

julesser:


bindings/python/crocoddyl/core/activations/quadratic-barrier.cpp
```diff
@@ -35,8 +35,8 @@ void exposeActivationQuadraticBarrier() {
       "The activation is zero when r is between the lower (lb) and upper (ub) bounds, beta\n"
       "determines how much of the total range is not activated. This is the activation\n"
       "equations:\n"
-      "a(r) = 0.5 * ||r||^2 for lb < r < ub\n"
-      "a(r) = 0. for lb >= r >= ub.",
+      "a(r) = 0.5 * ||r||^2 for lb >= r >= ub\n"
```
jcarpent, line 38: lb means lower bound. So the comparison operator is incorrect.

cmastalli: Yes, it should be
```cpp
"a(r) = 0.5 * ||r||^2 for lb > r > ub\n"
"a(r) = 0. for lb <= r <= ub.",
```

julesser: Done

cmastalli:
@julesser thank you fixing this error in the documentation :)

PR #806, opened 2020-07-27
Release/1.4.0

cmastalli:
This new release includes the following new features:
  - Allowed different values of `nu` in action models
  - Adapted the contact force cost to accept 3d contacts.
  - Added CppAD support to (weighted) quadratic barrier activation
  - Added the contact impulse cost
  - Added contact CoP cost,
fixes:
  - Fixed display of contact forces in Gepetto viewer
  - Fixed a bug in the definition of the terminal node in `updateNode` (defined in `ShootingProblem`)
  - Fixed memory allocation produced by Python binds.
  - Fixed printed error message in `ShootingProblem`
and improvements:
  - Added documentation of the differential action model
  - Deprecated legacy functions to define reference in cost functions.
  - Set to zero inactive contact forces for correctness
  - Simplified FDDP code by using `calcDiff` from DDP solver
  - Improved the display of benchmark timings
  - Improved the documentation of contact force cost
  - Improved few notebooks and added an extra one
  - Removed legacy Python unittest


cmastalli:
@nim65s do not merge it before checking everything from my side :)

cmastalli:
@nim65s This branch is ready to be merged. I had to fix a few issues in the examples and notebooks.

The commits that fixes these issues are merged (or in its process to be merged) into the devel branch.

nim65s:
Thanks. I'll update the CMake config and release this today.

nim65s:
That's ready for my side.

I found an issue when building in multiprocess and with python bindings, but it seems that it is fixed on devel, so I guess this should do.

We could wait for a better handling of the Boost Python export at the jrl-cmakemodule level, but I don't think that many people would build a package that depend on crocoddyl, on 16.04 and with python 3. For the same issue with Arch, I can just embed a `sed` to fix that in the `PKGBUILD`.

Other parts of the CMake module have changed:
- we don't set `CMAKE_CXX_STANDARD` directly anymore, there is a prefered check on the configuration of the current compiler and the user defined option to see if this is really necesary
- CppAD & codegen are found using custom CMake modules

And I've switched to yapf 0.29.0, as gepetto/linters seems to have done this by itself.

cmastalli:
> We could wait for a better handling of the Boost Python export at the jrl-cmakemodule level, but I don't think that many people would build a package that depend on crocoddyl, on 16.04 and with python 3. For the same issue with Arch, I can just embed a `sed` to fix that in the `PKGBUILD`.

My vote is to not wait because it would be good to have this release sooner. There are some deprecated functions that needs to be propagated in our community :)
In any case, I am planning the next time to have a much faster release with other deprecations (and this cmake thing).

> And I've switched to yapf 0.29.0, as gepetto/linters seems to have done this by itself.

I don't remember now. Are we using the same yapf version in Travis and Gepgitla buildfarm? I know we have some discrepancies that it would be good to agree.

nim65s:
0.29.0 was automatically updated on gitlab, so I updated the code in consequence in 4cb59a5 and updated the travis config in aab2e12.

cmastalli:
> 0.29.0 was automatically updated on gitlab, so I updated the code in consequence in [4cb59a5](https://github.com/loco-3d/crocoddyl/commit/4cb59a58428c53f3f058252a7d52ebcea8d31f77) and updated the travis config in [aab2e12](https://github.com/loco-3d/crocoddyl/commit/aab2e12c1d119a4cc81845a89837502af1a7e374).

Thank you

PR #809, opened 2020-07-27
Update benchmarks, quasi-static

proyan:
This PR:
1) updates the codegen benchmarks for legged robots, and removes the duplicate code.
2) adds the quasi-static implementation for the diff-action-models


benchmark/factory/arm-kinova.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+  // Next, we need to create an action model for running and terminal knots. The
+  // forward dynamics (computed using ABA) are implemented
+  // inside DifferentialActionModelFullyActuated.
+  boost::shared_ptr<DifferentialActionModelFreeFwdDynamics> runningDAM =
+      boost::make_shared<DifferentialActionModelFreeFwdDynamics>(state, actuation, runningCostModel);
+
+  // VectorXs armature(state->get_nq());
+  // armature << 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.;
+  // runningDAM->set_armature(armature);
```
cmastalli, line 82: You might want to benchmark the case with armature as well.
As you can see, it will run another part of the code defined in the action model.

benchmark/factory/legged-robots.hpp
```diff
@@ -115,11 +128,11 @@ void build_biped_action_models(boost::shared_ptr<crocoddyl::ActionModelAbstractT
       boost::make_shared<DifferentialActionModelContactFwdDynamics>(state, actuation, contact_models,
                                                                     terminalCostModel);
 
-  runningModel = boost::make_shared<IntegratedActionModelEuler>(runningDAM, Scalar(1e-3));
-  terminalModel = boost::make_shared<IntegratedActionModelEuler>(terminalDAM, Scalar(1e-3));
+  runningModel = boost::make_shared<IntegratedActionModelEuler>(runningDAM, Scalar(5e-3));
```
cmastalli, line 131: These `dt` are sort of big, they might trigger a few more iterations. 
I suggest you to keep it as before, so we encourage to bench a single backward/forward passes in the solvers.

benchmark/legged-robots-codegen.cpp
```diff
@@ -284,7 +289,7 @@ int main(int argc, char* argv[]) {
   // Forward pass timings
   for (unsigned int i = 0; i < T; ++i) {
     crocoddyl::Timer timer;
-    cg_ddp.forwardPass(0.5);
+    cg_ddp.forwardPass(0.005);
```
cmastalli, line 292: Indeed, it seems you reduced alpha because a big `dt` value as commented above.

bindings/python/crocoddyl/core/actuation-base.hpp
```diff
@@ -44,6 +44,10 @@ class ActuationModelAbstract_wrap : public ActuationModelAbstract, public bp::wr
   }
 
   boost::shared_ptr<ActuationDataAbstract> default_createData() { return this->ActuationModelAbstract::createData(); }
+  void get_actuated(const boost::shared_ptr<ActuationDataAbstract>& data, Eigen::Ref<Eigen::VectorXd>& u) {
```
cmastalli, line 47: Is this a pure virtual method? If not, this is not the proper way of bind it.

bindings/python/crocoddyl/core/diff-action-base.hpp
```diff
@@ -57,10 +57,22 @@ class DifferentialActionModelAbstract_wrap : public DifferentialActionModelAbstr
 
   boost::shared_ptr<DifferentialActionDataAbstract> default_createData() {
     return this->DifferentialActionModelAbstract::createData();
-  }
-};
+    void quasiStatic(const boost::shared_ptr<DifferentialActionDataAbstract>& data, Eigen::Ref<Eigen::VectorXd> u,
```
cmastalli, line 60: You should bind it as virtual method, not pure virtual.
We want to run the default `quasiStatic` is it is not defined in any derived Python / c++ class.

(see the `createData` example). Note that you will need to bind `quasiStatic` in the derived c++ classes as well. This is due to the fact that in Python the base classes become `XXX_wrap`.

bindings/python/crocoddyl/core/diff-action-base.hpp
```diff
@@ -57,10 +57,22 @@ class DifferentialActionModelAbstract_wrap : public DifferentialActionModelAbstr
...
+                     << "u has wrong dimension (it should be " + std::to_string(nu_) + ")");
+      }
+      return bp::call<void>(this->get_override("quasiStatic").ptr(), data, (Eigen::VectorXd)u, (Eigen::VectorXd)x,
+                            (std::size_t)maxiter, (double)tol);
+    }
+  };
 
 }  // namespace python
-}  // namespace crocoddyl
+}  // namespace python
```
cmastalli, line 76: Revert this

bindings/python/crocoddyl/core/diff-action-base.hpp
```diff
@@ -57,10 +57,22 @@ class DifferentialActionModelAbstract_wrap : public DifferentialActionModelAbstr
 
   boost::shared_ptr<DifferentialActionDataAbstract> default_createData() {
     return this->DifferentialActionModelAbstract::createData();
-  }
-};
+    void quasiStatic(const boost::shared_ptr<DifferentialActionDataAbstract>& data, Eigen::Ref<Eigen::VectorXd> u,
+                     const Eigen::Ref<const Eigen::VectorXd>& x, const std::size_t& maxiter, const double& tol) {
```
cmastalli, line 61: You should also define the default argument of `maxiter`, `tol` as done in the c++ interface. This will require to adapt the `.def` function in the `.cpp`

include/crocoddyl/core/actions/diff-lqr.hpp
```diff
@@ -39,6 +39,10 @@ class DifferentialActionModelLQRTpl : public DifferentialActionModelAbstractTpl<
   virtual boost::shared_ptr<DifferentialActionDataAbstract> createData();
   virtual bool checkData(const boost::shared_ptr<DifferentialActionDataAbstract>& data);
 
+  virtual void quasiStatic(const boost::shared_ptr<DifferentialActionDataAbstract>& data, Eigen::Ref<VectorXs> u,
```
cmastalli, line 42: Why do you implement it as empty function?
If you do not plan to do it, then remove this function declaration so it will run the default `quasiStatic` method.

include/crocoddyl/core/actuation-base.hpp
```diff
@@ -52,6 +52,8 @@ class ActuationModelAbstractTpl {
   const std::size_t& get_nu() const { return nu_; };
   const boost::shared_ptr<StateAbstract>& get_state() const { return state_; };
 
+  virtual void get_actuated(const boost::shared_ptr<ActuationDataAbstract>& data, Eigen::Ref<VectorXs>& u) = 0;
```
cmastalli, line 55: I don't see the objective of this function, could you explain me?

include/crocoddyl/core/diff-action-base.hpp
```diff
@@ -118,6 +118,9 @@ class DifferentialActionModelAbstractTpl {
    */
   void calcDiff(const boost::shared_ptr<DifferentialActionDataAbstract>& data, const Eigen::Ref<const VectorXs>& x);
 
+  virtual void quasiStatic(const boost::shared_ptr<DifferentialActionDataAbstract>& data, Eigen::Ref<VectorXs> u,
```
cmastalli, line 121: Why are you declaring it as pure virtual when in the base is declare as virtual?
As before, if you don't plan to implement it, then remove this function declaration.

include/crocoddyl/core/numdiff/actuation.hpp
```diff
@@ -80,6 +80,10 @@ class ActuationModelNumDiffTpl : public ActuationModelAbstractTpl<_Scalar> {
    */
   void set_disturbance(const Scalar& disturbance);
 
+  virtual void get_actuated(const boost::shared_ptr<ActuationDataAbstract>& data, Eigen::Ref<VectorXs>& u) {
```
cmastalli, line 83: This function is declared as virtual. If you expect some classes that won't declare it, then (1) this message should go to the base class and (2) this function declaration should be removed

include/crocoddyl/core/numdiff/diff-action.hpp
```diff
@@ -38,6 +38,9 @@ class DifferentialActionModelNumDiffTpl : public DifferentialActionModelAbstract
   virtual void calcDiff(const boost::shared_ptr<DifferentialActionDataAbstract>& data,
                         const Eigen::Ref<const VectorXs>& x, const Eigen::Ref<const VectorXs>& u);
   virtual boost::shared_ptr<DifferentialActionDataAbstract> createData();
+  virtual void quasiStatic(const boost::shared_ptr<DifferentialActionDataAbstract>& data, Eigen::Ref<VectorXs> u,
```
cmastalli, line 41: Same here: remove it or implement it

include/crocoddyl/multibody/actions/contact-fwddyn.hxx
```diff
@@ -163,6 +164,32 @@ DifferentialActionModelContactFwdDynamicsTpl<Scalar>::createData() {
...
+  DifferentialActionDataContactFwdDynamicsTpl<Scalar>* d =
+      static_cast<DifferentialActionDataContactFwdDynamicsTpl<Scalar>*>(data.get());
+  const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> q = x.head(state_->get_nq());
+  const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> v = x.tail(state_->get_nv());
+
+  // Check the velocity input is zero
+  assert_pretty(v.isZero(), "The velocity input should be zero for quasi-static to work.");
+
+  d->multibody.actuation->tau = pinocchio::rnea(pinocchio_, d->pinocchio, q, v, VectorXs::Zero(state_->get_nv()));
+  actuation_->get_actuated(d->multibody.actuation, u);
```
cmastalli, line 189: Again I don't see why we need this function.

include/crocoddyl/multibody/actions/contact-fwddyn.hxx
```diff
@@ -163,6 +164,32 @@ DifferentialActionModelContactFwdDynamicsTpl<Scalar>::createData() {
   return boost::allocate_shared<Data>(Eigen::aligned_allocator<Data>(), this);
 }
 
+template <typename Scalar>
+void DifferentialActionModelContactFwdDynamicsTpl<Scalar>::quasiStatic(
```
cmastalli, line 168: Did you create a unittest?
We should check it against the default implementation that I have coded quite time ago.

include/crocoddyl/multibody/actions/contact-fwddyn.hxx
```diff
@@ -163,6 +164,32 @@ DifferentialActionModelContactFwdDynamicsTpl<Scalar>::createData() {
...
+  // Static casting the data
+  DifferentialActionDataContactFwdDynamicsTpl<Scalar>* d =
+      static_cast<DifferentialActionDataContactFwdDynamicsTpl<Scalar>*>(data.get());
+  const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> q = x.head(state_->get_nq());
+  const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> v = x.tail(state_->get_nv());
+
+  // Check the velocity input is zero
+  assert_pretty(v.isZero(), "The velocity input should be zero for quasi-static to work.");
+
+  d->multibody.actuation->tau = pinocchio::rnea(pinocchio_, d->pinocchio, q, v, VectorXs::Zero(state_->get_nv()));
```
cmastalli, line 188: Indeed, this `quasiStatic` implementation seems wrong for me. It doesn't consider the active contacts

include/crocoddyl/multibody/actions/free-fwddyn.hxx
```diff
@@ -132,6 +133,31 @@ bool DifferentialActionModelFreeFwdDynamicsTpl<Scalar>::checkData(
     return false;
   }
 }
+template <typename Scalar>
+void DifferentialActionModelFreeFwdDynamicsTpl<Scalar>::quasiStatic(
```
cmastalli, line 137: Same here: we need unittest

include/crocoddyl/multibody/actuations/floating-base.hpp
```diff
@@ -44,6 +44,14 @@ class ActuationModelFloatingBaseTpl : public ActuationModelAbstractTpl<_Scalar>
     data->tau.tail(nu_) = u;
   };
 
+  virtual void get_actuated(const boost::shared_ptr<Data>& data, Eigen::Ref<VectorXs>& u) {
```
cmastalli, line 47: This can be obtained with `calc`.
If this is the reason, then we should remove and keep clean the API.

include/crocoddyl/multibody/actuations/full.hpp
```diff
@@ -44,6 +44,14 @@ class ActuationModelFullTpl : public ActuationModelAbstractTpl<_Scalar> {
     data->tau = u;
   };
 
+  virtual void get_actuated(const boost::shared_ptr<Data>& data, Eigen::Ref<VectorXs>& u) {
```
cmastalli, line 47: same here. Let's remove it!

include/crocoddyl/multibody/actuations/multicopter-base.hpp
```diff
@@ -63,6 +63,10 @@ class ActuationModelMultiCopterBaseTpl : public ActuationModelAbstractTpl<_Scala
 
     data->tau.noalias() = tau_f_ * u;
   }
+  virtual void get_actuated(const boost::shared_ptr<Data>& data, Eigen::Ref<VectorXs>& u) {
```
cmastalli, line 66: and here

include/crocoddyl/multibody/frames.hpp
```diff
@@ -137,7 +137,7 @@ struct FrameFrictionConeTpl {
 };
 
 template <typename _Scalar>
-class FrameCoPSupportTpl {
+struct FrameCoPSupportTpl {
```
cmastalli, line 140: Revert it back. This is a class became we need to protect the wrong use of some internal members, there is a reason for this. (Please keep focus in the goal of your PR)

include/crocoddyl/multibody/actuations/floating-base.hpp
proyan: get_actuated is the inverse of calc. Look at the implementation before you review.

include/crocoddyl/multibody/frames.hpp
proyan: The "goal" of this PR doesn't work, if there are some bad changes already introduced in the `devel` that produce errors. `FrameCoPSupport` has been forward declared as a struct, like _ALL_ other frames. Be consistent.

benchmark/factory/arm-kinova.hpp
proyan: We're moving armature would move to pinocchio, so we don't have to maintain it here.

benchmark/factory/legged-robots.hpp
proyan: This is the standard time step used in simulations. Ideally, we should be benching with even higher `dt`, to show the preview we can obtain. But I've kept it as this since it is widely used.

benchmark/legged-robots-codegen.cpp
proyan: `5e-3` is not big. If this time step doesn't work, it is because the problem is ill-defined, not because the time step is too high.

include/crocoddyl/multibody/actions/contact-fwddyn.hxx
proyan: Yes, it doesn't. I need to fix that.

proyan:
@nim65s I get the following error: 

```cmake
CMake Error at cmake/python.cmake:406 (ADD_CUSTOM_COMMAND):
  TARGET 'compile_pyc' was not created in this directory.
Call Stack (most recent call first):
  benchmark/CMakeLists.txt:46 (PYTHON_BUILD)
```

when compiling with the current cmake-module. I believe it has to do with the way we are implementing `PYTHON_BUILD` in crocoddyl. Do you have an insight?

benchmark/factory/arm-kinova.hpp
cmastalli: Sounds good!

benchmark/legged-robots-codegen.cpp
cmastalli: I agree with it. I referred to `dt=5e-2` as "big" in my other comment (probably I read it wrong).

include/crocoddyl/multibody/frames.hpp
cmastalli: This is not a bad change. Take time to understand the reason of this choice. However, in short, we need to have private members in order to avoid user mistakes.

The other frameXXX don't need to make private members.

---
Consider that if you modify "bad" (or not) changes that don't correspond to the main goal of a PR, then it makes my job more difficult and slow since I need to pay attention to each line.

Typically, you made huge PRs. This is not good / safe for the project. There many things that I might miss, and it is the same for you. Then, in future, when I track merged PRs, I might lost these bits.

It is easy for you to make independent PRs when you found things that you believe we should change. I would be happy if you make that effort for me :)

proyan: Carlos, Please first read what my comment was, before replying.

`FrameCopSupport`, like all other frames, has been forward declared as a `struct` here: https://github.com/loco-3d/crocoddyl/blob/devel/include/crocoddyl/multibody/fwd.hpp#L78. Re-declaring it as a `class` produces ABI errors. I think it is better to keep this as a struct, like all other frames, and define the private and public members explicitly, as is being done.

> It is easy for you to make an independent PR when you found things that you believe we should change

Implementation errors are not always avoidable. However many of your PRs in crocoddyl are merged really quickly without any review _at all_. This introduces a lot of them, and it is diffcult to continuously branch everytime I find a problem. This was a minor issue, and the change proposed was _consistent with the API_.

Finally, I like 50% of your reviews, which often provide constructive remarks. However, starting a review with "Revert it back". or "Remove this.", when you don't understand what the change is, is counter-productive and leads to frivolous debates like this. Let's avoid that in the future :).

proyan:
@cmastalli I think you would be much faster than me in implementing the bindings for quasi-static as a virtual function. Do you think you can take care of it?

proyan:
@nim65s It seems like I'm getting formatting errors for files that are not touched in this PR (https://travis-ci.org/github/loco-3d/crocoddyl/jobs/713375309#L327)

cmastalli:
> @cmastalli I think you would be much faster than me in implementing the bindings for quasi-static as a virtual function. Do you think you can take care of it?

I could help you on Monday, today I am overbooked. Let me know if you want me to work on Monday on this topic

benchmark/arm-kinova-codegen.cpp
```diff
@@ -0,0 +1,369 @@
...
+  cg_ddp.setCandidate(xs, us, false);
+  boost::shared_ptr<crocoddyl::ActionDataAbstract> cg_runningData = cg_runningModel->createData();
+  boost::shared_ptr<crocoddyl::ActionDataAbstract> runningData = runningModel->createData();
+  Eigen::VectorXd x_rand = cg_runningModel->get_state()->rand();
+  Eigen::VectorXd u_rand = Eigen::VectorXd::Random(cg_runningModel->get_nu());
+  runningModel->calc(runningData, x_rand, u_rand);
+  runningModel->calcDiff(runningData, x_rand, u_rand);
+  cg_runningModel->calc(cg_runningData, x_rand, u_rand);
+  cg_runningModel->calcDiff(cg_runningData, x_rand, u_rand);
+  assert_pretty(cg_runningData->xnext.isApprox(runningData->xnext), "Problem in xnext");
```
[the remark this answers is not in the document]
proyan: Yes, they are good for debugging the benches

bindings/python/crocoddyl/core/actuation-base.hpp
proyan: Thanks for the help in fixing it.

include/crocoddyl/core/actuation-base.hpp
proyan: It is the inverse of the calc, in order to obtain the actuated control given the final actuation tau.

include/crocoddyl/multibody/actions/contact-fwddyn.hxx
proyan: discussed above

proyan:
I'm waiting for https://github.com/jrl-umi3218/jrl-cmakemodules/pull/426 to be merged. Afterwards, we can update the cmake modules and merge this PR.

This PR introduces a temporary fix in https://github.com/loco-3d/crocoddyl/pull/809/files#diff-5b71ffe86b6c7bf2fea1367edf59eb93R14. This is already fixed in the current pinocchio/devel, and would not be needed in future releases. Once pinocchio v2.4.8 is released, we should remove this line.

cmastalli:
@proyan I didn't finish the python bindings. I have cleaned up the code, removed the unnecessary `get_actuated`, and fixed issues (regarding different inactive contacts and general under-actuation dimension).

A part of the bindings, it is needed to debug why it is crashing this branch for the quadrupedal examples (I suspected it's related with impulse models). I will try to debug myself tomorrow, so hopefully you won't need to do anything else.

proyan:
@cmastalli I would really appreciate it if you don't overwrite on changes that I introduce without any discussion. This is frustrating to deal with. If you feel like there is some stuff to be done, you can review the PR, or discuss privately.
Please don't single-handedly revert back changes based on what you feel is right. Moreover, you have now introduced bugs in the implementation, that I have to deal with.

cmastalli:
> @cmastalli I would really appreciate it if you don't overwrite on changes that I introduce without any discussion. This is frustrating to deal with. If you feel like there is some stuff to be done, you can review the PR, or discuss privately.
> Please don't single-handedly revert back changes based on what you feel is right. Moreover, you have now introduced bugs in the implementation, that I have to deal with.

I did tell you that we need to remove `get_actuated`, there is no need to add this function and make the API more complex. I have no changed anything else. The other changes are basically cleaning the code that were producing plenty of warning.

proyan:
Carlos we have a modular code, where the type of actuation doesn't depend on the action model being used. Unfortunately, you would remove this requirement. 

However, I don't see any discussion about `get_actuated`. You told me to revert it back without any reason, and then you told me to do it because you thought "it was the same as calc". Both of these are not true. It is not nice to simply go and revert changes that you don't understand on a PR that someone proposed!

cmastalli:
I will let you work on this as you prefer but this PR won't be merged if you don't remove `get_actuated`. This function was not needed before in our Python legacy code. You could check my statement here: https://github.com/loco-3d/crocoddyl/blob/master/unittest/python/crocoddyl/floating_contact.py#L117

include/crocoddyl/multibody/fwd.hpp
```diff
@@ -299,6 +299,10 @@ typedef CostDataFrameRotationTpl<double> CostDataFrameRotation;
 typedef ImpulseModelAbstractTpl<double> ImpulseModelAbstract;
 typedef ImpulseDataAbstractTpl<double> ImpulseDataAbstract;
 
+enum ContactType { Contact3D, Contact6D, ContactUndefined };
+
+enum ImpulseType { Impulse3D, Impulse6D, ImpulseUndefined };
```
jcarpent, line 303: I don't think this one is really needed.
The contact can be either 2D, 3D, 6D, etc. and can be solved using a Force, Impulse or Impact scheme.

proyan: You are right, we don't need these two separate enums. Currently there is additional support of Contact2D in another PR, while 2D impulse is not yet supported. I need to add support of Impulse2D as well, and then I can merge these two together.

include/crocoddyl/multibody/actions/contact-fwddyn.hxx
```diff
@@ -192,15 +192,17 @@ void DifferentialActionModelContactFwdDynamicsTpl<Scalar>::quasiStatic(
       pinocchio::rnea(pinocchio_, d->pinocchio, q, VectorXs::Zero(state_->get_nv()), VectorXs::Zero(state_->get_nv()));
 
   contacts_->calc(d->multibody.contacts, x);
+  // Allocates memory
   MatrixXs jc_view = d->multibody.contacts->Jc.template leftCols<6>();
```
jcarpent: Why to not use a view directly?

proyan: I couldn't get `Eigen::Map`, or `Eigen::Block` to work with `JacobiSVD` module.

include/crocoddyl/core/diff-action-base.hxx
```diff
@@ -31,6 +31,37 @@ void DifferentialActionModelAbstractTpl<Scalar>::calc(const boost::shared_ptr<Di
...
+  const std::size_t& ndx = state_->get_ndx();
+  VectorXs dx = VectorXs::Zero(ndx);
+  if (nu_ == 0) {
+    // TODO(cmastalli): create a method for autonomous systems
+  } else {
+    VectorXs du = VectorXs::Zero(nu_);
+    for (std::size_t i = 0; i < maxiter; ++i) {
+      calc(data, x, u);
+      calcDiff(data, x, u);
+      du = -pseudoInverse(data->Fu) * data->xout;
```
jcarpent, line 56: You should use the noalias here.

cmastalli:
@proyan I am very unhappy with your behavior, and it won't help if you don't reflect.
You merged this PR without my consent.

@nim65s what should I do? Do I really need to remove his rights again?

@proyan we need to solve this issue, otherwise I would need to take a hard decision with you.

cmastalli:
Just to clarify, a consent requires a review.
Even if you modify the code as I expected, you should wait (or ping me) to review this code.
Doing the opposite brings more commutation issues.

nim65s:
I wasn't expecting that. Just to be clear, @cmastalli, you didn't consent on that `get_actuated` function, right ? Or is there any other technical issue ?

cmastalli:
> I wasn't expecting that. Just to be clear, @cmastalli, you didn't consent on that `get_actuated` function, right ? Or is there any other technical issue ?

I didn't have time to discuss with him neither review this PR.
I am having a quick look now, it seems there is no `get_actuated`.
However, if this the case, it requires to review again the re-implementation of `quasiStatic`.
I need time for it, and it is important since this function might break many codes.

nim65s:
Ok, then, for the public & technical part, I suggest that you take your time, and when you're ready, open another PR to propose your fixes.

For the rest, let's go back to private mails :)

PR #811, opened 2020-07-28
c++98 compatibility of deprecate macro

cmastalli:
Hi @proyan 

These modifications should fixed the issue. However, I couldn't test it fully since there is a problem with Pinocchio.

@nim65s we should merge this into the release topic, when this PR gets approved


cmastalli:
@nim65s I have picked this commit in the release/1.4.0 branch

nim65s:
thanks :)

Issue #812, opened 2020-07-28
`WORLD` and `LOCAL_WORLD_ALIGNED` reference for force / impulse cost functions


cmastalli:
It is important to allow users to define various reference for contact force / impulse cost functions as well. This will require tricky conversion of forces --and their derivatives-- from `LOCAL` to `WORLD` and `LOCAL_WORLD_ALIGNED`. Note that our contact dynamics computes everything at the LOCAL level.

This task could be done potentially done inside Pinocchio (i.e. the new contact dynamics). @proyan, @jcarpent  have you though about it? Could you add this in the new contact dynamics of Pinocchio? or do we need to handle it here?

In any case, @jcarpent could you help me with this task?

cmastalli:
This is handled in #1123.

PR #813, opened 2020-07-28
Frame velocity cost accepts `WORLD` and `LOCAL_WORLD_ALIGNED` references as well

cmastalli:
Before this PR, we have coded a function that was able to track only the `LOCAL` reference. However, it is convenient to have different options (in particular the `LOCAL_WORLD_ALIGNED` reference).

This PR proposes:
  1. To add the `pinocchio::FrameReference` inside `crocoddyl::FrameMotion`
  2. To keep the API by defining `LOCAL` as default argument in 1
  3. To update the needed Pinocchio version since it requires new routines.
  4. To make unittest only of `LOCAL` costs (the other cases are tested inside Pinocchio)

In future, we need to do this task for force / impulse cost functions. For more details see #812.


bindings/python/crocoddyl/multibody/frames.cpp
```diff
@@ -56,13 +56,16 @@ void exposeFrames() {
...
                           "It defines a frame motion (tangent of SE(3) point) for a given frame ID",
-                          bp::init<FrameIndex, pinocchio::Motion>(bp::args("self", "frame", "oMf"),
-                                                                  "Initialize the frame motion.\n\n"
-                                                                  ":param frame: frame ID\n"
-                                                                  ":param oMf: Frame motion w.r.t. the origin"))
+                          bp::init<FrameIndex, pinocchio::Motion, bp::optional<pinocchio::ReferenceFrame> >(
+                              bp::args("self", "frame", "oMf", "reference"),
+                              "Initialize the frame motion.\n\n"
+                              ":param frame: frame ID\n"
+                              ":param oMf: Frame motion w.r.t. the origin\n"
```
proyan, line 63: oMf is used in pinocchio to denote SE3 objects. It's not a good name for a motion reference

bindings/python/crocoddyl/multibody/frames.cpp
```diff
@@ -56,13 +56,16 @@ void exposeFrames() {
...
+                              ":param reference: Reference frame (default LOCAL)"))
```
proyan, line 64: similarly, when you see the name "reference", it means the vector that should be given. Use the name reference_frame please.

bindings/python/crocoddyl/multibody/frames.cpp
proyan: If you are saving the reference frame in a separate object, it is better to not use the base index like o or i to denote the frame where the motion vector is defined.

cmastalli: > oMf is used in pinocchio to denote SE3 objects. It's not a good name for a motion reference

Any suggested notation? I was though to use `omf` but it is hard to read.
As a side node, if we change this notation then we'll break the Python API.
So my question is "is this worth to do?"

> If you are saving the reference frame in a separate object, it is better to not use the base index like o or i to denote the frame where the motion vector is defined.

We don't save reference frame in other objects. The design choice is to be able to change reference frame as well.
Is this answer your question?

bindings/python/crocoddyl/multibody/costs/frame-velocity.cpp
```diff
@@ -92,17 +92,7 @@ void exposeCostFrameVelocity() {
           bp::args("self", "model", "data"),
           "Create frame velocity cost data.\n\n"
           ":param model: frame Velocity cost model\n"
-          ":param data: shared data")[bp::with_custodian_and_ward<1, 2, bp::with_custodian_and_ward<1, 3> >()])
-      .add_property("joint", bp::make_getter(&CostDataFrameVelocity::joint), "joint index")
```
proyan, line 96: I think it is better to have it as deprecated rather than removing completely.

bindings/python/crocoddyl/multibody/frames.cpp
proyan: Then I would suggest to deprecate this call and simply return the new motion object whenever this call is made. In the next release, we can finish the change of API.

> We don't save reference frame in other objects. The design choice is to be able to change reference frame as well.

What I mean is, Motion vector is defined in a base frame, and normally we write it like ov, or iv to say that it is denoted in frame o or i. But since you are saving a reference_frame as well inside this FrameMotion class, if you name this as ov, and then define the reference_frame to be LOCAL, that is conflicting. So I suggest simply to use a name without a base frame like o or i attached to it.

cmastalli:
@proyan let's discuss generally the API here. I could make all the changes in this PR if they're needed. These are my thoughts after you're feedback. I will make the following propositions:
  1. Keep `reference` as done (if you reads `frameMotion.reference`, then it is meaningful, right?)
  2. Rename `frame` to `id` (same motivation of 1)
  3. Rename `oMf` to things like `placement`, `motion`, `force`, `cone`.

By doing 2 and 3, I need to deprecate the Python methods that retrieve these objects. So for Python it is not a problem, but for c++ code I need to delete them. The good thing is the code won't compile so the user might to quickly fix the issue.

Do you agree with it?

bindings/python/crocoddyl/multibody/frames.cpp
cmastalli: Sorry the problem is not the constructor itself since it won't change the arguments (just their names). The problem is the member objects of the structure.

For instance, if we change `oMf` to `velocity` then we need to change the name of the member object. Let's my write a Python code to be more clear
```python
import crocoddyl
import pinocchio

frameId = 3
vref = crocoddyl.FrameMotion(frameId, pinocchio.Motion.Random(), pinocchio.LOCAL)
# This is the current names
print vref.oMf
# but the change of name in the constructor means to change it to
print vref.velocity
```

proyan: I think it should be possible to duplicate `vref.velocity` under the name `vref.oMf`. Then if the user calls `vref.oMf`, he get the deprecated message and the value of `vref.velocity`

proyan:
I don't think "reference" is a good name for a base frame. Let's use reference_frame to be sure what we are talking about.

cmastalli:
@proyan there are many changes to be done. I am correctly working on them.
Today, I will create another PR with them to make easy the review.

In the meantime I will merge this PR since the requested changes are regarding the API.

PR #814, opened 2020-07-28
[Python] remove debug with wrong syntax

nim65s:
Couldn't compile devel with python 3 otherwise.


nim65s:
It looks like this was unintentionnaly added in #808

Issue #807, opened 2020-07-27
General underactuated root actuation model


gfadini:
Hi,

I need a model of a robot for which there's an under-actuated base free to move along a prismatic guide.
Now one solution would be to develop a custom model from the ```ActuationAbstract``` and this would lead to writing a new actuation model either with bindings or in C++.
However I think that all this not necessary, and the class ```ActuationModelFloatingBase``` can be extended to include a more general under-actuated robot class. I guess that the change in this will also have a positive impact for other classes that would have by default the same C++ implementation.
I get the impression that it would be enough removing all the 1st joint DOFs avoiding the check for it to be a free-flyer.
Am I making a wrong assumption? Are there other parts of the code that would be affected by such change or doing internal checks?

cmastalli:
What it could be done is to parse the first joint of the URDF?
Then, we could be able to define any kind of floating-base joint (i.e. not only the free-flyer case).
Is this what you need?

gfadini:
Hi @cmastalli, yes that's roughly what I had in mind. With the URDF the thing should be the same, just that now I have a pinocchio model built from scratch. The class ```ActuationModelFloatingBase``` by itself can deal with a wider set of robots than what is implemented just by processing the model differently and neglecting always the first joint DOF

gfadini:
Hi @cmastalli, yes that's roughly what I had in mind. With the URDF the thing should be the same, just that now I have a pinocchio model built from scratch. The class ```ActuationModelFloatingBase``` by itself can deal with a wider set of robots than what is implemented just by processing the model differently and neglecting always the first joint DOF

cmastalli:
When is your deadline? or is this something that you feel you can modify inside Crocoddyl?

gfadini:
Hi @cmastalli at the moment I can simply prototype with the bindings,
In the floating-base.hpp there's a check for the first joint to be explicitly of the free flyer type
```
explicit ActuationModelFloatingBaseTpl(boost::shared_ptr<StateMultibody> state) : Base(state, state->get_nv() - 6) {
    pinocchio::JointModelFreeFlyerTpl<Scalar> ff_joint;
    if (state->get_pinocchio()->joints[1].shortname() != ff_joint.shortname()) {
      throw_pretty("Invalid argument: "
                   << "the first joint has to be free-flyer");
    }
  };
```
I think this can be generalized to also for the prismatic types, it was a general idea that would require little effort and that I may try

cmastalli:
[...]

PR #816, opened 2020-07-29
General underactuated base

gfadini:
This PR should be a solution to #807, I still have to properly test it on some robot whose base is not FloatingBase, but it doesn't change anything if it is

### DIFF
--- a/include/crocoddyl/multibody/actuations/floating-base.hpp
+++ b/include/crocoddyl/multibody/actuations/floating-base.hpp
@@ -26,12 +26,11 @@ class ActuationModelFloatingBaseTpl : public ActuationModelAbstractTpl<_Scalar> 
   typedef typename MathBase::VectorXs VectorXs;
   typedef typename MathBase::MatrixXs MatrixXs;
 
-  explicit ActuationModelFloatingBaseTpl(boost::shared_ptr<StateMultibody> state) : Base(state, state->get_nv() - 6) {
+  explicit ActuationModelFloatingBaseTpl(boost::shared_ptr<StateMultibody> state) : Base(state, state->get_nv() - state->get_pinocchio()->joints[1].nv()) {
     pinocchio::JointModelFreeFlyerTpl<Scalar> ff_joint;
     if (state->get_pinocchio()->joints[1].shortname() != ff_joint.shortname()) {
-      throw_pretty("Invalid argument: "
-                   << "the first joint has to be free-flyer");
-    }
+      std::cout << "Warning: the first joint should usually be of free-flyer type";
+    };
   };
   virtual ~ActuationModelFloatingBaseTpl(){};
 
@@ -52,8 +51,9 @@ class ActuationModelFloatingBaseTpl : public ActuationModelAbstractTpl<_Scalar> 
   };
 
   virtual boost::shared_ptr<Data> createData() {
+    boost::shared_ptr<StateMultibody> state = boost::dynamic_pointer_cast<StateMultibody> (state_);
     boost::shared_ptr<Data> data = boost::allocate_shared<Data>(Eigen::aligned_allocator<Data>(), this);
-    data->dtau_du.diagonal(-6).fill((Scalar)1.);
+    data->dtau_du.diagonal(-state->get_pinocchio()->joints[1].nv()).fill((Scalar)1.);
 #ifndef NDEBUG
     dtau_du_ = data->dtau_du;
 #endif
--- a/include/crocoddyl/multibody/actuations/floating-base.hpp
+++ b/include/crocoddyl/multibody/actuations/floating-base.hpp
@@ -25,9 +25,11 @@ class ActuationModelFloatingBaseTpl : public ActuationModelAbstractTpl<_Scalar> 
   typedef StateMultibodyTpl<Scalar> StateMultibody;
   typedef typename MathBase::VectorXs VectorXs;
   typedef typename MathBase::MatrixXs MatrixXs;
+  int nvFirstJoint;
 
   explicit ActuationModelFloatingBaseTpl(boost::shared_ptr<StateMultibody> state) : Base(state, state->get_nv() - state->get_pinocchio()->joints[1].nv()) {
     pinocchio::JointModelFreeFlyerTpl<Scalar> ff_joint;
+    nvFirstJoint = state->get_pinocchio()->joints[1].nv();
     if (state->get_pinocchio()->joints[1].shortname() != ff_joint.shortname()) {
       std::cout << "Warning: the first joint should usually be of free-flyer type";
     };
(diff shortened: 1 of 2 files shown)


include/crocoddyl/multibody/actuations/floating-base.hpp
```diff
@@ -25,13 +25,14 @@ class ActuationModelFloatingBaseTpl : public ActuationModelAbstractTpl<_Scalar>
   typedef StateMultibodyTpl<Scalar> StateMultibody;
   typedef typename MathBase::VectorXs VectorXs;
   typedef typename MathBase::MatrixXs MatrixXs;
+  int nvFirstJoint;
```
cmastalli, line 28: We do not need to create this object. I will provide a solution above

include/crocoddyl/multibody/actuations/floating-base.hpp
```diff
@@ -25,13 +25,14 @@ class ActuationModelFloatingBaseTpl : public ActuationModelAbstractTpl<_Scalar>
...
 
-  explicit ActuationModelFloatingBaseTpl(boost::shared_ptr<StateMultibody> state) : Base(state, state->get_nv() - 6) {
+  explicit ActuationModelFloatingBaseTpl(boost::shared_ptr<StateMultibody> state) : Base(state, state->get_nv() - state->get_pinocchio()->joints[1].nv()) {
     pinocchio::JointModelFreeFlyerTpl<Scalar> ff_joint;
+    nvFirstJoint = state->get_pinocchio()->joints[1].nv();
```
cmastalli, line 32: Remove this line as mentioned above

include/crocoddyl/multibody/actuations/floating-base.hpp
```diff
@@ -53,7 +54,7 @@ class ActuationModelFloatingBaseTpl : public ActuationModelAbstractTpl<_Scalar>
 
   virtual boost::shared_ptr<Data> createData() {
     boost::shared_ptr<Data> data = boost::allocate_shared<Data>(Eigen::aligned_allocator<Data>(), this);
-    data->dtau_du.diagonal(-6).fill((Scalar)1.);
+    data->dtau_du.diagonal(-nvFirstJoint).fill((Scalar)1.);
```
cmastalli, line 57: Change this code to:
```cpp
std::size_t fb_nv = state->get_pinocchio()->joints[1].nv();
data->dtau_du.diagonal(-fb_nv).fill((Scalar)1.);
```

gfadini: I tried that line before but it's not working, just not to add the variable I was not sure how to recover the state information I think I should use the ```state_``` there? Can you help me with this doubt?

proyan:
Hi @gfadini 
The actuationfloatingbase is built in coherence with the StateMultibody, and both of them assume that only the first joint can be a free-flyer.
I think you should fix the constructor of StateMultibody as well to be more generic. Could you take care of that in this PR?

include/crocoddyl/multibody/actuations/floating-base.hpp
cmastalli: Yes, that's correct.

include/crocoddyl/multibody/actuations/floating-base.hpp
```diff
@@ -26,12 +26,11 @@ class ActuationModelFloatingBaseTpl : public ActuationModelAbstractTpl<_Scalar>
   typedef typename MathBase::VectorXs VectorXs;
   typedef typename MathBase::MatrixXs MatrixXs;
 
-  explicit ActuationModelFloatingBaseTpl(boost::shared_ptr<StateMultibody> state) : Base(state, state->get_nv() - 6) {
+  explicit ActuationModelFloatingBaseTpl(boost::shared_ptr<StateMultibody> state) : Base(state, state->get_nv() - state->get_pinocchio()->joints[1].nv()) {
     pinocchio::JointModelFreeFlyerTpl<Scalar> ff_joint;
     if (state->get_pinocchio()->joints[1].shortname() != ff_joint.shortname()) {
```
cmastalli, line 31: Let's remove this warming message

include/crocoddyl/multibody/actuations/floating-base.hpp
```diff
@@ -50,10 +49,11 @@ class ActuationModelFloatingBaseTpl : public ActuationModelAbstractTpl<_Scalar>
     assert_pretty(data->dtau_dx == MatrixXs::Zero(state_->get_nv(), state_->get_ndx()), "dtau_dx has wrong value");
     assert_pretty(data->dtau_du == dtau_du_, "dtau_du has wrong value");
   };
-
+  boost::shared_ptr<StateMultibody> state = boost::dynamic_pointer_cast<StateMultibody> (state_);
```
cmastalli, line 52: Do not create this variable.
You can downcast it inside `createData`, i.e.
```cpp
virtual boost::shared_ptr<Data> createData() { 
  typedef StateMultibodyTpl<Scalar> StateMultibody;
  boost::shared_ptr<StateMultibody> state = boost::static_pointer_cast<StateMultibody>(state_);
  std::size_t fb_nv = state->get_pinocchio()->joints[1].nv();
...
}
```

include/crocoddyl/multibody/states/multibody.hxx
```diff
@@ -16,24 +16,17 @@ template <typename Scalar>
...
-    nq0 = 7;
-    lb_.template head<7>() = -std::numeric_limits<Scalar>::infinity() * VectorXs::Ones(7);
-    ub_.template head<7>() = std::numeric_limits<Scalar>::infinity() * VectorXs::Ones(7);
-  } else if (model->joints[1].shortname() == "JointModelSphericalZYX") {
-    joint_type_ = Spherical;
-  }
+
+  // Define internally the free-flyer limits by removing the first joint DOFs
+
+  const std::size_t nq0 = model->joints[1].nq();
```
cmastalli, line 29: Define the unbounded condition of the unactuated joint, i.e.
```cpp
    lb_.template head(nq0) = -std::numeric_limits<Scalar>::infinity() * VectorXs::Ones(nq0);
    ub_.template head(nq0) = std::numeric_limits<Scalar>::infinity() * VectorXs::Ones(nq0);
```

include/crocoddyl/multibody/states/multibody.hxx
```diff
@@ -16,24 +16,17 @@ template <typename Scalar>
...
-    joint_type_ = FreeFlyer;
-    // Define internally the free-flyer limits
-    nq0 = 7;
-    lb_.template head<7>() = -std::numeric_limits<Scalar>::infinity() * VectorXs::Ones(7);
-    ub_.template head<7>() = std::numeric_limits<Scalar>::infinity() * VectorXs::Ones(7);
-  } else if (model->joints[1].shortname() == "JointModelSphericalZYX") {
-    joint_type_ = Spherical;
-  }
+
+  // Define internally the free-flyer limits by removing the first joint DOFs
```
cmastalli, line 27: // Define internally the limits of the unactuated joint

gfadini:
You're welcome, thanks for the tips in modifying the code!

include/crocoddyl/multibody/states/multibody.hxx
proyan: You should not use the `template` keyword, when the call is not calling a template function. `lb_.head()` does not have a template argument, and this code should produce errors. I'm surprised that it passed CI!

cmastalli: Do you think this fails only with codegen compilation?

PR #817, opened 2020-07-30
2D contacts implementation for Solo 8 DoF

edantec:
Implementing 2D contact derivatives and python unittest. Contact force is considered to be in the (x,z) plan of the local frame. This contact scheme needs to be used for simulation of Solo 8 DoF in contacts.


edantec:
Not the good branch to merge...

PR #824, opened 2020-08-05
Benchmarks: remove redundant code

proyan:
-> Removes redundant code from the different benchmarks
-> changes multithreading and codegen macros to make them crocoddyl specific.
-> changes number of multithreading threads to be ncpu_threads/2, which gives best perf
-> outputs the benchmarks as csv files. provides python script to read these benchmarks.


CMakeLists.txt
```diff
@@ -69,23 +69,31 @@ OPTION(BUILD_WITH_CODEGEN_SUPPORT "Build the library with the Code Generation su
...
     MESSAGE(WARNING "the number of threads have to be an interger value, set to ${BUILD_WITH_NTHREADS}")
   ENDIF()
 ENDIF()
 
 IF(BUILD_WITH_CODEGEN_SUPPORT)
-  SET(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/find-external/CppAD/" ${CMAKE_MODULE_PATH})
-  ADD_PROJECT_DEPENDENCY(cppad)
-  ADD_PROJECT_DEPENDENCY(cppadcg)
+  ADD_PROJECT_DEPENDENCY(cppad 20200000.0 REQUIRED PKG_CONFIG_REQUIRES "cppad >= 20200000.0")
+  ADD_PROJECT_DEPENDENCY(cppadcg 2.4.1 REQUIRED PKG_CONFIG_REQUIRES "cppadcg >= 2.4.1")
```
proyan, line 90: @nim65s Could you please verify that this is the proper way to add these dependencies?

nim65s: I updated this in #806 (9a378d448c77ad7283cd012fc11016853061f21b) :
- we can't get the version number for now in cppad / cppadcg in our [custom finders](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/find-external/CppAD/Findcppad.cmake). So there is no use in defining them. If you really need to check their version number, and at the CMake level, we have to implement that in the submodule, so this calls for an issue or a PR there.
- I didn't put REQUIRED, which was a mistake of mine. So I guess you should rebase this PR on devel, and add this REQUIRED.
- PKG_CONFIG_REQUIRES would be useful only if someone was using crocoddyl.pc before, and was counting on finding a cppad requirement there. But I really hope that all projects dependent on crocoddyl are using its CMake exports right now, so there is no need for this transitional workaround.

proyan: So if I understand correctly:

1) PKG_CONFIG_REQUIRES would continue to support for pkg-config
2) if we leave version numbers here, even if they are not useful, they are not harmful. So a future update of cmake module will be all that is needed to add version checking?

I would like to check the cppad and cppadcg versions, since there is no back-compatibility in cppadcg. 2.4.1 is compatible with cppad 2020, and not the versions before. And there are some API changes between cppad 2018 and cppad 2020

nim65s: 1. PKG_CONFIG_REQUIRES idea is to deprecate support for pkg-config instead of brutally dropping it. But as we moved all our projects to pure CMake, nothing is tested with pkg-config anymore, so we'll have to drop its support at some point anyway. The size of the workarounds in the jrl-cmakemodule makes it really hard to maintain.
2. Is the compatibility between cppadcg 2.4.1 and cppad 2020 really an issue of yours, and not of cppadcg, or maybe package providers ?

proyan: Very well, let's drop the pkg-config support then 
> Is the compatibility between cppadcg 2.4.1 and cppad 2020 really an issue of yours, and not of cppadcg, or maybe package providers ?

The problem is from cppadcg. Before cppadcg 2.4.1, there was no minimum version requirement in cppadcg for cppad. So, I did have to struggle a bit (pinocchio still asks for cppad>=2018). 
The minimal way to make sure of compatibility would be to only check for cppadcg>=2.4.1.

benchmark/all-robots.cpp
```diff
@@ -153,35 +197,8 @@ void print_benchmark(RobotEENames robot) {
...
-      }
-      duration[i] = timer.get_us_duration();
-    }
-    avg[ithread] = AVG(duration);
-    stddev[ithread] = STDDEV(duration);
-    std::cout << ithread + 1 << " threaded diff calc [us]: \t" << avg[ithread] << " +- " << stddev[ithread]
-              << " (max: " << duration.maxCoeff() << ", min: " << duration.minCoeff()
-              << ", per nodes: " << avg[ithread] * (ithread + 1) / N << " +- " << stddev[ithread] * (ithread + 1) / N
-              << ")" << std::endl;
+    csv << "calc" << (ithread + 1) << false << avg[ithread] << stddev[ithread] << duration.maxCoeff()
```
cmastalli, line 200: Is this printing an output in the terminal as well?
It would be good if it does so.

include/crocoddyl/core/utils/file-io.hpp
```diff
@@ -0,0 +1,85 @@
...
+    file.endrow();
+    return file;
+  }
+
+  void endrow() {
+    fs_ << std::endl;
+    is_first_ = true;
+  }
+
+  CsvStream& operator<<(CsvStream& (*val)(CsvStream&)) { return val(*this); }
```
cmastalli, line 49: It seems to me this file doesn't print an output in the terminal.
I believe it would be good to do it as well.

I know you have prepared a Python script for nice processing, this is good. However, I still think that having an extra output in the terminal is useful.

CMakeLists.txt
proyan: I think this PR should solve this issue: https://github.com/jrl-umi3218/jrl-cmakemodules/pull/428

benchmark/all-robots.cpp
proyan: We already have a readable terminal output. Polluting that with an un-readable and duplicate stream is not a good idea

include/crocoddyl/core/utils/file-io.hpp
proyan: See my comment above.

benchmark/all-robots.cpp
cmastalli: I see it now. Thanks for the explanation

cmastalli:
@proyan is this ready to merge?

CMakeLists.txt
proyan: Should I wait for https://github.com/jrl-umi3218/jrl-cmakemodules/pull/428 to be handled before merging this?

nim65s: Yes, I'll review that in the day.

benchmark/read_csv.py
```diff
@@ -0,0 +1,70 @@
...
+
+benches = Benchmarks()
+
+filenames = sys.argv[1:]
+for filename in filenames:
+    if not exists(filename):
+        continue
+    benchname = splitext(basename(filename))[0]
+    seq = parseCsvFile(filename)
+    benches.__dict__[benchname] = seq
```
nim65s, line 67: usually, we try to avoid direct calls of dunder,. In this case, here is an example:
```python
settattr(benches, benchname, seq)
```

unittest/python/crocoddyl/fddp.py
```diff
@@ -5,8 +5,8 @@
 from .utils import raiseIfNan
 
 
-def rev_enumerate(l):
-    return reversed(list(enumerate(l)))
+def rev_enumerate(lname):
+    return reversed(list(enumerate(lname)))
```
nim65s, line 9: this `rev_enumerate` is duplicated. Maybe it could go in `.utils`.

benchmark/read_csv.py
proyan: cool, thanks!

PR #830, opened 2020-08-11
CoP cost for impulse dynamics

julesser:
This PR introduces a cost for center of pressure impulses. It includes: 
- The CoP impulse cost code (details on the mathematical formulation are provided with #792)
- Python bindings
- A unittest to check the validity of the code


bindings/python/crocoddyl/multibody/costs/impulse-cop-position.cpp
```diff
@@ -0,0 +1,74 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2020, University of Duisburg-Essen, University of Edinburgh
```
jcarpent, line 4: I don't think that `University of Edinburgh` should appear here, except if @cmastalli has contributed to it.

cmastalli: We have been discussing this work, and furthermore the code is based on other previous code.

jcarpent: It may might be then better to add `adapted from`.

PR #831, opened 2020-08-11
quasiStatic with multithreading support

cmastalli:
This PR proposes a multithreading support for computing the quasi-static commands along a state trajectory. This function helps us to compute faster a warm-start.

To do so, we need to define data for the `quasiStatic` functions defined in free / contact forward dynamics.
Note that the `ShootingProblem::quasiStatic` is bound in Python.

### DIFF
--- a/include/crocoddyl/core/action-base.hxx
+++ b/include/crocoddyl/core/action-base.hxx
@@ -48,6 +48,12 @@ void ActionModelAbstractTpl<Scalar>::quasiStatic(const boost::shared_ptr<ActionD
     throw_pretty("Invalid argument: "
                  << "x has wrong dimension (it should be " + std::to_string(state_->get_nx()) + ")");
   }
+#ifndef NDEBUG
+  const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> v = x.tail(state_->get_nv());
+#endif
+
+  // Check the velocity input is zero
+  assert_pretty(v.isZero(), "The velocity input should be zero for quasi-static to work.");
 
   const std::size_t& ndx = state_->get_ndx();
   VectorXs dx = VectorXs::Zero(ndx);
--- a/include/crocoddyl/core/action-base.hxx
+++ b/include/crocoddyl/core/action-base.hxx
@@ -48,12 +48,8 @@ void ActionModelAbstractTpl<Scalar>::quasiStatic(const boost::shared_ptr<ActionD
     throw_pretty("Invalid argument: "
                  << "x has wrong dimension (it should be " + std::to_string(state_->get_nx()) + ")");
   }
-#ifndef NDEBUG
-  const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> v = x.tail(state_->get_nv());
-#endif
-
   // Check the velocity input is zero
-  assert_pretty(v.isZero(), "The velocity input should be zero for quasi-static to work.");
+  assert_pretty(x.tail(state_->get_nv()).isZero(), "The velocity input should be zero for quasi-static to work.");
 
   const std::size_t& ndx = state_->get_ndx();
   VectorXs dx = VectorXs::Zero(ndx);
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -253,6 +253,9 @@ std::vector<typename MathBaseTpl<Scalar>::VectorXs> ShootingProblemTpl<Scalar>::
     const std::vector<VectorXs>& xs) {
   std::vector<VectorXs> us;
   us.resize(T_);
+  for (std::size_t i = 0; i < T_; ++i) {
+    us[i] = VectorXs::Zero(running_models_[i]->get_nu());
+  }
   quasiStatic(us, xs);
   return us;
 }
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -230,7 +230,7 @@ std::vector<typename MathBaseTpl<Scalar>::VectorXs> ShootingProblemTpl<Scalar>::
 }
 
 template <typename Scalar>
-void ShootingProblemTpl<Scalar>::quasiStatic(const std::vector<VectorXs>& us, std::vector<VectorXs>& xs) {
+void ShootingProblemTpl<Scalar>::quasiStatic(std::vector<VectorXs>& us, const std::vector<VectorXs>& xs) {
   if (xs.size() != T_) {
     throw_pretty("Invalid argument: "
                  << "xs has wrong dimension (it should be " + std::to_string(T_) + ")");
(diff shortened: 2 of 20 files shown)


include/crocoddyl/core/optctrl/shooting.hxx
```diff
@@ -231,6 +229,34 @@ std::vector<typename MathBaseTpl<Scalar>::VectorXs> ShootingProblemTpl<Scalar>::
...
+  if (us.size() != T_) {
+    throw_pretty("Invalid argument: "
+                 << "us has wrong dimension (it should be " + std::to_string(T_) + ")");
+  }
+
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#pragma omp parallel for
+#endif
+  for (std::size_t i = 0; i < T_; ++i) {
+    running_models_[i]->quasiStatic(running_datas_[i], xs[i], us[i]);
```
proyan, line 247: The API for quasiStatic is defined as quasiStatic(data, u, x)

include/crocoddyl/core/optctrl/shooting.hxx
```diff
@@ -231,6 +229,34 @@ std::vector<typename MathBaseTpl<Scalar>::VectorXs> ShootingProblemTpl<Scalar>::
   return xs;
 }
 
+template <typename Scalar>
+void ShootingProblemTpl<Scalar>::quasiStatic(const std::vector<VectorXs>& us, std::vector<VectorXs>& xs) {
```
proyan, line 233: `us` should not be constant, and `xs` should not be constant.

include/crocoddyl/multibody/actions/contact-fwddyn.hpp
```diff
@@ -119,6 +123,8 @@ struct DifferentialActionDataContactFwdDynamicsTpl : public DifferentialActionDa
   MatrixXs Kinv;
   MatrixXs df_dx;
   MatrixXs df_du;
+  VectorXs x_static;
+  MatrixXs J_static;
```
proyan, line 127: These are temporary variables that are not useful in the API. It would be better not to give them proper names. Rename them to `tmp_XXX` please.

include/crocoddyl/core/optctrl/shooting.hxx
cmastalli: Good catch! I have solved this issue and added unittest of all main shooting problem functions

include/crocoddyl/core/optctrl/shooting.hxx
cmastalli: done

include/crocoddyl/multibody/actions/contact-fwddyn.hpp
cmastalli: Ok, it sounds good to follow this approach. I have modified as suggested, note that there are plenty of places that don't follow it.

cmastalli:
@proyan I have solved all the issues that you pointed out plus added unittests.

You probably note that we are reaching the time limits in Travis job for debug mode. I am planning to remove those jobs in Travis since we have active the gepgitlab one, are you OK with that?

include/crocoddyl/core/action-base.hxx
```diff
@@ -48,6 +48,12 @@ void ActionModelAbstractTpl<Scalar>::quasiStatic(const boost::shared_ptr<ActionD
     throw_pretty("Invalid argument: "
                  << "x has wrong dimension (it should be " + std::to_string(state_->get_nx()) + ")");
   }
+#ifndef NDEBUG
```
proyan, line 51: You could remove this `ifndef` and simply do assert of x.tail(state_->get_nv()).isZero(). That would be simpler.

proyan:
Are the jobs on gepgitlab buildfarm the same as the travis ones?

nim65s:
They are not:
- Travis downloads and install dependencies from robotpkg for each and everytime, while gitlab uses a premade docker image with everything already installed
- Travis has 16.04/18.04 x release/devel, plus a clang-format, while gitlab has a main job 18.04 & python 3 & release, plus one where we change the 18.04 into other available distro (16.04 & debian 9 at the moment, but 20.04 & debian 10 are also ready), plus one where we change the python 3 into python 2, plus one where we change the release into devel (and use that one to generate documentation & coverage), plus one for clang-format & flake8 & yapf

cmastalli:
> Are the jobs on gepgitlab buildfarm the same as the travis ones?

In gitlab-ci we have Bionic Debug which is call doc-coverage. However, we don't have Xenial Debug.
Note that I didn't turn off these jobs, only allowed to fail. The problem is they are reaching often the maximum amount of time that Travis imposed. This happened with the other PR, and it occurs often when there are many jobs running.

I already discussed with @nim65s the possibility to add Xenial Debug in gitlab-ci.

proyan:
Is it possible to extend the time limit for certain builds instead? I know that we've done this before

cmastalli:
> Is it possible to extend the time limit for certain builds instead? I know that we've done this before

No, it is not possible. I already asked Travis guys to extend it, by default is 60 min and now we have 90 min.

cmastalli:
This will be much worse when we started to test code-generation stuff. I hope @nim65s give us more support with gitlab-ci. We could keep some of the fast-to-compile jobs in Travis so we don't overload the gitlab-ci server.

cmastalli:
@proyan note that we don't test examples in Debug mode due to this limitation. It would be great to start to test them in gitlab-ci as well.

proyan:
How about dividing the jobs on Travis then? Currently we do benchmarks, examples, unittests all in one single job. Perhaps it would be better to have _2_ runs, one which compiles and tests examples and unittests, and one which compiles benchmarks and python-binding based unittests. Is that a valid way out?

nim65s:
I guess you could make a pipeline on travis: one job that compiles the main stuff, and once this one is done, you could use its output to start other jobs (one for unittest, one for python bindings, one for examples…).

nim65s:
https://docs.travis-ci.com/user/build-stages/#data-persistence-between-stages-and-jobs oh, wait, you can't, this sucks.

cmastalli:
> How about dividing the jobs on Travis then? Currently we do benchmarks, examples, unittests all in one single job. Perhaps it would be better to have _2_ runs, one which compiles and tests examples and unittests, and one which compiles benchmarks and python-binding based unittests. Is that a valid way out? And is it possible to do?

We are not running the benchmarks in any of the jobs neither. So, what we has is
  - for Release jobs: unittest (c++ / python) and examples
  - for Debug jobs: unittest (c++ / python)

However, I agree with you, we could have a better set of tests, however I am not sure what it is the best way to optimize it. For instance, we could do something like this:
  - Job 1: c++ unittest
  - Job 2: python unittest
  - Job 3: examples
  - Job 4: c++ benchmarks
  - Job 5: python benchmarks
  - (in future) Job 6: c++ codegen

Each job could be run in both distros (Bionic and Xenial) and in both modes (Debug and Release). Additionally the python jobs could be run in Python 2 and 3.

From Travis side, I would be incline to define the Debug cases as "allowed to fail". This is due to the fact that most of the time is needed to compile library.

What it doesn't convince me of this path is the need of many jobs in Travis. I suspect this could be contra-productive, @nim65s and @wxmerkt could you give us your feedback?

cmastalli:
Guys, I have opened an issue regarding the CI. I believe this topic requires significant work / discussion for another PR.
I am very keen to implement (in Travis side) any solution that we have agreed.

Issue #832, opened 2020-08-13
Improving CI in both: Travis and gitlab-ci


cmastalli:
> How about dividing the jobs on Travis then? Currently we do benchmarks, examples, unittests all in one single job. Perhaps it would be better to have _2_ runs, one which compiles and tests examples and unittests, and one which compiles benchmarks and python-binding based unittests. Is that a valid way out? And is it possible to do?

We are not running the benchmarks in any of the jobs neither. So, what we has is
  - for Release jobs: unittest (c++ / python) and examples
  - for Debug jobs: unittest (c++ / python)

However, I agree with you, we could have a better set of tests, however I am not sure what it is the best way to optimize it. For instance, we could do something like this:
  - Job 1: c++ unittest
  - Job 2: python unittest
  - Job 3: examples
  - Job 4: c++ benchmarks
  - Job 5: python benchmarks
  - (in future) Job 6: c++ codegen

Each job could be run in both distros (Bionic and Xenial) and in both modes (Debug and Release). Additionally the python jobs could be run in Python 2 and 3.

From Travis side, I would be incline to define the Debug cases as "allowed to fail". This is due to the fact that most of the time is needed to compile library.

What it doesn't convince me of this path is the need of many jobs in Travis. I suspect this could be contra-productive, @nim65s and @wxmerkt could you give us your feedback?

_Originally posted by @cmastalli in https://github.com/loco-3d/crocoddyl/pull/831#issuecomment-673405557_

cmastalli:
@nim65s @proyan and @wxmerkt 
I am very keen to improve this. Could we discuss it here?

Otherwise, this discussion could be lost.

wxmerkt:
I believe Guilhem has the most experience and is the best to comment on this. 

I don't know if our builds are amenable to artefact re-use or caching or how splitting jobs would work. I just opened #834 to see if we see any gains in compilation time using ccache (for repeated builds within each distro + build type). Alternatively, we could try having one job build the artefacts, then cache, and another retrieve the cache and run the tests.

> From Travis side, I would be incline to define the Debug cases as "allowed to fail"

Given there is `assert` used in quite a few places (e.g. in the recent #831 for checking user input), I would keep at least one Debug job as required to pass.

cmastalli:
> Given there is `assert` used in quite a few places (e.g. in the recent #831 for checking user input), I would keep at least one Debug job as required to pass.

[...]

PR #834, opened 2020-08-14
[WIP] Travis: Enable ccache

wxmerkt:
To explore options to speed-up build or re-use caches (cf. [travis docs](https://docs.travis-ci.com/user/caching/)).

Relates to #832 

Other options:
- Use a job to build the artefacts & cache, then use another job to recall the artefacts and run the tests/examples/benchmarks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: cpp
 os: [
   "linux"
 ]
+cache: ccache
 
 jobs:
   fast_finish: true


cmastalli:
@wxmerkt do you think it's working? it seems the building time is the same that before.

wxmerkt:
It's not, sadly. The reason is that we do not have an overlap between the build type / distro keys. I'll close this PR.

cmastalli:
Thanks you for trying it :+1:

PR #839, opened 2020-08-21
Rk4 integrator

PepMS:
This PR adds an RK4 integrator that has been coded following the structure in https://github.com/loco-3d/crocoddyl/blob/4708428676f596f93ffe2df739faeb5cc38d2326/unittest/python/crocoddyl/integrated_action.py#L89.
I've found that in the reference python code there was a `*dt` missing in the cost computation (this is probably the only difference besides the use of Jtransport method).

It would be nice that you put special attention in the following items since I've had some doubts when implementing them:
- Inside `calc()` and `calcDiff()` the code that has to be inside the `enable_integration_` if clause.
- Inside `calc()` the code that has to be inside the `with_cost_residual_` if clause.
- `checkData` method
- `quasiStatic`method
- Also take a look on the data I've declared to store all auxiliary computations for the method: both the naming as well as the kind of container I've used.

I've checked several examples using this integrator and comparing it with the Euler solution and the results are close. Also, I've compared the c++ code with the python integrator (adding the change that I mentioned above) and both are the same.

I'm also adding the python bindings. Let me know if you want me to implement some tests for this integrator.


proyan:
Thanks @PepMS for this work!
We currently lack the infrastructure to test integrators. Meanwhile, once you update the bindings, I think it would be nice if you can

1) Modify the unittest/python/test_rk4.py to reflect the new c++ version. 
2) Add your test comparing the current implementation against the legacy implementation that I did.

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2018-2020, LAAS-CNRS, IRI: CSIC-UPC
+// Copyright note valid unless otherwise stated in individual files.
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#include <iostream>
```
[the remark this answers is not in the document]
PepMS: Cannot do this because of https://github.com/loco-3d/crocoddyl/blob/bc20d4dfc8b2efd4df61a488a15b6aea57bf861d/include/crocoddyl/core/integrator/rk4.hxx#L28

PepMS:
I've just  pushed all the requested changes.

Regarding the test against legacy code. I've done as we discussed. That is, I've changed the legacy python code inside unittest to `crocoddyl_legacy`. Also, I've done some minor changes to the legacy RK4 integrator to make it compaitble with the current bindings. Besides, I've added a python script named `test_rk4_comp.py` that containts the comparison for the quadrotor example and the arm manipulation example.

I think that this is all. If any further change is needed, let me know please.

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+
+  const std::size_t& nv = differential_->get_state()->get_nv();
+
+  // Static casting the data
+  boost::shared_ptr<Data> d = boost::static_pointer_cast<Data>(data);
+
+  // Computing the acceleration and cost
+  d->y[0] = x;
+  differential_->calc(d->differential[0], d->y[0], u);
+  d->ki[0] << d->y[0].tail(nv), d->differential[0]->xout;
```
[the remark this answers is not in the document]
cmastalli: I agree with @proyan

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  const std::size_t& nv = differential_->get_state()->get_nv();
+
+  boost::shared_ptr<Data> d = boost::static_pointer_cast<Data>(data);
+
+  differential_->calcDiff(d->differential[0], d->y[0], u);
+
+  d->dki_dy[0].topRightCorner(nv, nv).diagonal().array() = (Scalar)1;
+  d->dki_dy[0].bottomRows(nv).noalias() = d->differential[0]->Fx;
```
cmastalli: No need for `noalias()`, please remove it. The `noalias()` is using to avoid Eigen to allocate memory for a third matrix that comes from the results of multiplying two matrices.

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+  d->dyi_dx[0].diagonal().array() = (Scalar)1;
+  d->dki_dx[0] = d->dki_dy[0];
+  d->dki_du[0].bottomRows(nv).noalias() = d->differential[0]->Fu;
```
cmastalli: same here

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+    differential_->get_state()->Jintegrate(x, d->dx_rk4[i], d->dyi_dx[i], d->dyi_dx[i], first, addto);
+    d->dki_dx[i].noalias() = d->dki_dy[i] * d->dyi_dx[i];
+
+    d->dyi_du[i].noalias() = d->dki_du[i - 1] * rk4_c_[i] * time_step_;
+    differential_->get_state()->JintegrateTransport(x, d->dx_rk4[i], d->dyi_du[i], second);
+    d->dki_du[i].noalias() = d->dki_dy[i] * d->dyi_du[i];
+    d->dki_du[i].bottomRows(nv) += d->differential[i]->Fu;
+
+    d->dli_dx[i].noalias() = d->differential[i]->Lx.transpose() * d->dyi_dx[i];
+    d->dli_du[i].noalias() = d->differential[i]->Lu.transpose() + d->differential[i]->Lx.transpose() * d->dyi_du[i];
```
cmastalli: You need to split this operation to avoid memory allocation in matrix product, i.e.
```cpp
d->dli_du[i].noalias() = d->differential[i]->Lu.transpose();
d->dli_du[i].noalias().noalias() += d->differential[i]->Lx.transpose() * d->dyi_du[i];
```

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+    d->ddli_ddx[i] = d->dyi_dx[i].transpose() * d->differential[i]->Lxx * d->dyi_dx[i];
```
cmastalli: Three matrix product allocated dynamically memory. The workaround is the follows:
  1. Create a temporary data in structure to store the value of `Lxx * dyi_dx`
  2. Compute 1 using `noalias()`
  3. Modify this line to use the result of 2 and including `noalias()`

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+    d->Luu_partialx[i] = d->differential[i]->Lxu.transpose() * d->dyi_du[i];
```
cmastalli: Use `noalias()`

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+    d->ddli_ddu[i] = d->differential[i]->Luu + d->Luu_partialx[i].transpose() + d->Luu_partialx[i] +
```
cmastalli: Split this operation as described above

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+                     d->dyi_du[i].transpose() * d->differential[i]->Lxx * d->dyi_du[i];
+    d->ddli_dxdu[i] = d->dyi_dx[i].transpose() * d->differential[i]->Lxu +
```
cmastalli: same here, just do not forget to add `noalias()` for the matrices product.

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+                      d->dyi_dx[i].transpose() * d->differential[i]->Lxx * d->dyi_du[i];
+  }
+
+  if (enable_integration_) {
+    d->Fx.noalias() = time_step_ / 6 * (d->dki_dx[0] + 2 * d->dki_dx[1] + 2 * d->dki_dx[2] + d->dki_dx[3]);
```
cmastalli: Scalar typecasting in `6` and `2` values

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+    differential_->get_state()->JintegrateTransport(x, d->dx, d->Fx, second);
+    differential_->get_state()->Jintegrate(x, d->dx, d->Fx, d->Fx, first, addto);
+
+    d->Fu.noalias() = time_step_ / 6 * (d->dki_du[0] + 2 * d->dki_du[1] + 2 * d->dki_du[2] + d->dki_du[3]);
+    differential_->get_state()->JintegrateTransport(x, d->dx, d->Fu, second);
+
+    d->Lx.noalias() = time_step_ / 6 * (d->dli_dx[0] + 2 * d->dli_dx[1] + 2 * d->dli_dx[2] + d->dli_dx[3]);
```
cmastalli: Same for the other lines, there are a few of them as you can see.

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+    d->Lu.noalias() = time_step_ / 6 * (d->dli_du[0] + 2 * d->dli_du[1] + 2 * d->dli_du[2] + d->dli_du[3]);
+
+    d->Lxx.noalias() = time_step_ / 6 * (d->ddli_ddx[0] + 2 * d->ddli_ddx[1] + 2 * d->ddli_ddx[2] + d->ddli_ddx[3]);
+    d->Luu.noalias() = time_step_ / 6 * (d->ddli_ddu[0] + 2 * d->ddli_ddu[1] + 2 * d->ddli_ddu[2] + d->ddli_ddu[3]);
+    d->Lxu.noalias() =
+        time_step_ / 6 * (d->ddli_dxdu[0] + 2 * d->ddli_dxdu[1] + 2 * d->ddli_dxdu[2] + d->ddli_dxdu[3]);
+  } else {
+    differential_->get_state()->Jintegrate(x, d->dx, d->Fx, d->Fx);
+    d->Fu.setZero();
+    d->Lx = d->differential[0]->Lx;
```
cmastalli, line 174: Do not forget to add `dt` in the cost integration. To add more explanation, the legacy code was assuming that the cost is discrete time. But this is not good and consistent.

You could use here any integration scheme, and maybe it is interesting to add other options in future. For now just keep the trapezoidal quadrature as done in Euler integrator.

cmastalli (commented):
@PepMS this work and very clean.
There are a few changes to be done, a part of all the @proyan's requests.

bindings/python/crocoddyl/core/integrator/rk4.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace crocoddyl {
+namespace python {
+
+void exposeIntegratedActionRK4() {
+  bp::class_<IntegratedActionModelRK4, bp::bases<ActionModelAbstract> >(
+      "IntegratedActionModelRK4",
+      "RK4 integrator for differential action models.\n\n"
+      "This class implements an RK4 integrator (a.k.a semi-implicit\n"
+      "integrator) give a differential action model, i.e.:\n"
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hpp
```diff
@@ -1,14 +1,152 @@
...
+                           const Scalar& tol = Scalar(1e-9));
+
+  const boost::shared_ptr<DifferentialActionModelAbstract>& get_differential() const;
+  const Scalar& get_dt() const;
+
+  void set_dt(const Scalar& dt);
+  void set_differential(boost::shared_ptr<DifferentialActionModelAbstract> model);
+
+ protected:
+  using Base::has_control_limits_;  //!< Indicates whether any of the control limits
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hpp
```diff
@@ -1,14 +1,152 @@
...
+    const std::size_t& nx = model->get_state()->get_nx();
+
+    const std::size_t& nu = model->get_nu();
+    for (std::size_t i = 0; i < 4; ++i) {
+      differential.push_back(
+          boost::shared_ptr<DifferentialActionDataAbstractTpl<Scalar>>(model->get_differential()->createData()));
+    }
+
+    dx = VectorXs::Zero(ndx);
+    integral = {0, 0, 0, 0};
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+  // Computing the acceleration and cost
+  d->y[0] = x;
+  differential_->calc(d->differential[0], d->y[0], u);
+  d->ki[0] << d->y[0].tail(nv), d->differential[0]->xout;
+  d->integral[0] = d->differential[0]->cost;
+  for (std::size_t i = 1; i < 4; ++i) {
+    d->dx_rk4[i].noalias() = time_step_ * rk4_c_[i] * d->ki[i - 1];
+    differential_->get_state()->integrate(x, d->dx_rk4[i], d->y[i]);
+    differential_->calc(d->differential[i], d->y[i], u);
+    d->ki[i] << d->y[i].tail(nv), d->differential[i]->xout;
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+    d->integral[i] = d->differential[i]->cost;
+  }
+
+  // Computing the next state (discrete time)
+  if (enable_integration_) {
+    d->dx = (d->ki[0] + 2 * d->ki[1] + 2 * d->ki[2] + d->ki[3]) * time_step_ / 6;
+    differential_->get_state()->integrate(x, d->dx, d->xnext);
+    d->cost = (d->integral[0] + 2 * d->integral[1] + 2 * d->integral[2] + d->integral[3]) * time_step_ / 6;
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+                 << "u has wrong dimension (it should be " + std::to_string(nu_) + ")");
+  }
+
+  const std::size_t& nv = differential_->get_state()->get_nv();
+
+  boost::shared_ptr<Data> d = boost::static_pointer_cast<Data>(data);
+
+  differential_->calcDiff(d->differential[0], d->y[0], u);
+
+  d->dki_dy[0].topRightCorner(nv, nv).diagonal().array() = (Scalar)1;
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+                 << "u has wrong dimension (it should be " + std::to_string(nu_) + ")");
+  }
+
+  const std::size_t& nv = differential_->get_state()->get_nv();
+
+  // Static casting the data
+  boost::shared_ptr<Data> d = boost::static_pointer_cast<Data>(data);
+
+  // Computing the acceleration and cost
+  d->y[0] = x;
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: I'm not sure I'm getting how the dt has to be added here. Can you point me at the equivalent line of the Euler integrator please?

bindings/python/crocoddyl/core/integrator/rk4.cpp
```diff
@@ -0,0 +1,85 @@
...
+      "  [q+, v+] = State.integrate([q, v], dt / 6 (k0 + k1 + k2 + k3)) with \n"
+      "k0 = f(x, u) \n"
+      "k1 = f(x + dt / 2 * k1, u) \n"
+      "k2 = f(x + dt / 2 * k2, u) \n"
+      "k3 = f(x + dt * k2, u) \n",
+      bp::init<boost::shared_ptr<DifferentialActionModelAbstract>, bp::optional<double, bool> >(
+          bp::args("self", "diffModel", "stepTime", "withCostResidual"),
+          "Initialize the RK4 integrator.\n\n"
+          ":param diffModel: differential action model\n"
+          ":param stepTime: step time\n"
```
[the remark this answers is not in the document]
PepMS: done

bindings/python/crocoddyl/core/integrator/rk4.cpp
```diff
@@ -0,0 +1,85 @@
...
+  bp::class_<IntegratedActionDataRK4, bp::bases<ActionDataAbstract> >(
+      "IntegratedActionDataRK4", "RK4 integrator data.",
+      bp::init<IntegratedActionModelRK4*>(bp::args("self", "model"),
+                                          "Create RK4 integrator data.\n\n"
+                                          ":param model: RK4 integrator model"))
+      .add_property(
+          "differential",
+          bp::make_getter(&IntegratedActionDataRK4::differential, bp::return_value_policy<bp::return_by_value>()),
+          "differential action data")
+      .add_property("dx", bp::make_getter(&IntegratedActionDataRK4::dx, bp::return_internal_reference<>()),
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

bindings/python/crocoddyl/core/integrator/euler.cpp
```diff
@@ -42,9 +42,9 @@ void exposeIntegratedActionEuler() {
                                                 const Eigen::Ref<const Eigen::VectorXd>&,
                                                 const Eigen::Ref<const Eigen::VectorXd>&)>(
           "calcDiff", &IntegratedActionModelEuler::calcDiff, bp::args("self", "data", "x", "u"),
-          "Compute the time-discrete derivatives of a differential action model.\n\n"
-          "It computes the time-discrete partial derivatives of a differential\n"
-          "action model. It assumes that calc has been run first.\n"
+          "Compute the time-discrete derivatives of the integrated action model.\n\n"
```
cmastalli, line 45: This comment is wrong, it was alright as before.
The integrator convert continuous to discrete, the differential action model is continuous thing.

The integrated action takes a differential and then produces an action model.
When we say compute ... of a differential ..., it means that it takes the differential to produce an action model.

bindings/python/crocoddyl/core/integrator/rk4.cpp
```diff
@@ -0,0 +1,116 @@
...
+          ":param x: state vector\n"
+          ":param u: control input")
+      .def<void (IntegratedActionModelRK4::*)(const boost::shared_ptr<ActionDataAbstract>&,
+                                              const Eigen::Ref<const Eigen::VectorXd>&)>(
+          "calc", &ActionModelAbstract::calc, bp::args("self", "data", "x"))
+      .def<void (IntegratedActionModelRK4::*)(const boost::shared_ptr<ActionDataAbstract>&,
+                                              const Eigen::Ref<const Eigen::VectorXd>&,
+                                              const Eigen::Ref<const Eigen::VectorXd>&)>(
+          "calcDiff", &IntegratedActionModelRK4::calcDiff, bp::args("self", "data", "x", "u"),
+          "Compute the time-discrete derivatives of the integrated action model.\n\n"
```
cmastalli, line 49: Same problem here

include/crocoddyl/core/integrator/rk4.hpp
```diff
@@ -1,14 +1,156 @@
...
+    dli_dx = std::vector<VectorXs>(4, VectorXs::Zero(ndx));
+    dli_du = std::vector<VectorXs>(4, VectorXs::Zero(nu));
+    ddli_ddx = std::vector<MatrixXs>(4, MatrixXs::Zero(ndx, ndx));
+    ddli_ddu = std::vector<MatrixXs>(4, MatrixXs::Zero(nu, nu));
+    ddli_dxdu = std::vector<MatrixXs>(4, MatrixXs::Zero(ndx, nu));
+    Luu_partialx = std::vector<MatrixXs>(4, MatrixXs::Zero(nu, nu));
+    Lxx_partialx = std::vector<MatrixXs>(4, MatrixXs::Zero(ndx, ndx));
+    Lxx_partialu = std::vector<MatrixXs>(4, MatrixXs::Zero(ndx, nu));
+  }
+  ~IntegratedActionDataRK4Tpl() {}
```
cmastalli, line 113: Define the destructor as virtual
```cpp
virtual ~IntegratedActionDataRK4Tpl() {}
```

include/crocoddyl/core/integrator/rk4.hxx
cmastalli: https://github.com/loco-3d/crocoddyl/blob/master/include/crocoddyl/core/integrator/euler.hxx#L117-L121

unittest/python/crocoddyl_legacy/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ SET(${PROJECT_NAME}_PYTHON_FILES
   )
 
 FOREACH(python ${${PROJECT_NAME}_PYTHON_FILES})
-  PYTHON_BUILD(../crocoddyl ${python})
+  PYTHON_BUILD(../crocoddyl_legacy ${python})
```
cmastalli, line 29: This line is making possible to build our Python bindings, it doesn't have anything to do with legacy code.
Furthermore, we don't want to build the legacy code, they will be erase asap.

In short, you need to revert this change together with the first comment.

CMakeLists.txt
```diff
@@ -155,7 +155,7 @@ IF(BUILD_PYTHON_INTERFACE)
   # Include and add the bindings directory
   INCLUDE_DIRECTORIES(bindings)
   ADD_SUBDIRECTORY(bindings)
-  ADD_SUBDIRECTORY(unittest/python/${PROJECT_NAME})
+  ADD_SUBDIRECTORY(unittest/python/${PROJECT_NAME}_legacy)
```
cmastalli, line 158: This needs to be reverted since it doesn't include the Python unittest, which doesn't have anything to do with legacy unittest. Furthermore, there is no `crocodyl_legacy` folder and we don't want to build/unittest legacy code.

There is another revert to be done regarding this topic, see the comment below.

unittest/python/crocoddyl_legacy/integrated_action.py
```diff
@@ -1,10 +1,11 @@
 import numpy as np
+import crocoddyl
```
cmastalli, line 2: Do not create this folder (i.e. `crocoddyl_legacy`) and file here.
Put the Python integrator inside here: https://github.com/loco-3d/crocoddyl/blob/master/bindings/python/crocoddyl/utils/__init__.py#L390

Furthermore, I am not sure why we have an Euler integrator as well. This is included in the file that I have just shared.

cmastalli: For the moment, I am keeping all the Python classes used to check the c++ code. Note that this file is not installed, only built internally for the unittest purpose.

unittest/python/crocoddyl_legacy/integrated_action.py
```diff
@@ -1,10 +1,11 @@
...
 
 
 class IntegratedActionModelEuler:
```
cmastalli, line 5: You need to revert the change of folder of each file inside the `unittest/python/crocoddyl`.
You should not code anything there.

unittest/python/crocoddyl_legacy/robots.py
```diff
@@ -4,7 +4,7 @@
 import pinocchio
 from pinocchio.robot_wrapper import RobotWrapper
 
-pinocchio.switchToNumpyMatrix()
+# pinocchio.switchToNumpyMatrix()
```
cmastalli, line 7: Again revert all theses changes.
We don't want to modify them, they are static files to be erased.

unittest/python/test_actions.py
```diff
@@ -1,7 +1,7 @@
 import unittest
 
 import numpy as np
-from crocoddyl import ActionModelNumDiff, ActionModelUnicycle, ActionModelUnicycleVar
+from crocoddyl_legacy import ActionModelNumDiff, ActionModelUnicycle, ActionModelUnicycleVar
```
cmastalli, line 4: And same here and the below files

unittest/python/test_rk4_comp.py
```diff
@@ -0,0 +1,168 @@
+import numpy as np
+
+import pinocchio
+import example_robot_data
+
+import crocoddyl
+import crocoddyl_legacy
```
cmastalli, line 7: You will need to include the utils module, see this example: https://github.com/loco-3d/crocoddyl/blob/master/unittest/bindings/test_actions.py#L10

unittest/python/test_rk4_comp.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class SimpleQuadrotorProblem:
```
cmastalli, line 10: Simplify the unittest. Let me guide you:
  1. Create a base class for the unitest where you compare both problems (e.g. https://github.com/loco-3d/crocoddyl/blob/devel/unittest/bindings/test_actions.py#L14)
  2. Define the different problems (i.e. quadrotor and arm) in a derived class (e.g. https://github.com/loco-3d/crocoddyl/blob/devel/unittest/bindings/test_actions.py#L67)
 3. List the test (e.g. https://github.com/loco-3d/crocoddyl/blob/devel/unittest/bindings/test_actions.py#L132)

cmastalli (changes requested):
Hi @PepMS 

All the changes regarding the legacy code needs to be reverted. I think you got confused of how we do our Python test. I have written instructions about it, do not hesitate to write me in private if you are still confuse.

Furthermore, it would be important to clean up the unittest code for the integrator. I also have gave instructions (and examples) of how it should be done.

Finally, there are a couples of problem in the Python documentation of the Euler and RK4 integrator. I have written an explanation of how, if you think we could extend the documentation to be more clear to third parties.

bindings/python/crocoddyl/core/integrator/euler.cpp
proyan: > The integrator convert continuous to discrete, the differential action model is continuous thing.

This statement is misleading. The differential action model is continuous is true. But `x(t) = integral(t0,t )( f(x) )` is the continuous integration. If we had implemented this, calcDiff would still return the time derivatives. The integrator that we implement doesn't *convert* continuous to discrete. It is either continous, or discrete. 

The statement should be "Computes the derivatives of the integrated action model wrt state and control".

include/crocoddyl/core/integrator/rk4.hxx
PepMS: The integration is done in these lines: https://github.com/PepMS/crocoddyl/blob/238d01eae8c0595888231667681eb4c5bc426f7d/include/crocoddyl/core/integrator/rk4.hxx#L156-L176

The ones that you say are inside the `else`clause.

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+      time_step_(time_step),
+      with_cost_residual_(with_cost_residual),
+      enable_integration_(true) {
+  Base::set_u_lb(differential_->get_u_lb());
+  Base::set_u_ub(differential_->get_u_ub());
+  if (time_step_ < Scalar(0.)) {
+    time_step_ = Scalar(1e-3);
+    std::cerr << "Warning: dt should be positive, set to 1e-3" << std::endl;
+  }
+  if (time_step == 0.) {
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+    enable_integration_ = false;
+  }
+  rk4_c_ = {0.0, 0.5, 0.5, 1.0};
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,239 @@
...
+    d->dx_rk4[i].noalias() = time_step_ * rk4_c_[i] * d->ki[i - 1];
+    differential_->get_state()->integrate(x, d->dx_rk4[i], d->y[i]);
+    differential_->calc(d->differential[i], d->y[i], u);
+    d->ki[i] << d->y[i].tail(nv), d->differential[i]->xout;
+    d->integral[i] = d->differential[i]->cost;
+  }
+
+  // Computing the next state (discrete time)
+  if (enable_integration_) {
+    d->dx = (d->ki[0] + 2 * d->ki[1] + 2 * d->ki[2] + d->ki[3]) * time_step_ / 6;
```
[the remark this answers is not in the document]
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: The integration is done in these lines: https://github.com/PepMS/crocoddyl/blob/238d01eae8c0595888231667681eb4c5bc426f7d/include/crocoddyl/core/integrator/rk4.hxx#L156-L176

The ones that you say are inside the `else`clause.

bindings/python/crocoddyl/core/integrator/euler.cpp
PepMS: done

bindings/python/crocoddyl/core/integrator/rk4.cpp
PepMS: done

unittest/python/crocoddyl_legacy/CMakeLists.txt
PepMS: done

include/crocoddyl/core/integrator/rk4.hpp
PepMS: This is the data. In case this has to be virtual, should I change also the `IntegratedActionDataEulerTpl` ?

CMakeLists.txt
PepMS: done

unittest/python/crocoddyl_legacy/integrated_action.py
PepMS: done

unittest/python/crocoddyl_legacy/robots.py
PepMS: done

unittest/python/test_actions.py
PepMS: done

PepMS:
Ok, next try:
- I've reverted all the crocoddyl_legacy code to crocoddyl
- I've included the legacy RK4 integrator into the `bindings/utils/__init__.py`
- I've used the testing structure in `unittest/bindings/test_actions.py` to create two more tests related to the RK4 integrator.
- all test passed when `make test`

I hope this is the right direction.

unittest/python/test_rk4_comp.py
PepMS: done

unittest/python/test_actions.py
PepMS: done

unittest/python/crocoddyl_legacy/robots.py
PepMS: done

unittest/python/crocoddyl_legacy/integrated_action.py
PepMS: done

unittest/python/crocoddyl_legacy/integrated_action.py
PepMS: done

unittest/python/test_rk4_comp.py
PepMS: done

bindings/python/crocoddyl/utils/__init__.py
```diff
@@ -437,6 +437,167 @@ def __init__(self, model):
...
+        for i in range(4):
+            self.differential.calcDiff(data.differential[i], data.y[i], u)
+            data.dki_dy[i] = np.bmat([[np.zeros([nv, nv]), np.identity(nv)], [data.differential[i].Fx]])
+
+        data.dki_du[0] = np.vstack([np.zeros([nv, nu]), data.differential[0].Fu])
+
+        data.Lx[:] = data.differential[0].Lx
+        data.Lu[:] = data.differential[0].Lu
+
+        data.dy_dx[0] = np.identity(nv * 2)
```
cmastalli, line 492: This is tiny detail about the use of the Python bindings. If you do
```cpp
data.dy_dx[0] = np.identity(nv * 2)
```
what it is occurring is that Python is allocating a new object, instead of copying internally the data.

To solve this, you should write the internal data, e.g.
```cpp
data.dy_dx[0][:, :] = np.identity(nv * 2)
```
which it is important to run in parallel this code (but this feature is not yet available).
Feel free to make these changes, however, do not worry because it is not that important.

Btw, the previous operation is needed to be performed once, right? If that is correct then it makes sense to do it in the data creation. With this approach, you avoid to fill constant values. If you have this in the c++ code, then I will suggest to do it there to improve efficiency (not needed to do it in this Python class).

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  const std::size_t& nv = differential_->get_state()->get_nv();
+
+  boost::shared_ptr<Data> d = boost::static_pointer_cast<Data>(data);
+
+  differential_->calcDiff(d->differential[0], x, u);
+
+  if (enable_integration_) {
+    d->dki_dy[0].topRightCorner(nv, nv).diagonal().array() = (Scalar)1;
```
cmastalli, line 112: I will move this operation to the data creation, so we avoid to fill the same value each time

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,257 @@
...
+    d->dki_dy[0].bottomRows(nv) = d->differential[0]->Fx;
+    d->dyi_dx[0].diagonal().array() = (Scalar)1;
```
cmastalli, line 114: same here

include/crocoddyl/core/integrator/rk4.hxx
```diff
@@ -0,0 +1,257 @@
...
+    d->dli_dx[0] = d->differential[0]->Lx;
+    d->dli_du[0] = d->differential[0]->Lu;
+
+    d->ddli_ddx[0] = d->differential[0]->Lxx;
+    d->ddli_ddu[0] = d->differential[0]->Luu;
+    d->ddli_dxdu[0] = d->differential[0]->Lxu;
+
+    for (std::size_t i = 1; i < 4; ++i) {
+      differential_->calcDiff(d->differential[i], d->y[i], u);
+      d->dki_dy[i].topRightCorner(nv, nv).diagonal().array() = (Scalar)1;
```
cmastalli, line 127: same here

cmastalli (commented):
@PepMS the unittest code is very nice now, well done!

I left a small comment regarding the Python usage which you could ignore.

There are also a few comments that improve the efficiency of the RK4 calcDiff routine, I believe it is important to handle before the merge.

include/crocoddyl/core/integrator/rk4.hpp
cmastalli: Yes, please do it for the Euler integrator as well.

bindings/python/crocoddyl/utils/__init__.py
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

include/crocoddyl/core/integrator/rk4.hxx
PepMS: done

include/crocoddyl/core/integrator/rk4.hpp
PepMS: done. Which is the reason for doing the destructor virtual?

proyan (approved):

include/crocoddyl/core/integrator/rk4.hpp
proyan: So that if there are any classes that inherit from this class, and pointers are resolved dynamically, the destructor of the derived class is used and not of the base class.

proyan:
Thanks for the work @PepMS ! If @cmastalli is satisfied, we can merge this.

cmastalli (approved):
Everything good now :)
Thanks @PepMS for this work, thanks @proyan for the review

include/crocoddyl/core/integrator/rk4.hpp
cmastalli: In top of what Rohan said, if you don't call the derived destructor it will occurs memory leak. You might want to expect 5 minutes reading the following link: https://www.quantstart.com/articles/C-Virtual-Destructors-How-to-Avoid-Memory-Leaks/

PepMS:
Great! It's a pleasure! :)

PR #840, opened 2020-08-21
SolverAbstract: Throw exception if the candidate xs/us have the wrong dimensions

wxmerkt:
This caught me out again hard :cry:. I checked and I believe this was the last `assert_pretty` that could have unintended consequences.

assert_pretty is silent in Release mode. Passing in a wrong-sized xs/us
will be copied and messes up the entire memory resulting in optimisation
rubbish. Moving these checks to if / throw_pretty resolves this issue.


cmastalli (approved):
Thanks @wxmerkt for handling these tiny problems

PR #843, opened 2020-08-24
create kkt solver bindings.

proyan:


cmastalli (approved):
Thank you so much @proyan.

Even thought that I approve the PR as it is, it could be interesting to add one simple test inside `test_solvers.py`

cmastalli:
@proyan I have asked @nim65s to disable the Buster jobs, and it should be done. This PR works doesn't have anything to do with the failure on those jobs.

I will merge this PR, again it should produce any error in the merging jobs. If so, we need to ask @nim65s to check it.

cmastalli:
I will later try to solve this issue with our unittest system in Buster.

Issue #842, opened 2020-08-21
Boost.Python: Optional arguments in overloaded function break Python-derived models


wxmerkt:
Managed to reliably reproduce the issue I referred to in #841. Here is a MWE:

Note, that each of the `ddp.solve(...)` options below should work the same. However, the only one that calls the `calc` method in our Python-derived class (which prints x and u and raises an exception to test) is the one with no arguments. The other calls to `ddp.solve(xs, us, etc.)` never call back into the Python-derived class.

```python
#!/usr/bin/env python
# coding: utf-8
from __future__ import print_function, division

import numpy as np
import crocoddyl

class DifferentialActionModelTestIssue842(crocoddyl.DifferentialActionModelAbstract):
    def __init__(self):
        crocoddyl.DifferentialActionModelAbstract.__init__(self, crocoddyl.StateVector(4), 12, 2)

    def calc(self, data, x, u=None):
        print("[calc]", "x", x, "u", u)
        raise

    def calcDiff(self, data, x, u=None):
        pass


# Creating the DAM
def create_integrated_action_model():
    testDAM = DifferentialActionModelTestIssue842()
    testND = crocoddyl.DifferentialActionModelNumDiff(testDAM, True)
    testIAM = crocoddyl.IntegratedActionModelEuler(testND, 1e-2)
    return testIAM

# Creating the shooting problem
x0 = np.array([0., 0., 0., 0.])
T = 50

terminal_model = create_integrated_action_model()
running_models = [create_integrated_action_model() for _ in range(T)]

# Initial control sequence depends on the gait status:
us_init = [np.ones(running_models[t].nu) for t in range(T)]
xs_init = [np.array(x).copy() for x in np.linspace([0,0], [1,0], T + 1).tolist()]
  
problem = crocoddyl.ShootingProblem(x0, running_models, terminal_model)

# Solving it using DDP
ddp = crocoddyl.SolverFDDP(problem)
ddp.setCallbacks([crocoddyl.CallbackVerbose()])

# ddp.setCandidate(xs_init, us_init, False) # this is not working due to #841

ddp.solve(xs_init)                           # never calls DifferentialActionModelTestIssue842.calc
ddp.solve(xs_init, us_init)                  # never calls DifferentialActionModelTestIssue842.calc
ddp.solve(xs_init, us_init, 100)             # never calls DifferentialActionModelTestIssue842.calc
[...]

PR #844, opened 2020-08-24
Check that the dimension of warm-start is correct

cmastalli:
From discussion in #842, we can see that the user might struggle to debug an error in the warm-start dimensions.
This PR solves this issue, @wxmerkt you might be interested to review the code.


wxmerkt (approved):
I was just about to open a pull request with this, thank you! :-)

cmastalli:
I will merge this PR since the jobs that are failing are due to Buster issues.
Those jobs has been allowed to fail for the time being.

PR #845, opened 2020-09-02
Fixed broken python benchmarks 

cmastalli:


benchmark/lqr_optctrl.py
```diff
@@ -1,5 +1,5 @@
 import crocoddyl
-from crocoddyl.utils import LQRDerived
+from crocoddyl.utils import LQRModelDerived
```
jcarpent, line 2: @cmastalli Do you think that it is worth to use the suffix `Derived` in the common API?

cmastalli: This class is not part of the user API, its only purpose is to test python bindings.

cmastalli: I will merge this PR.

However, we could later rename those classes, if you still believe it is important.
Just be aware that they are only pure Python code used only for internal purposes (unittest, and some times benchmarking)

PR #846, opened 2020-09-07
[CMake] update Boost use

nim65s:


jcarpent (changes requested):
@nim65s You have forgotten to setup the Boost default options. See Pinocchio for such use case.

CMakeLists.txt
```diff
@@ -140,8 +138,8 @@ IF(UNIX)
   SET_TARGET_PROPERTIES(${PROJECT_NAME} PROPERTIES LINKER_LANGUAGE CXX)
   TARGET_INCLUDE_DIRECTORIES(${PROJECT_NAME} PUBLIC $<INSTALL_INTERFACE:include>)
   TARGET_LINK_LIBRARIES(${PROJECT_NAME} pinocchio::pinocchio)
-  TARGET_LINK_LIBRARIES(${PROJECT_NAME} ${Boost_FILESYSTEM_LIBRARY} ${Boost_SYSTEM_LIBRARY}
-      ${Boost_SERIALIZATION_LIBRARY})
+  TARGET_LINK_LIBRARIES(${PROJECT_NAME} Boost::filesystem Boost::system
```
jcarpent, line 141: might not work for all the OS.

nim65s: For now, I'm just checking if this fix an issue for @edantec, but thanks for the heads-up, I'll add that if this works for him.

nim65s: done

jcarpent: Again, this might not work for all the versions of Boost.

jcarpent (changes requested):

CMakeLists.txt
nim65s: I am not aware of any supported OS on which this wouldn't work. But if this can reassure you to make this a bit cleaner, I can find the minimal version of Boost on which this work, and set that as the minimal required Boost version.

jcarpent: I think OS X won't work for instance and if I where, I would not change it, because there is no real need.

nim65s: I found the issue on @edantec's computer, it was unrelated to CMake. I'll close this.

jcarpent: Thanks for letting us know.

jcarpent:
You should keep at least the new way of using Boost.

PR #848, opened 2020-09-08
[solver/ddp] if gaps are zero, solution is feasible

proyan:
This comes from a problem that is documented here: https://gitlab.laas.fr/aparag/warmStart/-/blob/master/unicycle/Feasible%20warmstart.ipynb

If the solver is being started with isFeasible=False. This boolean is only updated to true when you make a full step. And if for some reason, this doesn't happen (which normally should not occur), this boolean remains False. We don't set this boolean on the basis of the gaps. we set it on the basis of the user input.

### DIFF
--- a/src/core/solvers/ddp.cpp
+++ b/src/core/solvers/ddp.cpp
@@ -19,6 +19,7 @@ SolverDDP::SolverDDP(boost::shared_ptr<ShootingProblem> problem)
       regmax_(1e9),
       cost_try_(0.),
       th_grad_(1e-12),
+      th_gaptol_(1e-9),
       th_stepdec_(0.5),
       th_stepinc_(0.01),
       was_feasible_(false) {
@@ -171,7 +172,7 @@ double SolverDDP::calcDiff() {
       const boost::shared_ptr<ActionDataAbstract>& d = datas[t];
       model->get_state()->diff(xs_[t + 1], d->xnext, fs_[t + 1]);
       if(could_be_feasible) {
-        if (fs_[t + 1].lpNorm<Eigen::Infinity>() >= 1e-9) {
+        if (fs_[t + 1].lpNorm<Eigen::Infinity>() >= th_gaptol_) {
           could_be_feasible = false;
         }
       }
@@ -406,6 +407,8 @@ const double& SolverDDP::get_th_stepinc() const { return th_stepinc_; }
 
 const double& SolverDDP::get_th_grad() const { return th_grad_; }
 
+const double& SolverDDP::get_th_gaptol() const { return th_gaptol_; }
+
 const std::vector<Eigen::MatrixXd>& SolverDDP::get_Vxx() const { return Vxx_; }
 
 const std::vector<Eigen::VectorXd>& SolverDDP::get_Vx() const { return Vx_; }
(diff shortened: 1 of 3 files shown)


src/core/solvers/ddp.cpp
```diff
@@ -164,11 +164,16 @@ double SolverDDP::calcDiff() {
     const std::size_t& T = problem_->get_T();
     const std::vector<boost::shared_ptr<ActionModelAbstract> >& models = problem_->get_runningModels();
     const std::vector<boost::shared_ptr<ActionDataAbstract> >& datas = problem_->get_runningDatas();
+
+    bool could_be_feasible = true;
     for (std::size_t t = 0; t < T; ++t) {
       const boost::shared_ptr<ActionModelAbstract>& model = models[t];
       const boost::shared_ptr<ActionDataAbstract>& d = datas[t];
       model->get_state()->diff(xs_[t + 1], d->xnext, fs_[t + 1]);
+      if (fs_[t + 1].lpNorm<Eigen::Infinity>() >= 1e-9) could_be_feasible = false;
```
cmastalli, line 173: To avoid unnecessary evaluations of this norm, then let's include a if-sentence, i.e. we need to evaluate it if 'could_be_feasible' is false.

cmastalli (commented):
Thanks @proyan.

I was thinking that, at this point, is no needed the 'isFeasible' argument in 'solve'.

proyan:
actually I agree. If we are setting it from the gaps, we don't need isFeasible anymore.

cmastalli:
> actually I agree. If we are setting it from the gaps, we don't need isFeasible anymore.

Then, it could be good to handle it in this PR. Do not forget to deprecate the current signature

proyan:
This would require changes in all solvers. I'll keep this patch as self-contained, and handle it in a different pr

cmastalli:
> This would require changes in all solvers. I'll keep this patch as self-contained, and handle it in a different pr

Ok, it sounds good as well.

src/core/solvers/ddp.cpp
```diff
@@ -164,11 +164,20 @@ double SolverDDP::calcDiff() {
...
     const std::vector<boost::shared_ptr<ActionModelAbstract> >& models = problem_->get_runningModels();
     const std::vector<boost::shared_ptr<ActionDataAbstract> >& datas = problem_->get_runningDatas();
+
+    bool could_be_feasible = true;
     for (std::size_t t = 0; t < T; ++t) {
       const boost::shared_ptr<ActionModelAbstract>& model = models[t];
       const boost::shared_ptr<ActionDataAbstract>& d = datas[t];
       model->get_state()->diff(xs_[t + 1], d->xnext, fs_[t + 1]);
+      if(could_be_feasible) {
+        if (fs_[t + 1].lpNorm<Eigen::Infinity>() >= 1e-9) {
```
jcarpent, line 174: I do think that this value of 1e-9 should be set from an option structure to let the end user selecting the best value for his/her own problem.

cmastalli (approved):
Everything looks alright.

Thanks @proyan for improving the feasibility evaluation.

proyan:
I'll merge this since Travis is passing, and gitlab-ci is either being allowed to fail, or failing because of time-out.

PR #851, opened 2020-09-17
crocoddyl logo

proyan:


jcarpent:
I would remove the next figure, the one below the logo. It is too much.

cmastalli:
> I would remove the next figure, the one below the logo. It is too much.

@jcarpent  I like that figure. It gives a good impression IMHO.

cmastalli (approved):
I am happy with the changes.
I don't think we need to modify other parts of the readme file.
Probably, I would add the figure in the doxygen documentation (cover page). However this could be done later.

jcarpent:
The figure should appear later in the page now the logo is present. I suggest that for readability and clarity.

cmastalli:
> The figure should appear later in the page now the logo is present. I suggest that for readability and clarity.

Ok, this sounds good for me as well. But I would say "still close to the introduction".

proyan:
I've moved the image to Crocoddyl features.

jcarpent (approved):
Much better I think. Thanks a lot for considering this change.

PR #860, opened 2020-09-26
Fix build with C++98

nim65s:
Hi,

As we didn't yet state that we support only C++ >= 11, I think we must continue to ensure that everything builds fine on C++98.
Here are 2 tiny fixes for this.

While here, I added a newline to make the linters happy.

Also, please note that some tests are not passing on some distributions:
- on 16.04, `test_boxqp.cpp(70): error in "boost::bind(&test_unconstrained_qp)": check (sol.x - xkkt).isMuchSmallerThan(1.0, 1e-9) failed`
-  on Debian stretch & buster, some tests are passing, but boost segfaults right after, and they are marked as failed. This is an anoying issue that we have in many projects. I sometimes managed to workaround it by splitting some boost unit tests, but this might indicate that some memory management is not good.

Best,
Guilhem.


cmastalli:
@nim65s thanks for fixing it.
These lines were merged a few PR ago, unfortunately, we didn't realize the issue

Should we enforce c++98 in our CI setups?
Of course, if we want to compile codegen, then it is not possible. But we are not yet there :)

nim65s:
Since the last release, C++98 is already tested on the 16.04 builds on gitlab :)

nim65s:
But the unit-tests are not passing on 16.04, so there is still an "allow-failure" flag.

cmastalli:
Good to know :)

Issue #859, opened 2020-09-25
No to_python (by-value) converter found for C++ type: boost::shared_ptr<crocoddyl::SolverFDDP>


edantec:
When prototyping in Python, I noticed there is no python to C++ binding for a shared pointer of SolverFDDP object. There is one for `boost::shared_ptr<crocoddyl::SolverDDP>` though.

cmastalli:
Because we don't need to bind any shared pointer for any solver. The code that you mentioned (for the DDP solver) is indeed not needed.

Are you facing an issue with this?

edantec:
I have a C++ function which returns a `boost::shared_ptr<crocoddyl::SolverFDDP>` object, and I have Python bindings for this function. When trying to use this function in Python I encounter the error `No to_python (by-value) converter found for C++ type: boost::shared_ptr<crocoddyl::SolverFDDP>`. I do not have this error when I change the type from SolverFDDP to SolverDDP in C++.

proyan:
Are you changing the #include to fddp as well?

edantec:
Yes, I included `#include <crocoddyl/core/solvers/fddp.hpp>` but still the same error

cmastalli:
> Yes, I included `#include <crocoddyl/core/solvers/fddp.hpp>` but still the same error

I don't understand this comment.  You get this message when you haven't register a converter.
Check out the latest devel branch, it should be solved.

proyan:
@edantec could you rebase on top of the current devel? #861 should have solved your issue.

cmastalli:
solved by #861

PR #861, opened 2020-09-28
Register converters for shared_ptrs of solver classes

cmastalli:
This PR should solve the request in #859.

@edantec could you test it from your side?


cmastalli:
I will merge this PR, but it would be still useful if this fixes your issue @edantec.

edantec:
I'll test it as soon as possible, thank you

edantec:
It solves my problem and I can now use FDDP with python bindings, thank!

cmastalli:
> It solves my problem and I can now use FDDP with python bindings, thank!

Happy for that :)

PR #862, opened 2020-10-06
Minor modification: initialized a few QP data

cmastalli:


proyan:
@cmastalli Please wait for a review before handling this.

cmastalli:
@proyan I did a mistake. I don't want to merge this PR now.
But it would be happy if you provide any feedback.

We have a primer version of ALTRO that uses this work.

PR #865, opened 2020-10-07
Memory alignement of frame classes with boost python

### LINKED ISSUE #864: Memory alignement issue with boost python
I'm facing a memory alignment issue, which I hope @jmirabel or @jcarpent could help with:

When trying to access the `reference` oject of `CostModelContactForce`  in python, (which is of type `crocoddyl::FrameForceTpl<double>` defined [here](https://github.com/loco-3d/crocoddyl/blob/devel/include/crocoddyl/multibody/frames.hpp#L121)), sometimes I encounter a segfault. It doesn't come always, sometimes it works, and sometimes it doesn't.

With @nmansard, we made a simple test based on the existing `utils/biped.py` file, in order to check what is happening. We were able to reproduce the issue in both our systems. 

The file is attached. When I try to access this `reference` object after the simulation, I get the segfault. 

This is the line which leads to segfault:
```python
a=m.differential.costs.costs['force16'].cost.reference
```

When investigating further, the error seems to be coming while allocating creating an instance of pinocchio::ForceTpl<double> object in python. `gdb` tells me that the segfault arrives _after_ the code has left `get_reference` in the python bindings, and when the new python instance of `FrameForce` object (and consequently the `pinocchio::Force` object) is being created.

When I look at the assembly code of what is happening, it looks clear that this is a memory alignment issue. The segfault is coming from this line in Eigen:

```gdb
Thread 1 "ipython" received signal SIGSEGV, Segmentation fault.
0x00007f1432dc2306 in Eigen::DenseStorage<double, 6, 6, 1, 0>::DenseStorage (this=0x7f142c675bc8, other=...) at /usr/include/eigen3/Eigen/src/Core/DenseStorage.h:194
194	    DenseStorage(const DenseStorage& other) : m_data(other.m_data) {
```
where the 6 dimensional m_data object is being copied from the pinocchio::Force object to the boost python instance (not sure which).  In order to do this copy, following are the assembly instructions:

```asm
; The following lines move the m_data from the pinocchio::Force object (at memory location given by %r14) to the register xmm. 
; Each line moves two doubles to the register.

movaps 0x10(%r14),%xmm0        
movaps 0x20(%r14),%xmm1
movaps 0x30(%r14),%xmm2

; The following lines move the m_data from the registers to the memory location given by %rbx
movaps %xmm0,0x50(%rbx)   ; Segfaults
movaps %xmm1,0x60(%rbx)
movaps %xmm2,0x70(%rbx)
```

If I look at the address `r14`, I see that it points to an address that is 128 bit aligned in memory. 

```gdb
(gdb) info registers r14
r14            0x7fffbd4924d0      140736369075408
```
Thus, the first `mov` succeeds.
However, the same is not true for the register `rbx`

```gdb
(gdb) info registers rbx
rbx            0x7f142c675b78      139724621044600
```
The address ends with 78, and is not 128 bit aligned in memory, and I think this is the reason that the segfault appears.

CostModelContactForce`, FrameForce, and pinocchio::Force, all have the macro `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` defined in public, and thus they are all always aligned in memory. However, the boost python class that creates binds these quantities don't have this macro. Hence, I suspect that the object instance is being created at a misaligned memory location by boost python. 

My question is, since this issue is not appearing in pinocchio::Force, what is being done differently there that is not being done here when we do our bindings? For reference, the class `ContactForce` is being binded here: https://github.com/loco-3d/crocoddyl/blob/devel/bindings/python/crocoddyl/multibody/costs/contact-force.cpp#L91

In order to reproduce, please run this file in python (https://gist.github.com/proyan/61bf620ee887b872e1c895af905940b3) , and then run the following line
```python
a=m.differential.costs.costs['force16'].cost.reference
```
It may, or maynot work (based on the alignment, there is a 50% chance that some change in the code would make it not working). The assembly trace above is for the cases it doesn't work.

proyan:
Fixes #864

### DIFF
--- a/bindings/python/crocoddyl/multibody/frames.cpp
+++ b/bindings/python/crocoddyl/multibody/frames.cpp
@@ -16,7 +16,6 @@
 
 #include "pinocchio/bindings/python/utils/std-aligned-vector.hpp"
 
-
 EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(crocoddyl::FramePlacement)
 EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(crocoddyl::FrameForce)
 EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(crocoddyl::FrameMotion)
@@ -213,10 +212,10 @@ void exposeFrames() {
 
   pinocchio::python::StdAlignedVectorPythonVisitor<FrameRotation, true>::expose("StdVec_FrameRotation");
 
-  pinocchio::python::StdAlignedVectorPythonVisitor<FrameTranslation, true>::expose("StdVec_FrameTranslation");  
+  pinocchio::python::StdAlignedVectorPythonVisitor<FrameTranslation, true>::expose("StdVec_FrameTranslation");
   pinocchio::python::StdAlignedVectorPythonVisitor<FrameForce, true>::expose("StdVec_FrameForce");
 
-  pinocchio::python::StdAlignedVectorPythonVisitor<FrameMotion, true>::expose("StdVec_FrameMotion");  
+  pinocchio::python::StdAlignedVectorPythonVisitor<FrameMotion, true>::expose("StdVec_FrameMotion");
 }
 
 }  // namespace python
--- a/bindings/python/crocoddyl/multibody/frames.cpp
+++ b/bindings/python/crocoddyl/multibody/frames.cpp
@@ -6,10 +6,20 @@
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
+#include <eigenpy/memory.hpp>
+#include <eigenpy/eigen-to-python.hpp>
+
 #include "python/crocoddyl/multibody/multibody.hpp"
 #include "crocoddyl/multibody/frames.hpp"
 #include "python/crocoddyl/utils/printable.hpp"
 #include "python/crocoddyl/utils/deprecate.hpp"
+
+#include "pinocchio/bindings/python/utils/std-aligned-vector.hpp"
+
+
+EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(crocoddyl::FramePlacement)
+EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(crocoddyl::FrameForce)
+EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(crocoddyl::FrameMotion)
 
 namespace crocoddyl {
 namespace python {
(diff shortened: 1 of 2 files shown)


include/crocoddyl/multibody/frames.hpp
```diff
@@ -41,6 +41,11 @@ struct FrameTranslationTpl {
     return os;
   }
 
+  template <typename OtherScalar>
+  bool operator==(const FrameTranslationTpl<OtherScalar>& other) const {
+    return id == other.id && translation == other.translation.template cast<Scalar>();
```
jcarpent, line 46: Should not be good to do a cast here.

jcarpent (changes requested):

include/crocoddyl/multibody/frames.hpp
proyan: I forgot what type `translation` was. Thanks for the catch!

proyan: ah no. It seems to be okay. Why is the cast not a good idea here?

jcarpent: Because if cast a Scalar type to a larger type, then you will lose resolutions and things won't match.
The operator== should only be with the type `FrameTranslationTpl<Scalar>`

bindings/python/crocoddyl/multibody/frames.cpp
```diff
@@ -198,6 +207,15 @@ void exposeFrames() {
                             deprecated<bp::return_value_policy<bp::return_by_value> >("Deprecated. Use id.")),
           bp::make_function(&FrameCoPSupport::set_id, deprecated<>("Deprecated. Use id.")), "frame ID")
       .def(PrintableVisitor<FrameCoPSupport>());
+
+  pinocchio::python::StdAlignedVectorPythonVisitor<FramePlacement, true>::expose("StdVec_FramePlacement");
```
cmastalli, line 211: Can we use our implementation instance (see below)? 
https://github.com/loco-3d/crocoddyl/blob/master/bindings/python/crocoddyl/utils/vector-converter.hpp#L117

It is good to do not depend on internal choices that Pinocchio developers might make.
Without saying that a developer needs to go internally into Pinocchio for understanding how the binds are done.

proyan: > Can we use our implementation instead (see below)?
https://github.com/loco-3d/crocoddyl/blob/master/bindings/python/crocoddyl/utils/vector-converter.hpp#L117
This is an aligned vector, and not an std::vector.

I'm not in favour of duplicating code in crocoddyl. I think the better place for keeping this class is in eigenpy rather than pinocchio, but @jcarpent can answer to that. However, I don't think duplicating upstream code at multiple places is good practice. It might have been okay to copy if we have to depend on a new library otherwise, but we already have this dependency.

cmastalli: I agree with you @proyan, this should go into Eigenpy. Does @jcarpent  agree with us?

The question is what do we now (during this transition period)?
I see the following options:
  1. Use Crocoddyl implementation and wait for Eigenpy
  2. Use a mixture of implementations (Crocoddyl and Pinocchio); this is the current solution
  3. Remove the Crocoddyl option and update the entire code with Pinocchio (then substitute Pinocchio implementation once it has ported the code to Eigenpy)

I prefer to follow option 1 now. The reason is due to it requires less work for us.
I would discard the option 2 because it is a dirty solution.
I am OK if you want to commit for option 3 in this PR, but this is an avoidable effort from your side (I would say).

jcarpent: I would suggest first applying 3. We will later for porting this code in EigenPy.

cmastalli: The historical reason of this divergence is that
  1. Pinocchio doesn't have Pythonic support in the very begining.
  2. Pinocchio doesn't have std map support (not sure if it has now to be honest).

jcarpent: It supports well std::map. Apparently, you are not following the current devs.

proyan: I'll merge this PR for now. 
@cmastalli If you want, we can continue this conversation in a dedicated issue.

cmastalli: No need to discuss.
I am happy if you handle in another PR.

proyan: Just to be clear, I'm not in favor of duplicating code from one package to another. So any PR that I make would be to remove this duplication :)

cmastalli: Thanks for ensuring we are in the same page.
Yes, I am in favor to remove the duplicated code as well.

PR #867, opened 2020-10-08
DDP objects change size when the actionmodels.nu size < nu_max. Fix this bug

proyan:
This PR deals with a bug introduced with #787 . 

All ddp (and derived classes) objects that depend on the action model nu (e.g. `ddp.k`, `ddp.Qu` etc) are stored in an std::vector<>. This std::vector is initialized with each object having a size nu_max (number of rows or number of cols). 

In the `ddp.solve()`, we are assigning to these objects directly from the problem models. For e.g., in https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solvers/ddp.cpp#L221, `Qu` (which is size `nu_max`) is being assigned to an object `d->Lu` (which is size `nu`). Since `nu_max`> `nu`, we haven't had any segfaults yet. But this is problematic implementation.

This PR fixes this bug. 

TODO: 
~~1) Box-DDP and Box-FDDP solvers~~
~~2) KKT solver~~
~~3) add multiple nu action models in unittests~~

### DIFF
--- a/src/core/solvers/kkt.cpp
+++ b/src/core/solvers/kkt.cpp
@@ -257,7 +257,7 @@ void SolverKKT::allocateData() {
   nu_ = 0;
   const std::size_t& nx = problem_->get_nx();
   const std::size_t& ndx = problem_->get_ndx();
-  //const std::size_t& nu = problem_->get_nu_max();
+  // const std::size_t& nu = problem_->get_nu_max();
   for (std::size_t t = 0; t < T; ++t) {
     if (t == 0) {
       xs_try_[t] = problem_->get_x0();
--- a/src/core/solvers/kkt.cpp
+++ b/src/core/solvers/kkt.cpp
@@ -257,13 +257,14 @@ void SolverKKT::allocateData() {
   nu_ = 0;
   const std::size_t& nx = problem_->get_nx();
   const std::size_t& ndx = problem_->get_ndx();
-  const std::size_t& nu = problem_->get_nu_max();
+  //const std::size_t& nu = problem_->get_nu_max();
   for (std::size_t t = 0; t < T; ++t) {
     if (t == 0) {
       xs_try_[t] = problem_->get_x0();
     } else {
       xs_try_[t] = Eigen::VectorXd::Constant(nx, NAN);
     }
+    const std::size_t& nu = problem_->get_runningModels()[t]->get_nu();
     us_try_[t] = Eigen::VectorXd::Constant(nu, NAN);
     dxs_[t] = Eigen::VectorXd::Zero(ndx);
     dus_[t] = Eigen::VectorXd::Zero(nu);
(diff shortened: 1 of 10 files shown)


unittest/test_solvers.cpp
```diff
@@ -68,6 +68,9 @@ void test_solver_against_kkt_solver(SolverTypes::Type solver_type, ActionModelTy
   boost::shared_ptr<crocoddyl::SolverAbstract> solver =
       boost::static_pointer_cast<crocoddyl::SolverKKT>(solver_factory.create(solver_type, action_type, T));
 
+  //boost::shared_ptr<crocoddyl::SolverAbstract> solverMultipleActions =
```
cmastalli, line 71: Is there a proposed to keep this commented lines?

cmastalli (approved):
@proyan thanks for the fixing.
Are you planning to work on the other topics, you've written in the TODO list, in this PR?

src/core/solvers/kkt.cpp
```diff
@@ -257,13 +257,14 @@ void SolverKKT::allocateData() {
...
   const std::size_t& ndx = problem_->get_ndx();
-  const std::size_t& nu = problem_->get_nu_max();
+  //const std::size_t& nu = problem_->get_nu_max();
   for (std::size_t t = 0; t < T; ++t) {
     if (t == 0) {
       xs_try_[t] = problem_->get_x0();
     } else {
       xs_try_[t] = Eigen::VectorXd::Constant(nx, NAN);
     }
+    const std::size_t& nu = problem_->get_runningModels()[t]->get_nu();
```
cmastalli, line 267: I didn't get this.
Why do you change from `nu_max` to `nu`?

unittest/test_solvers.cpp
proyan: No, I've removed this comment.

src/core/solvers/kkt.cpp
proyan: `kkt` solver doesn't have a notion of nu_max. This should have always been `sum(nu[i])`

cmastalli: Ok, I see. Thanks for the clarification :)

proyan:
Currently, the getter functions such as `solver.get_us()`, `solver.get_k` etc return vector with elements of size `nu_max`, while the actual size is `nu`.
I don't think we can handle it with the current implementation that we have, or by this fix.

cmastalli:
> Currently, the getter functions such as `solver.get_us()`, `solver.get_k` etc return vector with elements of size `nu_max`, while the actual size is `nu`.
> I don't think we can handle it with the current implementation that we have, or by this fix.

Yes, I am not sure if there is a clever solution for that. In my mind, the best we could do is to document this fact. It is always possible to retrieve `nu` by the user (i.e. `solver.problem.getRunningModels()[k].nu`)

cmastalli:
Thanks for improving this implementation detail @proyan :)

PR #869, opened 2020-10-13
Improvement in the smooth-abs activation

cmastalli:
This PR makes two improvements:
  1. Added the smoothing factor (before it wasn't possible to change it)
  2. Improved the documentation for more clarity.

Furthermore, the unittest was adapted in such a way that we could test run tests with random values of the smoothing parameter.


include/crocoddyl/core/activations/smooth-abs.hpp
```diff
@@ -30,20 +42,40 @@ class ActivationModelSmoothAbsTpl : public ActivationModelAbstractTpl<_Scalar> {
...
+   */
   virtual void calc(const boost::shared_ptr<ActivationDataAbstract>& data, const Eigen::Ref<const VectorXs>& r) {
     if (static_cast<std::size_t>(r.size()) != nr_) {
       throw_pretty("Invalid argument: "
                    << "r has wrong dimension (it should be " + std::to_string(nr_) + ")");
     }
     boost::shared_ptr<Data> d = boost::static_pointer_cast<Data>(data);
 
-    d->a = (r.array().cwiseAbs2().array() + Scalar(1)).array().cwiseSqrt();
+    d->a = (r.array().cwiseAbs2().array() + eps_).array().cwiseSqrt();
```
proyan, line 69: This is not the implementation that we discussed.
What we need is sqrt(eps+r**2). This would be somewhat like this:
```
d->a = sqrt(r.array().cwiseAbs2().sum() + 1);
```

Please update the doc for this function as well

include/crocoddyl/core/activations/smooth-abs.hpp
```diff
@@ -55,12 +87,18 @@ class ActivationModelSmoothAbsTpl : public ActivationModelAbstractTpl<_Scalar> {
     data->Arr.diagonal() = d->a.cwiseProduct(d->a).cwiseProduct(d->a).cwiseInverse();
```
proyan, line 87: The derivative would change as well

proyan (changes requested):

cmastalli:
@proyan -- thank you for your feedback, unfortunately, I believe we shouldn't do it (at least in this PR).
See a more elaborated explanation and motivation below:

As far as discussed, I mentioned to do the following:
  - add the smoothing parameter
  - fix the documentation

What you have asked me:
  - it will change the behavior of this class.
  - It might break somebody's code.

This sort of actions could be extremely annoying for a user, and it should be done with a strong justification. 

Said that, in my humble point of view, the current implementation is as assertive as the previous one (maybe better as well). I don't see a motivation to modify the function.

If you wish to change its behavior, I suggest to write an issue.
Then we give us enough time to figure out if the proposed formulation is better.
We might also want to benchmark it against different optimal control problems.

Meantime, I would ask you to accept the current implementation :)

include/crocoddyl/core/activations/smooth-abs.hpp
proyan: Carlos, the current implementation is wrong.
1) It returns an array, and assigns it to a scalar.
2) It doesn't do the computation needed.

What is expected from this function is sqrt(eps+r**2). I think we should change to that.

cmastalli: Think about it in another way.
r is a scalar value, and then you apply the smooth-abs function for each single coefficient of the residual vector.

This is not wrong neither right. There are only two different approaches.

Btw, this is the original implementation of the smooth-abs function. I didn't changed it when I was coding the c++ version.

cmastalli: Well looking twice, I see it was changed but frankly speaking I don't remember I did.
See https://github.com/loco-3d/crocoddyl/blob/master/unittest/python/crocoddyl/activation.py#L160

proyan: After internal discussion the formulation should be kept as implemented.

proyan:
Thanks for the PR carlos.  Could you add the bindings for eps in this PR as well please?

I'm not sure why the test is not passing, but once the CI passes and the bindings are done, I can merge the PR.

cmastalli:
> Thanks for the PR carlos. Could you add the bindings for eps in this PR as well please?
> 
> I'm not sure why the test is not passing, but once the CI passes and the bindings are done, I can merge the PR.

There bindings are already there. Note that `eps` is an optional value, its default value is 1 for compatibility reason.

It is weird that the unittest is not passing. I need more time to investigate it this issue. I will let you know when it is done!

proyan:
ah yes. Then please handle the CI issues.

proyan (approved):

proyan:
We should deprecate smooth-abs, and rename it to smooth-1norm to keep with the convention.
We can do it either in this PR, or a separate one. @cmastalli Would you have time to handle it immediately? If not, I don't think it is worth to hold this PR and we can merge it when CI is ready.

cmastalli:
> We can do it either in this PR, or a separate one. @cmastalli Would you have time to handle it immediately? If not, I don't think it is worth to hold this PR and we can merge it when CI is ready.

It would try to handle (i.e. renamed the function) in this PR. If I can't, then I will create an issue so we can track it.
Tomorrow, I will spend some time trying to figure out the issue with the CI as well.

proyan:
In order to deprecate it, I would suggest to

1) rename smooth-abs to smooth-1norm (check caps) everywhere.

2) In the `fwd.hpp`, add a typedef that defines smooth-abs from smooth-1norm
3) deprecate this typedef

bindings/python/crocoddyl/deprecation.py
```diff
@@ -0,0 +1,8 @@
+from . import libcrocoddyl_pywrap as _crocoddyl
+from .deprecated import deprecated, DeprecatedWarning
+
+
+class ActivationModelSmoothAbs(_crocoddyl.ActivationModelSmooth1Norm):
+    @deprecated("crocoddyl::ActivationModelSmoothAbs is deprecated: Use ActivationModelSmooth1Norm")
+    def __init__(self, nr, eps=1.):
```
proyan, line 7: Since you are inheriting, you don't need to redefine the `__init__` function.

include/crocoddyl/core/fwd.hpp
```diff
@@ -193,8 +195,10 @@ typedef ActivationModelWeightedQuadraticBarrierTpl<double> ActivationModelWeight
 typedef ActivationModelQuadTpl<double> ActivationModelQuad;
 typedef ActivationModelWeightedQuadTpl<double> ActivationModelWeightedQuad;
 typedef ActivationDataWeightedQuadTpl<double> ActivationDataWeightedQuad;
-typedef ActivationModelSmoothAbsTpl<double> ActivationModelSmoothAbs;
-typedef ActivationDataSmoothAbsTpl<double> ActivationDataSmoothAbs;
+DEPRECATED("Use ActivationModelSmooth1Norm", typedef ActivationModelSmooth1NormTpl<double> ActivationModelSmoothAbs;)
+DEPRECATED("Use ActivationDataSmooth1Norm", typedef ActivationDataSmooth1NormTpl<double> ActivationDataSmoothAbs;)
```
proyan, line 199: Thanks for adding this here.

We also need to make `template <typename Scalar>ActivationModelSmoothAbsTpl` available as a deprecated class.

I think the way to do it that is still compatible with <C++11 standards would be to inherit from the new class

```
template <typename T>
class ActivationModelSmoothAbsTpl : public ActivationModelSmooth1NormTpl<T> { /* deprecated */};
```
and the same for the ActivationDataSmoothAbsTpl.

 Please do that in the file `core/activations/smooth-abs.hpp`, so that the header API isn't broken.

proyan (changes requested):

bindings/python/crocoddyl/deprecation.py
cmastalli: I do in order to include the @deprecated decorator.
As far I know, this is the approach to be taken for deprecating a class in Python.

include/crocoddyl/core/fwd.hpp
cmastalli: done!
I also tested locally that the approach works.

cmastalli:
@proyan there are a couple of extra bits:
  - Registered the shared_ptr of centroidal-momentum cost
  - Created the bindings of ActivationDataSmooth1Norm (they were missing)

It still remains the CI debugging task.

bindings/python/crocoddyl/deprecation.py
proyan: I see. Thanks for the explanation

cmastalli:
@proyan the CI issue should be done now.
Let's see if Travis and Gitlab-ci agree with me xD

include/crocoddyl/core/activations/smooth-1norm.hpp
```diff
@@ -50,8 +50,12 @@ class ActivationModelSmooth1NormTpl : public ActivationModelAbstractTpl<_Scalar>
    * @param[in] nr   Dimension of the residual vector
    * @param[in] eps  Smoothing factor (default: 1.)
    */
-  explicit ActivationModelSmooth1NormTpl(const std::size_t& nr, const Scalar& eps = Scalar(1.))
-      : Base(nr), eps_(eps){};
+  explicit ActivationModelSmooth1NormTpl(const std::size_t& nr, const Scalar& eps = Scalar(1.)) : Base(nr), eps_(eps) {
+    if (eps <= Scalar(0.)) {
```
proyan, line 54: If the user wants eps to be zero, he should be allowed to do that. It is a choice that should be left to the user, not to us

include/crocoddyl/core/activations/smooth-2norm.hpp
```diff
@@ -49,8 +49,12 @@ class ActivationModelSmooth2NormTpl : public ActivationModelAbstractTpl<_Scalar>
    * @param[in] nr   Dimension of the residual vector
    * @param[in] eps  Smoothing factor (default: 1.)
    */
-  explicit ActivationModelSmooth2NormTpl(const std::size_t& nr, const Scalar& eps = Scalar(1.))
-      : Base(nr), eps_(eps){};
+  explicit ActivationModelSmooth2NormTpl(const std::size_t& nr, const Scalar& eps = Scalar(1.)) : Base(nr), eps_(eps) {
+    if (eps <= Scalar(0.)) {
```
proyan, line 53: Same. here. I think this should be `eps < Scalar(0)`

proyan (changes requested):

include/crocoddyl/core/activations/smooth-1norm.hpp
cmastalli: If we accept `eps=0`, then we don't have a smooth function.
Indeed, it becomes the absolute function, which it is not useful neither (there is a jump in the derivatives at the origin).

This was the rationality of imposing `<=`. I would vote to keep it. Do you agree with it?

include/crocoddyl/core/activations/smooth-2norm.hpp
cmastalli: In this case the motivation is slightly different.
If we accept `eps=0`, then this function becomes the quadratic one, which it is already defined.

We could accept the zero condition even if this is a weird choice since we could use the other activation model.

include/crocoddyl/core/activations/smooth-1norm.hpp
proyan: The default is 1. So the user is _definitely_ going to use zero by mistake.
Now if he _wants_ to set it to zero, we should not babyfeed him that it is not the expected use of this function. The user has chosen this function, and he should be allowed to play with it.

Thinking clearly, if he wants to set it to a negative value, this is a conscious choice on his part. It should not be restricted from our side. I think we should remove this check completely.

include/crocoddyl/core/activations/smooth-2norm.hpp
proyan: 1) what is defined is the squarenorm, not the norm.
2) But that is beside the point. Let  the user use this function in the manner that is useful to his/her application. We should not restrict from our side. Please see the previous comment

cmastalli: done

include/crocoddyl/core/activations/smooth-1norm.hpp
```diff
@@ -51,10 +51,15 @@ class ActivationModelSmooth1NormTpl : public ActivationModelAbstractTpl<_Scalar>
    * @param[in] eps  Smoothing factor (default: 1.)
    */
   explicit ActivationModelSmooth1NormTpl(const std::size_t& nr, const Scalar& eps = Scalar(1.)) : Base(nr), eps_(eps) {
-    if (eps <= Scalar(0.)) {
+    if (eps < Scalar(0.)) {
       throw_pretty("Invalid argument: "
                    << "eps should be a positive value");
     }
+    if (eps == Scalar(0.)) {
+      std::cerr
```
proyan, line 59: Crap. Sorry Carlos, I'm not thinking clearly. I was thinking this was an assert on eps <=0. But this was a throw, and a throw should be valid for this case.

Could you please restore the previous version, and use this new warning in the throw that you just added? Warning: eps=0 leads to derivatives discontinuoties in the origin, it becomes the absolute function

include/crocoddyl/core/activations/smooth-1norm.hpp
cmastalli: Ok, I have done the follows:
  1. Accept eps=0
  2. Print a warning message for eps=0 

To be honest our solvers cannot handle 2 (derivative base).
A warn message might be useful for users that are not aware of it.

include/crocoddyl/core/activations/smooth-1norm.hpp
cmastalli: hahaha I think you are tired.

You were right before, and not in this statement. Let me explain you :)

Given the following code
```python
import crocoddyl
a = crocoddyl.ActivationModelSmooth1Norm(3, 0.)
```
With the previous implementation, this code fails:
```python
----> 1 a = crocoddyl.ActivationModelSmooth1Norm(3, 0.)

RuntimeError: In /home/cmastalli/devel/crocoddyl/include/crocoddyl/core/activations/smooth-1norm.hpp
crocoddyl::ActivationModelSmooth1NormTpl<Scalar>::ActivationModelSmooth1NormTpl(const size_t&, const Scalar&) [with _Scalar = double; std::size_t = long unsigned int; crocoddyl::ActivationModelSmooth1NormTpl<Scalar>::Scalar = double] 56
Invalid argument: eps should be a positive value
```

Instead with the current code (i.e. using std::cerr), it doesn't fail (only warns you):
```python
Warning: eps=0 leads to derivatives discontinuities in the origin, it becomes the absolute function
```

So, the current code has the desired behavior that you requested.

proyan: We don't have something better than `std:;cerr` to throw warnings?

cmastalli: No, we don't. We are using in a few places `std::cerr`:
```bash
include/crocoddyl/multibody/friction-cone.hxx:    std::cerr << "Warning: nf has to be an even number, set to 4" << std::endl;
include/crocoddyl/multibody/friction-cone.hxx:    std::cerr << "Warning: normal is not an unitary vector, then we normalized it" << std::endl;
include/crocoddyl/multibody/friction-cone.hxx:    std::cerr << "Warning: min_nforce has to be a positive value, set to 0" << std::endl;
include/crocoddyl/multibody/friction-cone.hxx:    std::cerr << "Warning: max_nforce has to be a positive value, set to maximum value" << std::endl;
include/crocoddyl/multibody/wrench-cone.hxx:    std::cerr << "Warning: nf has to be an even number, set to 16" << std::endl;
include/crocoddyl/multibody/wrench-cone.hxx:    std::cerr << "Warning: min_nforce has to be a positive value, set to 0" << std::endl;
include/crocoddyl/multibody/wrench-cone.hxx:    std::cerr << "Warning: max_nforce has to be a positive value, set to maximun value" << std::endl;
include/crocoddyl/core/activations/smooth-1norm.hpp:      std::cerr
include/crocoddyl/core/integrator/euler.hxx:    std::cerr << "Warning: dt should be positive, set to 1e-3" << std::endl;
include/crocoddyl/core/integrator/rk4.hxx:    std::cerr << "Warning: dt should be positive, set to 1e-3" << std::endl;
```

If you know a more efficient way, then let me know. We could handle it in another PR.

cmastalli:
@proyan thanks for the review.

PR #873, opened 2020-10-20
add deprecated headers, fix #872

### LINKED ISSUE #872: FIx API change introduced from #857
The change in the header files, introduced in #857 breaks downstream packages which were using the previous header structure.

@cmastalli could you please take care of this.
Could you please 
1) recreate the original files as empty header files.
2) include the new header files inside the original headers
3) and deprecate the original?

nim65s:


cmastalli (approved):
Thanks @nim65s.
It looks OK for me.

Do you want me to create a release note after this being merged?

proyan:
Thanks for the quick PR Guilhem!
Is it possible to change the #pragma message to 
#pragma DEPRECATED(msg), to follow the deprecation structure we had before?

nim65s:
@cmastalli : as you prefer
@proyan : I found the [`DEPRECATED(msg, func)`](https://github.com/loco-3d/crocoddyl/blob/master/include/crocoddyl/core/utils/deprecate.hpp#L18) macro for functions and methods, but for whole files or includes. Did I miss anything ?

cmastalli:
> @proyan : I found the [`DEPRECATED(msg, func)`](https://github.com/loco-3d/crocoddyl/blob/master/include/crocoddyl/core/utils/deprecate.hpp#L18) macro for functions and methods, but nothing for whole files or includes. Did I miss anything ?

Yes, that's correct. `deprecated` macro cannot be used in this case.
I think the @nim65s approach is OK.

proyan (approved):

proyan:
You are right, we don't have the possibility of just a deprecation message. That would have to be some other time.

jcarpent:
Please see `PINOCCHIO_PRAGMA_DEPRECATED_HEADER` if you want to have inspiration. The idea of it is that we can change the format of the message in one single place.

proyan:
@nim65s 
Could you please introduce the macro `PRAGMA_DEPRECATED` and `PRAGMA_DEPRECATED_HEADER`( based on https://github.com/stack-of-tasks/pinocchio/blob/master/src/macros.hpp#L31)
These two macros would supplement the macro [`DEPRECATED`](https://github.com/loco-3d/crocoddyl/blob/master/include/crocoddyl/core/utils/deprecate.hpp#L18) that we already have

proyan:
Merci Justin

proyan:
@nim65s You may need to format the code

nim65s:
```
-#CROCODDYL_PRAGMA_DEPRECATED_HEADER(crocoddyl/multibody/costs/control.hpp,crocoddyl/core/costs/control.hpp)
+#CROCODDYL_PRAGMA_DEPRECATED_HEADER(crocoddyl / multibody / costs / control.hpp, \
+                                    crocoddyl / core / costs / control.hpp)
```

I'm not sure that's what we really want… But I'll fix it :)

proyan:
Hi Guilhem, I think you forgot to push :)

proyan:
ping :)

include/crocoddyl/core/utils/deprecate.hpp
```diff
@@ -1,7 +1,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2020, University of Edinburgh
+// Copyright (C) 2020, University of Edinburgh, CNRS, INRIA
```
cmastalli, line 4: Is OK to use CNRS?
I mean we always use LAAS-CNRS.

nim65s:
Fixed :)

include/crocoddyl/core/utils/deprecate.hpp
nim65s: CNRS is the real holder of copyright here. LAAS is not mandatory, but not an issue. If you prefer to uniformize, I can change it here.

cmastalli: No, I don't have any issue.

Issue #810, opened 2020-07-28
[NumDiff / Maybe other files as well] calcDiff dependency on calc should be enforced


ksyy:
Hi,

I recently ran into some problems while doing the [Cartpole Jupyter notebook](https://github.com/loco-3d/crocoddyl/blob/devel/examples/notebooks/cartpole_swing_up.py) provided as tutorials for Crocoddyl, because of the *not-so-well* documented dependency that the `calcDiff` function of `DifferentialActionModelNumDiff` has on its own `calc` function.

Indeed, if one calls [`calcDiff`](https://github.com/loco-3d/crocoddyl/blob/13fbdc2d2a419a426b6d0ec246d52c1bd1d030e7/include/crocoddyl/core/numdiff/diff-action.hxx#L46) without calling [`calc`](https://github.com/loco-3d/crocoddyl/blob/13fbdc2d2a419a426b6d0ec246d52c1bd1d030e7/include/crocoddyl/core/numdiff/diff-action.hxx#L28) beforehand, `data_0` is zero-initialised and therefore, all computations that follow are wrong.
I have nothing against that except that there is no warning at all, and the [solution provided](https://github.com/loco-3d/crocoddyl/blob/devel/examples/notebooks/solutions/cartpole_fxfu.py) falls into this trap.

I think there should be a `recalc=True` parameter somewhere in this function, or at least something clear to tell the user that he might not receive a meaningful result.

What do you think about that?

cmastalli:
Hi @kssy,

We have removed `recalc=True` to avoid branchiation when we code generate a `calcDiff` function. It is expected to run first `calc`. Furthermore, this behavior should be followed in the numdiff routine.

It is truth that we are missing to add documentation in many places of the library. This is a difficult task since we are at the same time pushing forward our research projects (we lack of resources).

Feel free to help us with the documentation, this is an entry task that could be done without knowing the complex implementation details.

cmastalli:
I will close this issue since this request goes against our design choice.

Feel free to make questions if you still have doubts about our design choice.

In any case, thank you for making propositions.

nim65s:
Well, isn't this still wrong https://github.com/loco-3d/crocoddyl/blob/devel/examples/notebooks/solutions/cartpole_fxfu.py ?

cmastalli:
> Well, isn't this still wrong https://github.com/loco-3d/crocoddyl/blob/devel/examples/notebooks/solutions/cartpole_fxfu.py ?

[...]

PR #876, opened 2020-10-24
Added documentation about calcDiff

cmastalli:
Concretely speaking, it explains that needs to be run first calc in calcDiff.
This information was creating confusion in users as reported in #810.

### DIFF
--- a/bindings/python/crocoddyl/multibody/contact-base.cpp
+++ b/bindings/python/crocoddyl/multibody/contact-base.cpp
@@ -37,7 +37,7 @@ void exposeContactAbstract() {
            "Compute the derivatives of contact holonomic constraint.\n\n"
            "The rigid contact model throught acceleration-base holonomic constraint\n"
            "of the contact frame placement.\n"
-           "It assumes that that calc has been run first.\n"
+           "It assumes that calc has been run first.\n"
            ":param data: contact data\n"
            ":param x: state vector\n")
       .def("updateForce", pure_virtual(&ContactModelAbstract_wrap::updateForce), bp::args("self", "data", "force"),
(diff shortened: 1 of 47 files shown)


bindings/python/crocoddyl/multibody/contact-base.cpp
```diff
@@ -37,6 +37,7 @@ void exposeContactAbstract() {
            "Compute the derivatives of contact holonomic constraint.\n\n"
            "The rigid contact model throught acceleration-base holonomic constraint\n"
            "of the contact frame placement.\n"
+           "It assumes that that calc has been run first.\n"
```
proyan, line 40: You should change "that that" by that :)

I think you can simply run the following command in your crocoddyl folder to change all occurrence of "that that" by "that":

```
find ./ -type f -readable -writable -exec sed -i "s/that that/that/g" {} \;
```

bindings/python/crocoddyl/multibody/contacts/contact-2d.cpp
```diff
@@ -43,6 +43,7 @@ void exposeContact2D() {
            "Compute the derivatives of the 2D contact holonomic constraint.\n\n"
            "The rigid contact model throught acceleration-base holonomic constraint\n"
            "of the contact frame placement.\n"
+           "It assumes that that calc has been run first.\n"
```
proyan, line 46: It is present in multiple documentation changes that you did. The above command should take care of all of them

bindings/python/crocoddyl/multibody/contact-base.cpp
cmastalli: Thanks for the command.
Now it is fixed!

proyan (approved):

PR #878, opened 2020-10-28
Clean up a few things related with cost classes

cmastalli:
This PR makes the following changes:
  - Improve the documentation of numdiff
  - Fixed a grammar mistake in the documentation
  - Moved cost numdiff to core folder

Note that cost numdiff is class used internally for unit-testing.
I have not deprecated the headers due to that reason.


jcarpent (commented):
Question: just to be sure that you did not move files without deprecating their usage, especially if external users are indeed using them.

cmastalli:
> Question: just to be sure that you did not move files without deprecating their usage, especially if external users are indeed using them.

Yes, I didn't deprecate them. It would be weird if this happens.
We are not neither creating bindings of these classes.
@proyan would you prefer to deprecate the header file?

jcarpent:
Yes, but I do think that having a checker with num-diff is important, both in Python and C++, to check that everything goes well. This should be present not only in the unit tests but also in the main corpus of the code. 
This will become a Swiss knife for every user in some sense.
So, at some point, certain users are relying on it.

cmastalli:
> Yes, but I do think that having a checker with num-diff is important, both in Python and C++, to check that everything goes well. This should be present not only in the unit tests but also in the main corpus of the code.
> This will become a Swiss knife for every user in some sense.
> So, at some point, certain users are relying on it.

Currently, we have numdiff routines for action and diff-action models. If you want to check a cost derivative, then you could use them. Of course this is running numdiff over the dynamics as well.
Do you think it is still useful to expose the cost one?
Note that for an user perspective it is harder to use the cost one. The reason is the user needs to be aware of passing the right reeval function (it does the computation expected in the dynamics).

jcarpent:
At least, having access to them in C++ is essential. For Python, certainly too. It allows to do easy code prototyping without implementing the derivatives.

cmastalli:
I have added the deprecation of the header file to be safer.

@jcarpent we could create Python bindings of the numdiff cost, it is not a bad idea.

To do that, at the moment, I encounter a problem which I don't know how to solve.
Let me explain you, so you might know how it could be done.

We require to pass a function that needs to be reevaluated during the numdiff routine. The function is defined as follows
https://github.com/loco-3d/crocoddyl/blob/7e8a6f694cc28d0eb2109fa12b77711c45c87070/include/crocoddyl/multibody/numdiff/cost.hpp#L101
and we use it in places such as:
https://github.com/loco-3d/crocoddyl/blob/7e8a6f694cc28d0eb2109fa12b77711c45c87070/include/crocoddyl/multibody/numdiff/cost.hxx#L54

In c++, what we do is to use boost::bind to define that function, e.g.
https://github.com/loco-3d/crocoddyl/blob/7e8a6f694cc28d0eb2109fa12b77711c45c87070/unittest/test_costs.cpp#L116

But I am not sure how I could bind this function in Python.
Do you know how we should do it?

jcarpent:
I don't understand the concept of `ReevaluationFunction`. Are you sure of your design pattern?

cmastalli:
> I don't understand the concept of `ReevaluationFunction`. Are you sure of your design pattern?

Imagine you want to get the numdiff of a frame placement cost (e.g \|M - Mref\|, where is a constant reference placement).
To do so, you need to calculate the frame placement for a disturbed state, i.e. x + dx.
We then call `cost::calc`, but due to our design, the cost function doesn't perform any Pinocchio computation (this is done by the differential action model).
Therefore, we need to create a mechanism to do these computations inside the numdiff cost body.
This is the role of the re-evaluation function.

Note that this method came from our initial Python implementation. Since there we haven't changed it because for efficiency reason all the computations are performed inside the diff-action models.

jcarpent:
I think to be efficient and versatile, a NumDiff struct should inheritate from the Cost or Dynamics and then it should have internally its own data for computing the deltas.

cmastalli:
> I think to be efficient and versatile, a NumDiff struct should inheritate from the Cost or Dynamics and then it should have internally its own data for computing the deltas.

**About action inheritance**
If we do that, then the numdiff routine depends on a specific action model, right?
For instance, the frame placement cost is used in two different differential action models (free / contact) and one action model (impulse).
And of course we want to check the cost derivatives across all the differential action models

But there is another option, we could construct the numdiff object by passing an action model pointer (not inheritance).
This solution is plausible, but I am still favor with the one we already designed.
The reason is due to the fact that it is more plain our current solution.
I mean the user can clearly define the re-evaluation functions needed for his/her cost function.
For instance, some cost function might be designed to work with a specific set of action models.

**About cost inheritance**
Instead for the cost case, I don't understand this proposition.
Note that all the computations are always carried out in the action models.
Therefore, it won't be useful to inherit from a cost model, right?
And if this could be, it will remain the same issue described above.

jcarpent:
Yes, sure no problem. I had in mind the CRTP version of this, but as you are relying on classic Inheritence with common base API, then this is fine for me.

cmastalli:
Then, we should come back to the original question.
Do you know how to bind this function in Python?
If there is no a clear solution now, then I will skip to do it in this PR.

jcarpent:
What you can imagine then is to have a class which takes a classic cost or dynamic and then compute for your the derivatives wrt control and state. I would vote for this strategy. Then you expose those classes, which are just wrapper, then easy to use.
Another solution is to add checker that does the job in the core solver when computing derivatives, and check the derivatives at each call.
This is simple to do, and there is no need of exposing them in Python

cmastalli:
> What you can imagine then is to have a class which takes a classic cost or dynamic and then compute for your the derivatives wrt control and state. I would vote for this strategy.

Could you create an issue?
This requires a proper discussion which is not relevant for this PR, I would say.



> Another solution is to add checker that does the job in the core solver when computing derivatives, and check the derivatives at each call.

I don't think it is a good idea. We want to test a cost without running a solver, i.e. to be unit.
It will increase unnecessary the complexity of our unit-test code, without mentioning that the computation time will be much worse (we are already reaching the limits of Travis).

cmastalli:
Can we move on?
@proyan do you have any feedback on this PR?

proyan (approved):

proyan:
@cmastalli @jcarpent If needed, please create an issue and reference this PR to track the history of discussion

PR #879, opened 2020-10-29
update credits

proyan:


README.md
```diff
@@ -191,13 +191,20 @@ The rest of the publications describes different component of **Crocoddyl**:
 You have a question or an issue? You may either directly open a [new issue](https://github.com/loco-3d/crocoddyl/issues) or use the mailing list <crocoddyl@laas.fr>.
 
 
+## Executive Committee
+
+**Crocoddyl** is being maintained by an executive committee which meets every two weeks to discuss the ongoing development.
```
jcarpent, line 196: `developments`

README.md
```diff
@@ -191,13 +191,20 @@ The rest of the publications describes different component of **Crocoddyl**:
 You have a question or an issue? You may either directly open a [new issue](https://github.com/loco-3d/crocoddyl/issues) or use the mailing list <crocoddyl@laas.fr>.
 
 
+## Executive Committee
```
jcarpent, line 194: Steering Committee sound better

README.md
cmastalli: What about to write
``**Crocoddyl** is being managed by  ...``?

I think it is better to use `managed` because we are not only maintaining the code, we are also defining the roadmap of it. Crocoddyl is still an experimental library, we are constantly developing new features.

README.md
cmastalli: Steering is more connected with making higher decision.
Executive is more connected with executing those decisions.

Both are OK, I would say, but they have slightly different meanings.

proyan: It sounds like Steering Committee is ok for both of you. I'll change it to that.

README.md
proyan: ok

cmastalli (approved):

PR #882, opened 2020-10-30
Removed useless typedef declarations in cost functions + Added a function to plot the gaps

cmastalli:


bindings/python/crocoddyl/__init__.py
```diff
@@ -461,6 +461,21 @@ def plotConvergence(costs, muLM, muV, gamma, theta, alpha, figIndex=1, show=True
         plt.show()
 
 
+def plotDynamicFeasibility(gaps, typeOfNorm=None, figIndex=1, show=True, figTitle=""):
```
proyan, line 464: I'm not in favour of these helper functions. These are user specific implementations, and should be handled by the user. We should not be doing the plotting for someone. 

The same goes for the existing "plotConvergence", or "plotOCSolution" that are part of the __init__.py file. Such helpers clutter the script, and don't add anything to the library.

proyan (changes requested):

bindings/python/crocoddyl/__init__.py
cmastalli: > I'm not in favour of these helper functions. These are user specific implementations, and should be handled by the user. We should not be doing the plotting for someone.

Unfortunately, I disagree with you.
For instance, **think about the Gepetto/Meshcat viewers, are they user specific implementations?**
According to your statement the answer is YES, right?
Furthermore, to be honest with you, the viewers are more intrusive than the plot functions since they require very specific libraries.
But the truth is that both (viewers and plot functions) are useful and likely used between among our users.

If we leave the philosophical discussion and go into a pragmatical way of thinking, I could say the follows: **we should remove these extra code is we don't want to maintain them**.
But you need to realize that we are already using these functions in our notebooks.
So, we will need to re-write the notebooks, and they won't be as instructive as before.

> The same goes for the existing "plotConvergence", or "plotOCSolution" that are part of the **init**.py file. Such helpers clutter the script, and don't add anything to the library.

About this topic, I agree with you.
Both viewers and plotters should be moved into their specific namespaces.
I propose to create two namespaces as described below:
```python
import crocoddyl.viewers  # GV / Meshcat 
import crocoddyl.plots    # Plot functions
```

What do you think?

proyan: There is a difference between supporting viewers, and simply creating a wrapper. The lines may get blurry in the middle, but in this situation it is clear enough.

Basically all that we are doing here is putting _our_ custom settings around one line of plot function. This is not our place, it is the user's choice. For example, I know why you want to specify rcParams here.
But this is not an issue that needs to be fixed in crocoddyl. These are personal settings, and should not be part of the library.

proyan: I am all for easing the logging of data with the logger classes, or for adding support for viewer. If we were outputting some hard-to-access data type somewhere, I would be for creating helpers to simplify access. However, it is not good to wrap simple use case scenarios with our preferences on top. What if the user wants a different figsize? or a different fonttype? or a different layout? or a different parameter to plot? 
Let's be rigorous please. If you want to create a wrapper, all these are things that would need to be specified by the user. But such a wrapper is impossible to work with. So let's not bind ourselves to maintaining such wrappers in the first place.

proyan: I agree with your idea about moving the viewer information from ```__init__.py``` to `crocoddyl.viewers`. It would be good to clean up the `__init__.py` and make access more intuitive

cmastalli: > . What if the user wants a different figsize? or a different fonttype? or a different layout? or a different parameter to plot?

You are thinking that these functions should be used for nice plotting, and as consequence, you believe we should provide user settings.
Instead, they are introspective tools and it is not in our interest to support user preferences.
I hope this clarify the role of these functions, which I think you misinterpreted.

As current action point of this PR, I will move the functions as discussed.
I don't think it is productive to keep discussing on this topic, I hope you can understand :)

proyan: I would prefer if you remove these function Carlos. Or we can talk if you want to.

proyan:
The rest of the PR is okay to go, so if you could remove this, I'll merge this and we can discuss this separately?

proyan:
ping :)

cmastalli:
Hi @proyan,

Sorry for the delay. I am very busy, I expect to work on this on Monday.
I will remove the function, but it could be useful to discuss about it later.

Regarding the creation of a different namespace for the viewers, I prefer to handle later since I have low bandwidth now.

Thanks for the feedback, I will ping you when I have removed the function.

Carlos

bindings/python/crocoddyl/__init__.py
cmastalli: done

cmastalli:
@proyan I have removed the functions, I also cleaned up some tiny issues with Python bindings of solver classes.

cmastalli:
@nim65s the coverage failed, could you let me know why?
Thanks :)

nim65s:
> No space left on device

I retried it.

bindings/python/crocoddyl/core/solvers/fddp.cpp
```diff
@@ -46,23 +46,6 @@ void exposeSolverFDDP() {
...
-      .def("computeDirection", &SolverFDDP::computeDirection,
-           SolverFDDP_computeDirections(
-               bp::args("self", "recalc"),
-               "Compute the search direction (dx, du) for the current guess (xs, us).\n\n"
-               "You must call setCandidate first in order to define the current\n"
-               "guess. A current guess defines a state and control trajectory\n"
-               "(xs, us) of T+1 and T elements, respectively.\n"
-               ":params recalc: true for recalculating the derivatives at current state and control.\n"
-               ":returns the search direction dx, du and the dual lambdas as lists of T+1, T and T+1 lengths."))
-      .def("expectedImprovement", &SolverFDDP::expectedImprovement,
```
proyan, line 58: Why are you removing `expectedImprovement`?

cmastalli: This is a duplicated code. Note that this is a virtual function which it reimplemented in this class. Furthermore, this function is already bind. Therefore, we don't need to rebind it again.

The historical context is I have created this code because it was not properly defined as virtual in the first implementation.

proyan: great, thanks.

proyan (approved):

proyan (approved):

PR #883, opened 2020-11-09
[cost] Allowed to set internal reference values in CoP cost

cmastalli:
We have already discussed how to allow reading/writing internal values in cost reference functions.
However, I wanted to share with you know the specific details of this solution.
To make things easier, I have pushed only the modifications for the CoP cost.

As you can see in this PR, it is possible to get/set references without down-casting the functions in c++.
As I mentioned in our discussion, this is the main reason to this design pattern:
```cpp
  using namespace crocoddyl;
  FrameCoPSupport ref(id, Eigen::Vector2d(0.1, 0.1));
  boost::shared_ptr<CostModelAbstract> = boost::make_shared<CostModelCoPPosition>(state, activation, ref);
  // Get references
  FrameCoPSupport ref_full = cost->get_reference<FrameCopSupport>();
  Eigen::Vector2d ref_box = cost->get_reference<Eigen::Vector2d>();
  FrameIndex ref_id = cost->get_reference<FrameIndex>();
  // Set references
  cost->set_reference<FrameCoPSupport>(ref_full);
  cost->set_reference<Eigen::Vector2d>(ref_box);
  cost->set_reference<FrameIndex>(ref_id);
```
And for the Python case, it is also possible to define the references as
```python
  ref = FrameCoPSupport(id, np.array([0.1, 0.1])
  cost = CostModelCoPPosition(state, activation, ref)
  # Get references
  ref_full = cost.reference
  ref_box = cost.reference_box
  ref_id = cost.reference_id
  # Set references
  cost.reference = ref_full
  cost.reference_box = ref_box
  cost.reference_id = ref_id
```
where it is necessary to define a specific name for each reference function.
This is the only difference with respect to the agreed proposition.
Unfortunately, I can't use the same name for each case; it is a boost python limitation.

Before merging this PR, I would need to replicate this scheme for each cost function as described below.
If you have any further feedback, then this is the right moment to do it.

- [x] Added set and get functions for internal references in each cost function
- [x] Created the Python bindings for internal references in each cost function


cmastalli:
I have pushed the extra bits for each cost class.

cmastalli:
There is also an extra task. I removed the `multibody/costs/control-cost.hpp` header file.

It was included by @nim65s with the aim to deprecate it, but we never have it.
Note that we are already deprecating the right header file, i.e. `multibody/costs/control.hpp`.

proyan:
Based on internal discussions, we've decided to not proceed with this design template. So I'm closing this PR.

PR #887, opened 2020-11-22
Enhance readme

jcarpent:


README.md
```diff
@@ -52,7 +52,7 @@ If you want to follow the current developments, you can directly refer to the [d
  * Euclidian and non-Euclidian geometry friendly via **[Pinocchio](https://github.com/stack-of-tasks/pinocchio)**
  * handle autonomous and nonautomous dynamical systems
  * numerical differentiation support
- * automatic differentiation support
+ * automatic differentiation support via **[Pinocchio](https://github.com/stack-of-tasks/pinocchio)**
```
cmastalli, line 55: Automatic differentiation support via CppAD

README.md
```diff
@@ -61,7 +61,7 @@ If you want to follow the current developments, you can directly refer to the [d
  * Python bindings (including models and solvers abstractions)
  * C++ 98/11/14/17/20 compliant
  * extensively tested
- * automatic code generation support
+ * automatic code generation support via **[Pinocchio](https://github.com/stack-of-tasks/pinocchio)**
```
cmastalli, line 64: Via CppADCogen

README.md
```diff
@@ -174,8 +175,16 @@ and the following one to reference this website:
    year = {2019}
 }
 ```
-
-The rest of the publications describes different component of **Crocoddyl**:
+Please also consider citing **[Pinocchio](https://github.com/stack-of-tasks/pinocchio)**, which largely contributes to make **Crocoddyl** an efficient and versatile framework:
```
cmastalli, line 178: I am not sure if we really need to include the bibtex since this information is already included in Pinocchio repository.
I prefer to keep the real contributions of Crocoddyl, e.g. solvers.
Instead, we should include it in the publication list.

jcarpent: I think this is downscaling the role played by Pinocchio, both in terms of performances and versatility. I would really appreciate that people who use Crocoddyl also acknowledge the work made by Pinocchio's developers to support and enhance everything related to robot modelling, collision and differentiation.
In addition, this is also crucial for the current request to get a dedicated software engineer to support Pinocchio and Crocoddyl developments funded by the French institutions.

README.md
jcarpent: I will also add CppADCogen.

README.md
jcarpent: Same here.

README.md
```diff
@@ -116,6 +116,7 @@ export PYTHONPATH=/opt/openrobots/lib/python2.7/site-packages:$PYTHONPATH
 **Crocoddyl** is c++ library with Python bindings for versatile and fast prototyping. It has the following dependencies:
 
 * [pinocchio](https://github.com/stack-of-tasks/pinocchio)
+* [eigenpy](https://github.com/stack-of-tasks/eigenpy)
```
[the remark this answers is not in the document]
jcarpent: You should add Eigen too. This is a direct dependency of your project, not something which is inherited from another dependency. This is how it works for packaging, we should be transparent on it.

jcarpent:
> To acknowledge Pinocchio, the best way is to write something like: for more details about how to cite Pinnochio read the Readme file, instead of including the Pinocchio bibtex. This is also more appropriate since, for instance, we might want to cite the analytical derivatives paper or collision avoidance stuffs.

FIne to refer to the bibtex in Pinocchio, but at least this should be explicitly mentioned. I let you handle it.

cmastalli:
@jcarpent  I have modified the readme as discussed.
I also included the tags for conda stuff.

Please confirm that you agree with it :)

Thank you for integrating Conda, I just have one single question.
Why is written BSD-2 in conda (https://anaconda.org/conda-forge/crocoddyl)?
Is this a mistake?

jcarpent:
> Why is written BSD-2 in conda (https://anaconda.org/conda-forge/crocoddyl)?
> Is this a mistake?

Yes, this my mistake. I've just fixed it.

jcarpent:
Looks good to me.

cmastalli:
Thanks again!
It is very useful indeed the Conda support :)

PR #889, opened 2020-11-26
Devel activation

edantec:
This PR adds two activation functions, quadlog and quadflat. quadlog is log(1 + r^2 / sigma), quadflat is 1 - exp(r^2 / sigma). Quadlog and quadflat allow to have quadratic behaviour near zero and near flat behavior far away from zero.


bindings/python/crocoddyl/core/activations/quadratic-flat.cpp
```diff
@@ -0,0 +1,45 @@
...
+namespace crocoddyl {
+namespace python {
+
+void exposeActivationQuadFlat() {
+  bp::class_<ActivationModelQuadFlat, bp::bases<ActivationModelAbstract> >(
+      "ActivationModelQuadFlat",
+      "Quadratic flat activation model.\n\n"
+      "A quadratic flat action describes a quadratic flat function that depends on the residual, i.e.\n"
+      "1 - exp(*||r||^2 / sigma2).",
+      bp::init<int,double>(bp::args("self", "nr","sigma2"),
```
cmastalli, line 22: Why calling it as sigma2? or better why do we call it as sigma?
(same for the other activation)

bindings/python/crocoddyl/core/activations/quadratic-flat.cpp
```diff
@@ -0,0 +1,45 @@
...
+           ":param data: activation data\n"
+           ":param r: residual vector")
+      .def("calcDiff", &ActivationModelQuadFlat::calcDiff, bp::args("self", "data", "r"),
+           "Compute the derivatives of a quadratic flat function.\n\n"
+           "Note that the Hessian is constant, so we don't write again this value.\n"
+           ":param data: activation data\n"
+           ":param r: residual vector \n")
+      .def("createData", &ActivationModelQuadFlat::createData, bp::args("self"),
+           "Create the quadratic flat activation data.\n\n")
+      .add_property("sigma2",
```
cmastalli, line 37: also here and in `get_sigma2`

include/crocoddyl/core/activations/quadratic-flat.hpp
```diff
@@ -0,0 +1,75 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_FLAT_HPP_
+#define CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_FLAT_HPP_
+
+#include <stdexcept>
+#include "crocoddyl/core/fwd.hpp"
+#include "crocoddyl/core/activation-base.hpp"
+#include "crocoddyl/core/utils/exception.hpp"
+#include <iostream>
```
cmastalli, line 16: This is just a detailed thing.
It would nice if you could move all the system-wide header first, i.e.
```cpp
#include <stdexcept>
#include <iostream>
#include "crocoddyl/core/fwd.hpp"
#include "crocoddyl/core/activation-base.hpp"
```

Note that this is a common good practice that we are following.

include/crocoddyl/core/activations/quadratic-flat.hpp
```diff
@@ -0,0 +1,75 @@
...
+  explicit ActivationModelQuadFlatTpl(const std::size_t& nr,const Scalar& sigma2)
+     : Base(nr), sigma2_(sigma2){};
+  virtual ~ActivationModelQuadFlatTpl(){};
+
+  virtual void calc(const boost::shared_ptr<ActivationDataAbstract>& data, const Eigen::Ref<const VectorXs>& r) {
+    if (static_cast<std::size_t>(r.size()) != nr_) {
+      throw_pretty("Invalid argument: "
+                   << "r has wrong dimension (it should be " + std::to_string(nr_) + ")");
+    }
+    data->a_value = (Scalar(1.0) - exp((-r.transpose() * r)[0] / sigma2_));
```
cmastalli, line 41: Why do you code  ```(-r.transpose() * r)[0]```?

include/crocoddyl/core/activations/quadratic-flat.hpp
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  virtual void calcDiff(const boost::shared_ptr<ActivationDataAbstract>& data, const Eigen::Ref<const VectorXs>& r) {
+    if (static_cast<std::size_t>(r.size()) != nr_) {
+      throw_pretty("Invalid argument: "
+                   << "r has wrong dimension (it should be " + std::to_string(nr_) + ")");
+    }
+    a0_ = Scalar(2.0) / sigma2_ * exp((-r.transpose() * r)[0] / sigma2_);
```
cmastalli, line 49: same here

cmastalli: To avoid duplication of computations, I suggest to create a data for ```exp((-r.transpose() * r)[0] / sigma2_``` and then reuse the value computed in `calc` here.

include/crocoddyl/core/activations/quadratic-flat.hpp
```diff
@@ -0,0 +1,75 @@
...
+    data->Ar = a0_ * r;
+    data->Arr.diagonal() = - Scalar(2.0) * a0_ / sigma2_ * (r * r.transpose()).diagonal();
```
cmastalli, line 51: Actually, it would be good to do the same for ```r * r.transpose()``` so we avoid to run three times the same routine.

include/crocoddyl/core/activations/quadratic-flat.hpp
```diff
@@ -0,0 +1,75 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2018-2019, LAAS-CNRS
```
cmastalli, line 4: License should be
```cpp
Copyright (C) 2020, LAAS-CNRS
````

File: bindings/python/crocoddyl/core/activations/quadratic-log.cpp [hunk 8]
```diff
@@ -0,0 +1,45 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2018-2020, LAAS-CNRS, University of Edinburgh
```
cmastalli, line 4:
```cpp
// Copyright (C) 2020, LAAS-CNRS, University of Edinburgh
```

bindings/python/crocoddyl/core/activations/quadratic-flat.cpp
```diff
@@ -0,0 +1,45 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2018-2020, LAAS-CNRS, University of Edinburgh
```
cmastalli, line 4:
```cpp
// Copyright (C) 2020, LAAS-CNRS, University of Edinburgh
```

include/crocoddyl/core/activations/quadratic-log.hpp
```diff
@@ -0,0 +1,75 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (C) 2018-2019, LAAS-CNRS
```
cmastalli, line 4: same here

include/crocoddyl/core/activations/quadratic-log.hpp
```diff
@@ -0,0 +1,75 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_LOG_HPP_
+#define CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_LOG_HPP_
+
+#include <stdexcept>
+#include "crocoddyl/core/fwd.hpp"
+#include "crocoddyl/core/activation-base.hpp"
+#include "crocoddyl/core/utils/exception.hpp"
+#include <iostream>
```
cmastalli, line 16: same here

include/crocoddyl/core/activations/quadratic-log.hpp
```diff
@@ -0,0 +1,75 @@
...
+  typedef ActivationModelAbstractTpl<Scalar> Base;
+  typedef ActivationDataAbstractTpl<Scalar> ActivationDataAbstract;
+  typedef typename MathBase::VectorXs VectorXs;
+  typedef typename MathBase::MatrixXs MatrixXs;
+
+  explicit ActivationModelQuadLogTpl(const std::size_t& nr,const Scalar& sigma2)
+     : Base(nr), sigma2_(sigma2){};
+  virtual ~ActivationModelQuadLogTpl(){};
+
+  virtual void calc(const boost::shared_ptr<ActivationDataAbstract>& data, const Eigen::Ref<const VectorXs>& r) {
```
cmastalli, line 36: Did you format the code?
This code looks that need to be formatted.
See the wiki for more information about how to do it.

include/crocoddyl/core/activations/quadratic-log.hpp
```diff
@@ -0,0 +1,75 @@
...
+    if (static_cast<std::size_t>(r.size()) != nr_) {
+      throw_pretty("Invalid argument: "
+                   << "r has wrong dimension (it should be " + std::to_string(nr_) + ")");
+    }
+    data->a_value = log(Scalar(1.0) + (r.transpose() * r)[0] / sigma2_);
```
cmastalli, line 41: Create data for ```r.transpose() * r``` and reuse it in calcDiff (as before)

unittest/factory/activation.cpp
```diff
@@ -58,12 +66,19 @@ boost::shared_ptr<crocoddyl::ActivationModelAbstract> ActivationModelFactory::cr
   Eigen::VectorXd lb = Eigen::VectorXd::Random(nr);
   Eigen::VectorXd ub = lb + Eigen::VectorXd::Ones(nr) + Eigen::VectorXd::Random(nr);
   Eigen::VectorXd weights = 0.1 * Eigen::VectorXd::Random(nr);
+  double sigma2 = 10;
```
cmastalli, line 69: What about to create a random double value?
See the code that we have for doing it

include/crocoddyl/core/activations/quadratic-flat.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include <stdexcept>
+#include "crocoddyl/core/fwd.hpp"
+#include "crocoddyl/core/activation-base.hpp"
+#include "crocoddyl/core/utils/exception.hpp"
+#include <iostream>
+
+namespace crocoddyl {
+
+template <typename _Scalar>
```
cmastalli, line 20: Include doxygen documentation (see examples in, for instance, action-base.hpp)

include/crocoddyl/core/activations/quadratic-log.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include <stdexcept>
+#include "crocoddyl/core/fwd.hpp"
+#include "crocoddyl/core/activation-base.hpp"
+#include "crocoddyl/core/utils/exception.hpp"
+#include <iostream>
+
+namespace crocoddyl {
+
+template <typename _Scalar>
```
cmastalli, line 20: also here, do not forget to create doxygen documentation for the class, functions and members

cmastalli (changes requested):
Thanks at lot @edantec 

There are a few things to be improved which I summarize below:
  1. Create extra data for avoiding calculation duplication
  2. Add doxygen documentation in classes
  3. Use random double for unit-test
  4. Clean up header inclusion

bindings/python/crocoddyl/core/activations/quadratic-flat.cpp
edantec: No real reason, I changed it for alpha

include/crocoddyl/core/activations/quadratic-log.hpp
edantec: I tried to clone Gepetto Linter in my home to format the code but apparently I don't have access to the remote repository

nim65s: this repository should be public: https://github.com/gepetto/linters
the docker image too : `docker pull gepetto/linters`

bindings/python/crocoddyl/core/activations/quadratic-flat.cpp
```diff
@@ -0,0 +1,49 @@
...
+namespace python {
+
+void exposeActivationQuadFlat() {
+  bp::class_<ActivationModelQuadFlat, bp::bases<ActivationModelAbstract>>(
+      "ActivationModelQuadFlat",
+      "Quadratic flat activation model.\n\n"
+      "A quadratic flat action describes a quadratic flat function that "
+      "depends on the residual, i.e.\n"
+      "1 - exp(||r||^2 / alpha).",
+      bp::init<int, double>(bp::args("self", "nr", "sigma2"),
```
cmastalli, line 23: You forgot to change this to `alpha` :)

include/crocoddyl/core/activations/quadratic-flat.hpp
```diff
@@ -0,0 +1,155 @@
...
+#include <stdexcept>
+
+namespace crocoddyl {
+
+/**
+ * @brief Quadratic-flat activation
+ *
+ * This activation function describes a quadratic exponential activation for
+ * each element of a residual vector, i.e. \f[ \begin{equation} sum^nr_{i=0} 1 -
+ * exp(\|r_i\|^2 / \alpha) \end{equation} \f] where \f$\alpha\f$ defines the
```
cmastalli, line 25: If I understand correctly, this is not an element-wise operation. If so, you should write:
```cpp
\begin{equation} 1 - exp(\|\mathbf{r}\|^2 / \alpha \end{equation}
```

include/crocoddyl/core/activations/quadratic-log.hpp
```diff
@@ -0,0 +1,152 @@
...
+#include <stdexcept>
+
+namespace crocoddyl {
+
+/**
+ * @brief Quadratic-log activation
+ *
+ * This activation function describes a logarithmic quadratic activation for
+ * each element of a residual vector, i.e. \f[ \begin{equation} sum^nr_{i=0}
+ * log(1 + \|r_i\|^2 / \alpha) \end{equation} \f] where \f$\alpha\f$ defines the
```
cmastalli, line 25: same here. This seems to not be element-wise, right?

cmastalli:
> SquaredNorm will be much more efficient.

@jcarpent  thanks for the feedback. I have changed to `squaredNorm`

cmastalli (approved):

cmastalli:
@proyan this PR is ready to merge IMHO
Would you like to make a final review?

include/crocoddyl/core/activations/quadratic-flat.hpp
```diff
@@ -0,0 +1,145 @@
...
+   */
+  virtual void calc(const boost::shared_ptr<ActivationDataAbstract> &data, const Eigen::Ref<const VectorXs> &r) {
+    if (static_cast<std::size_t>(r.size()) != nr_) {
+      throw_pretty("Invalid argument: "
+                   << "r has wrong dimension (it should be " + std::to_string(nr_) + ")");
+    }
+    boost::shared_ptr<Data> d = boost::static_pointer_cast<Data>(data);
+
+    d->a0 = exp(-r.squaredNorm() / alpha_);
+    data->a_value = Scalar(1.0) - d->a0;
```
proyan, line 78: The implementation is done as 1-exp(-||r||^2), while documentation mentions 1-exp(||r||^2)

include/crocoddyl/core/activations/quadratic-flat.hpp
```diff
@@ -0,0 +1,145 @@
...
+//
+// Copyright (C) 2020, LAAS-CNRS, University of Edinburgh
+// Copyright note valid unless otherwise stated in individual files.
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_FLAT_HPP_
+#define CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_FLAT_HPP_
+
+#include <iostream>
```
proyan, line 12: We shouldn't need to include iostream and stdexcept (utils/exceptions) is already included.

include/crocoddyl/core/activations/quadratic-log.hpp
```diff
@@ -0,0 +1,143 @@
...
+// Copyright (C) 2020, LAAS-CNRS, University of Edinburgh
+// Copyright note valid unless otherwise stated in individual files.
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_LOG_HPP_
+#define CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_LOG_HPP_
+
+#include <iostream>
+#include <stdexcept>
```
proyan, line 13: We shouldn't need to include iostream and stdexcept (utils/exceptions) is already included.

proyan (changes requested):
Thanks @edantec and @cmastalli for this PR. 
I just have some minor comments, the rest of the PR is okay to merge.

cmastalli:
@edantec you need to fix the format of the code
Please try to run the auto-formatting command in your code routine (before committing a code).

bindings/python/crocoddyl/core/activations/quadratic-flat-exp.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "crocoddyl/core/activations/quadratic-flat-exp.hpp"
+#include "python/crocoddyl/core/activation-base.hpp"
+#include "python/crocoddyl/core/core.hpp"
+
+namespace crocoddyl {
+namespace python {
+
+void exposeActivationQuadFlatExp() {
+  bp::class_<ActivationModelQuadFlatExp, bp::bases<ActivationModelAbstract>>(
```
cmastalli, line 17: This code doesn't support c++98, i.e. we need to change `>>` by `> >`

bindings/python/crocoddyl/core/activations/quadratic-flat-log.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "crocoddyl/core/activations/quadratic-flat-log.hpp"
+#include "python/crocoddyl/core/activation-base.hpp"
+#include "python/crocoddyl/core/core.hpp"
+
+namespace crocoddyl {
+namespace python {
+
+void exposeActivationQuadFlatLog() {
+  bp::class_<ActivationModelQuadFlatLog, bp::bases<ActivationModelAbstract>>(
```
cmastalli, line 17: same here

bindings/python/crocoddyl/core/activations/quadratic-flat-exp.cpp
```diff
@@ -0,0 +1,49 @@
...
+           "Note that the Hessian is constant, so we don't write again this "
+           "value.\n"
+           ":param data: activation data\n"
+           ":param r: residual vector \n")
+      .def("createData", &ActivationModelQuadFlatExp::createData,
+           bp::args("self"), "Create the quadratic flat activation data.\n\n")
+      .add_property(
+          "alpha",
+          bp::make_function(&ActivationModelQuadFlatExp::get_alpha,
+                            bp::return_value_policy<bp::return_by_value>()),
```
cmastalli, line 44: Let's remove `bp::return_value_policy<bp::return_by_value>()` since the function should return a `double` (rather than a `const double&`)

bindings/python/crocoddyl/core/activations/quadratic-flat-log.cpp
```diff
@@ -0,0 +1,49 @@
...
+           "Note that the Hessian is constant, so we don't write again this "
+           "value.\n"
+           ":param data: activation data\n"
+           ":param r: residual vector \n")
+      .def("createData", &ActivationModelQuadFlatLog::createData,
+           bp::args("self"), "Create the quadratic flat activation data.\n")
+      .add_property(
+          "alpha",
+          bp::make_function(&ActivationModelQuadFlatLog::get_alpha,
+                            bp::return_value_policy<bp::return_by_value>()),
```
cmastalli, line 44: same here

include/crocoddyl/core/activations/quadratic-flat-exp.hpp
```diff
@@ -0,0 +1,156 @@
...
+//
+// Copyright (C) 2020, LAAS-CNRS, University of Edinburgh
+// Copyright note valid unless otherwise stated in individual files.
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_FLAT_EXP_HPP_
+#define CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_FLAT_EXP_HPP_
+
+#include "crocoddyl/core/activation-base.hpp"
```
cmastalli, line 12: The order of the header is not good. This is the proper order
```cpp
#include <iostream>
#include <stdexcept>
#include "crocoddyl/core/fwd.hpp"
 #include "crocoddyl/core/activation-base.hpp" 
```

It is important to include first the system headers, and also the `fwd`

include/crocoddyl/core/activations/quadratic-flat-exp.hpp
```diff
@@ -0,0 +1,156 @@
...
+  /*
+   * @brief Initialize the quadFlatExp activation model
+   *
+   * The default `alpha` value is defined as 1.
+   *
+   * @param[in] nr     Dimension of the residual vector
+   * @param[in] alpha  Width of quadratic basin (default: 1.)
+   */
+
+  explicit ActivationModelQuadFlatExpTpl(const std::size_t &nr,
```
cmastalli, line 58: Let's do not use `const reference` for primitives (the primitives are `std::size_t`, `double`, `Scalar` or `bool`.
Thus the code should be modified to
```cpp
   explicit ActivationModelQuadFlatExpTpl(std::size_t nr, Scalar alpha = Scalar(1.))
```

include/crocoddyl/core/activations/quadratic-flat-exp.hpp
```diff
@@ -0,0 +1,156 @@
...
+   */
+  virtual boost::shared_ptr<ActivationDataAbstract> createData() {
+    boost::shared_ptr<ActivationDataAbstract> data =
+        boost::allocate_shared<ActivationDataAbstract>(
+            Eigen::aligned_allocator<ActivationDataAbstract>(), this);
+    data->Arr.diagonal().fill((Scalar)1.);
+    return data;
+  };
+
+  const Scalar &get_alpha() const { return alpha_; };
```
cmastalli, line 121: Same here. Let's change, in both functions, `const Scalar&` to `Scalar`

include/crocoddyl/core/activations/quadratic-flat-log.hpp
```diff
@@ -0,0 +1,154 @@
...
+//
+// Copyright (C) 2020, LAAS-CNRS, University of Edinburgh
+// Copyright note valid unless otherwise stated in individual files.
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_FLAT_LOG_HPP_
+#define CROCODDYL_CORE_ACTIVATIONS_QUADRATIC_FLAT_LOG_HPP_
+
+#include "crocoddyl/core/activation-base.hpp"
```
cmastalli, line 12: Let's reorder the headers as described above

include/crocoddyl/core/activations/quadratic-flat-log.hpp
```diff
@@ -0,0 +1,154 @@
...
+  /*
+   * @brief Initialize the quadFlatLog activation model
+   *
+   * The default `alpha` value is defined as 1.
+   *
+   * @param[in] nr   Dimension of the residual vector
+   * @param[in] alpha  Width of quadratic basin (default: 1.)
+   */
+
+  explicit ActivationModelQuadFlatLogTpl(const std::size_t &nr,
```
cmastalli, line 57: Same here. Let's do not use `const reference` in primitives

include/crocoddyl/core/activations/quadratic-flat-log.hpp
```diff
@@ -0,0 +1,154 @@
...
+   */
+  virtual boost::shared_ptr<ActivationDataAbstract> createData() {
+    boost::shared_ptr<ActivationDataAbstract> data =
+        boost::allocate_shared<ActivationDataAbstract>(
+            Eigen::aligned_allocator<ActivationDataAbstract>(), this);
+    data->Arr.diagonal().fill((Scalar)1.);
+    return data;
+  };
+
+  const Scalar &get_alpha() const { return alpha_; };
```
cmastalli, line 119: same here

cmastalli (changes requested):
Hi @edantec 

I found tiny details that I list as:
 1. Provide compatibility with c++98
 2. Avoid overhead due to the use of `const reference` in primitives
 3. Fix the order of header inclusion

include/crocoddyl/core/activations/quadratic-flat-exp.hpp
jcarpent: You can use const here. It will be helpful to check whether the quantity should be modified or not.

cmastalli: The quantity should be never modified, that should be clear for any developer.
Do you really think we need to care about it?

jcarpent: If you know that the quantities should not be modified, then you should mark it explicitly.
It may also help the compiler to optimize the code accordingly depending on the context.

cmastalli (approved):
@edantec I have already handle all my comments

@jcarpent I follow your suggestion, we are still using `const` in constructors and setting functions

@proyan now this code should pass our CI

cmastalli:
@proyan I have solved your points.
I think it is OK to merge it :)

PR #893, opened 2020-12-06
Separate increasing and decreasing regularization factors

cmastalli:
In our solvers, we use a Levenberg–Marquardt regularization that increases their robustness thanks to the possibility of moving from Gauss-Newton to steepest descent direction.
Furthermore, we have implemented and adaptive scheme--as recommended in the literature--which increases or decreases the damping parameter. In this approach, we use a single factor value for increasing and decreasing the damping.
However, it is often useful to define different factors for increasing or decreasing the damping values. This has been deeply studied in the literature, e.g. https://aip.scitation.org/doi/pdf/10.1063/1.168600. So, the goal of this PR is to allow our users to define both factors.

Furthermore, I noticed that some of the scope variables of the DDP solver do not follow the naming convention. So, I have fixed them.

Note that, for all the changes, I have properly deprecated the Python / c++ code!


cmastalli:
As a side note, I have tried two different suggested factors in wikipedia (also in the attached paper) across all the examples, i.e.
  1. Increase / Decrease factors: 2 / 3
  2. Increase / Decrease factors: 1.5 / 5

These two cases do not prove to be more effective that our current values (i.e., 10 / 10). However, this is not a final conclusion since there are still cases where it reduces the number of iterations.

src/core/solvers/ddp.cpp
```diff
@@ -402,11 +403,19 @@ void SolverDDP::allocateData() {
   fTVxx_p_ = Eigen::VectorXd::Zero(ndx);
 }
 
-const double& SolverDDP::get_regfactor() const { return regfactor_; }
+const double& SolverDDP::get_reg_incfactor() const { return reg_incfactor_; }
```
jcarpent, line 406: There is no reason to return a cont ref here for a double.

src/core/solvers/ddp.cpp
```diff
@@ -14,9 +14,10 @@ namespace crocoddyl {
 
 SolverDDP::SolverDDP(boost::shared_ptr<ShootingProblem> problem)
     : SolverAbstract(problem),
-      regfactor_(10.),
-      regmin_(1e-9),
-      regmax_(1e9),
+      reg_incfactor_(10.),
+      reg_decfactor_(10.),
+      reg_min_(1e-9),
```
jcarpent, line 19: Why should it be so high? This makes totally no sense. You should be able to go to zero.

jcarpent (changes requested):

src/core/solvers/ddp.cpp
jcarpent: This applies everywhere!

src/core/solvers/ddp.cpp
cmastalli: I agree with you, reg_min should be zero.
I was planning to change it in another PR to make the review easier and focus.

However, I will change it since you agree with me ;)

src/core/solvers/ddp.cpp
cmastalli: Could explain me the reason of this briefly?
What about using `const double&` or `const bool&` as in/out arguments of a function?

jcarpent: You have a larger overhead when using const ref for small objects such as bool or double. There is no reason to not return them by copy, as the memory will lower or equal.

src/core/solvers/ddp.cpp
cmastalli: done! 
I used `1e-18` since it is the double resolution.
Additionally, I updated the log files since the numerical behavior changed a bit

src/core/solvers/ddp.cpp
cmastalli: Done for all the solvers and shooting problem!

proyan:
@cmastalli please let me know when it is ready for review

src/core/solvers/ddp.cpp
wxmerkt: You can use `std::numeric_limits<double>::epsilon()` (from `<limits>`) or a quantity related to that (e.g. when you square/take the square root). For double, this is `2.22045e-16`.

src/core/solvers/ddp.cpp
```diff
@@ -16,7 +16,7 @@ SolverDDP::SolverDDP(boost::shared_ptr<ShootingProblem> problem)
     : SolverAbstract(problem),
       reg_incfactor_(10.),
       reg_decfactor_(10.),
-      reg_min_(1e-9),
+      reg_min_(1e-18),
```
jcarpent, line 19: This should be ZERO.

jcarpent (dismissed):

src/core/solvers/ddp.cpp
cmastalli: A zero value is not efficient because it makes our algorithm to decrease the damping factor to a level that doesn't influence the direction.
Furthermore, it adds unnecessarily iterations to increase the damping factor if the algorithm is recovering for very low-still not useful-value.

cmastalli: I believe this topic merits a proper analysis. But this is very hard and out of the scope of this PR.

In any case, thank you for the feedback!

jcarpent: I'm a bit afraid of your answer. I don't think you realize what you are writing so far.

If you want to solve a strictly feasible LQR problem, you don't need any regularization term at ALL.
While with your value, you are shifting your problem and then the solution!

In addition, `1e-18` makes totally no sense when thinking in terms of double arithmetic. 

Please @cmastalli don't escape from other contributors' remarks with very unfounded answers !!! It is very painful in the end !!!

cmastalli:
@proyan you could go through the review process.
Unfortunately, I had to modify many other places, which makes this PR a bit fatty.
This is the list of extra changes:
  - Removed extra overhead produced by using `const reference` for primitives (i.e., `double`, `bool`, `std::size_t` and `Scalar`).
  - Updated robot loading methods in unittest code
  - Defined the same values for `regMin`, `regMax` in KKT solver and Python derived solvers (it fixes the unittest)
  - Enabled the use of callback functions in Python derived solvers

src/core/solvers/ddp.cpp
cmastalli: @jcarpent I appreciate your feedback but this is not the right way of working.
Let me explain you why I believe this is not the right way of working.

This PR consists on adding the possibility of having two factors for the regularization.
Then, you asked two things which are out of that scope:
  1. Remove const reference for primitives
  2. Change a value that historical was there

Reviewing other parts of the code and asking to change then is not productive, and prone error as well.
It is not productive because the developer can't be focused on the PR itself!
For me, it is useful is you write an issue, and then I could allocate time for making a careful work.

However, for this case, I follow both suggestions.
You need to be aware that the point 2 is a delicate changed.
I have designed that value based on running our examples.
I encountered that having zero increases the number of iterations in a couple of problems.
As you might see, I don't have time to report each result for a requested that it is out of the scope.

cmastalli: I don't escape for discussions that have a clear solution.
But to keep me productivity, I will do it if the discussion is going anywhere!

cmastalli: Btw, the value is `1e-16`. I just forgot to push it.

cmastalli: Let me know if this solves your point.

And again, note that we have a routine that decreases the damping value. There is no sense to decrease it more than `1e-16`

src/core/solvers/ddp.cpp
proyan: I agree that we should start this value from zero. It doesn't make any sense to use such ridiculously low value. 

For the reg_max, I don't understand why you decided to increase it to 1e12? 1e9 is already quite high, and if problem cannot be solved with 1e9 as the regularization, there is something wrong with the problem

proyan: There is a parallel discussion below on the same topic. I'll duplicate my comment there instead.

src/core/solvers/ddp.cpp
proyan: 1) I agree that we should start this value from zero. It doesn't make any sense to use such ridiculously low value (putting 1e-18 or 1e-16 is non-sense).

2) For the reg_max, I don't understand why you decided to increase it to 1e12? 1e9 is already quite high, and if problem cannot be solved with 1e9 as the regularization, there is something wrong with the problem. 
1e9 is already an arbitrary choice, and so is 1e12. Perhaps it would be worth setting the max reg factor as a value relative 
to the problem Quu? But that would be too complicated. 


Regarding the scope of this PR:
1) Regularization is meant to help an improperly formed shooting problem. However, currently we are helping every problem, regardless of whether it is needed or not. So this was a problem in the initial design, which we have a chance of fixing now. So thanks @cmastalli for taking care of that.

2) There is already a bunch of mixed stuff in this PR. So it feels normal that when perusing the files, you can spot issues with existing code. You are within your rights @cmastalli to defer suggestions to a future PR (while keeping track of the suggestions). But once you have accepted and implemented the suggestion, the discussion has to be complete and the implementation proper. Changing 1e-9 to 1e-16 is not the solution of Justin's suggestion.

include/crocoddyl/multibody/numdiff/contact.hpp
```diff
@@ -73,16 +73,16 @@ class ContactModelNumDiffTpl : public ContactModelAbstractTpl<_Scalar> {
   /**
    * @brief Get the disturbance_ object
    *
-   * @return const Scalar&
+   * @return Scalar
    */
-  const Scalar& get_disturbance() const;
+  Scalar get_disturbance() const;
```
proyan, line 78: here

include/crocoddyl/multibody/numdiff/contact.hxx
```diff
@@ -79,12 +79,12 @@ const boost::shared_ptr<ContactModelAbstractTpl<Scalar> >& ContactModelNumDiffTp
 }
 
 template <typename Scalar>
-const Scalar& ContactModelNumDiffTpl<Scalar>::get_disturbance() const {
+Scalar ContactModelNumDiffTpl<Scalar>::get_disturbance() const {
```
proyan, line 82: here

include/crocoddyl/multibody/wrench-cone.hpp
```diff
@@ -113,22 +113,22 @@ class WrenchConeTpl {
   /**
    * @brief Return friction coefficient
    */
-  const Scalar& get_mu() const;
+  Scalar get_mu() const;
```
proyan, line 116: here

src/core/solvers/box-fddp.cpp
```diff
@@ -33,18 +33,18 @@ SolverBoxFDDP::~SolverBoxFDDP() {}
...
   for (std::size_t t = 0; t < T; ++t) {
     Quu_inv_[t] = Eigen::MatrixXd::Zero(nu, nu);
   }
   du_lb_.resize(nu);
   du_ub_.resize(nu);
 }
 
-void SolverBoxFDDP::computeGains(const std::size_t& t) {
-  const std::size_t& nu = problem_->get_runningModels()[t]->get_nu();
+void SolverBoxFDDP::computeGains(std::size_t t) {
```
proyan, line 46: This should be `void SolverBoxFDDP::computeGains(const std::size_t t)`

src/core/solvers/box-fddp.cpp
```diff
@@ -76,21 +76,21 @@ void SolverBoxFDDP::computeGains(const std::size_t& t) {
   }
 }
 
-void SolverBoxFDDP::forwardPass(const double& steplength) {
+void SolverBoxFDDP::forwardPass(double steplength) {
```
proyan, line 79: same here. use `const double`

Even if the `const` might not be useful for built-in types, `const` still improves the readability of the code.

src/core/solvers/box-qp.cpp
```diff
@@ -12,8 +12,7 @@
 
 namespace crocoddyl {
 
-BoxQP::BoxQP(const std::size_t nx, std::size_t maxiter, const double th_acceptstep, const double th_grad,
-             const double reg)
+BoxQP::BoxQP(std::size_t nx, std::size_t maxiter, double th_acceptstep, double th_grad, double reg)
```
proyan, line 15: same here

src/core/solvers/ddp.cpp
```diff
@@ -14,9 +14,10 @@ namespace crocoddyl {
 
 SolverDDP::SolverDDP(boost::shared_ptr<ShootingProblem> problem)
     : SolverAbstract(problem),
-      regfactor_(10.),
-      regmin_(1e-9),
-      regmax_(1e9),
+      reg_incfactor_(10.),
+      reg_decfactor_(10.),
+      reg_min_(1e-16),
+      reg_max_(1e12),
```
proyan, line 20: This part should follow the changes from https://github.com/loco-3d/crocoddyl/pull/893#discussion_r542122288

src/core/solvers/ddp.cpp
```diff
@@ -40,13 +41,13 @@ SolverDDP::SolverDDP(boost::shared_ptr<ShootingProblem> problem)
 SolverDDP::~SolverDDP() {}
 
 bool SolverDDP::solve(const std::vector<Eigen::VectorXd>& init_xs, const std::vector<Eigen::VectorXd>& init_us,
-                      const std::size_t& maxiter, const bool& is_feasible, const double& reginit) {
+                      std::size_t maxiter, bool is_feasible, double reginit) {
```
proyan, line 44: use of const here

src/core/solvers/ddp.cpp
```diff
@@ -40,13 +41,13 @@ SolverDDP::SolverDDP(boost::shared_ptr<ShootingProblem> problem)
...
   xs_try_[0] = problem_->get_x0();  // it is needed in case that init_xs[0] is infeasible
   setCandidate(init_xs, init_us, is_feasible);
 
   if (std::isnan(reginit)) {
-    xreg_ = regmin_;
-    ureg_ = regmin_;
+    xreg_ = reg_min_;
+    ureg_ = reg_min_;
```
proyan, line 50: I remember seeing regInit=1e-9 in many solvers. The logic that we have in this file should be copied to the other solvers (I think it was in fddp that this default value is set, among others)

src/core/solvers/ddp.cpp
```diff
@@ -118,24 +119,24 @@ bool SolverDDP::solve(const std::vector<Eigen::VectorXd>& init_xs, const std::ve
   return false;
 }
 
-void SolverDDP::computeDirection(const bool& recalcDiff) {
+void SolverDDP::computeDirection(bool recalcDiff) {
```
proyan, line 122: const here

src/core/solvers/ddp.cpp
```diff
@@ -118,24 +119,24 @@ bool SolverDDP::solve(const std::vector<Eigen::VectorXd>& init_xs, const std::ve
...
   if (recalcDiff) {
     calcDiff();
   }
   backwardPass();
 }
 
-double SolverDDP::tryStep(const double& steplength) {
+double SolverDDP::tryStep(double steplength) {
```
proyan, line 129: const here

src/core/solvers/ddp.cpp
```diff
@@ -438,28 +447,61 @@ const std::vector<Eigen::VectorXd>& SolverDDP::get_k() const { return k_; }
...
+void SolverDDP::set_reg_incfactor(double regfactor) {
+  if (regfactor <= 1.) {
+    throw_pretty("Invalid argument: "
+                 << "reg_incfactor value is higher than 1.");
+  }
+  reg_incfactor_ = regfactor;
+}
+
+void SolverDDP::set_reg_decfactor(double regfactor) {
+  if (regfactor <= 1.) {
```
proyan, line 459: 0. < regfactor <=1.

src/core/solvers/ddp.cpp
```diff
@@ -438,28 +447,61 @@ const std::vector<Eigen::VectorXd>& SolverDDP::get_k() const { return k_; }
 
 const std::vector<Eigen::VectorXd>& SolverDDP::get_fs() const { return fs_; }
 
-void SolverDDP::set_regfactor(const double& regfactor) {
+void SolverDDP::set_reg_incfactor(double regfactor) {
```
proyan, line 450: const

src/core/solvers/ddp.cpp
```diff
@@ -438,28 +447,61 @@ const std::vector<Eigen::VectorXd>& SolverDDP::get_k() const { return k_; }
...
+  if (regfactor <= 1.) {
+    throw_pretty("Invalid argument: "
+                 << "reg_incfactor value is higher than 1.");
+  }
+  reg_incfactor_ = regfactor;
+}
+
+void SolverDDP::set_reg_decfactor(double regfactor) {
```
proyan, line 458: const

src/core/solvers/ddp.cpp
```diff
@@ -438,28 +447,61 @@ const std::vector<Eigen::VectorXd>& SolverDDP::get_k() const { return k_; }
...
+  if (regfactor <= 1.) {
+    throw_pretty("Invalid argument: "
+                 << "reg_decfactor value is higher than 1.");
+  }
+  reg_decfactor_ = regfactor;
+}
+
+void SolverDDP::set_regfactor(double regfactor) {
```
proyan, line 466: const

src/core/solvers/ddp.cpp
```diff
@@ -438,28 +447,61 @@ const std::vector<Eigen::VectorXd>& SolverDDP::get_k() const { return k_; }
...
   if (regfactor <= 1.) {
     throw_pretty("Invalid argument: "
                  << "regfactor value is higher than 1.");
   }
-  regfactor_ = regfactor;
+  reg_incfactor_ = regfactor;
+  reg_decfactor_ = regfactor;
+}
+
+void SolverDDP::set_reg_min(double regmin) {
```
proyan, line 475: const

src/core/solvers/ddp.cpp
proyan: Since this is deprecated, please call set_reg_decfactor and set_reg_incfactor from this function instead

src/core/solvers/ddp.cpp
```diff
@@ -438,28 +447,61 @@ const std::vector<Eigen::VectorXd>& SolverDDP::get_k() const { return k_; }
...
+  if (0. > regmin) {
+    throw_pretty("Invalid argument: "
+                 << "regmin value has to be positive.");
+  }
+  reg_min_ = regmin;
 }
 
-void SolverDDP::set_regmin(const double& regmin) {
+void SolverDDP::set_regmin(double regmin) {
```
proyan, line 483: same here. please call the non-deprecated function here instead.

src/core/solvers/ddp.cpp
```diff
@@ -438,28 +447,61 @@ const std::vector<Eigen::VectorXd>& SolverDDP::get_k() const { return k_; }
...
   if (0. > regmin) {
```
proyan, line 484: should be changed following https://github.com/loco-3d/crocoddyl/pull/893#discussion_r542122288

src/core/solvers/ddp.cpp
```diff
@@ -438,28 +447,61 @@ const std::vector<Eigen::VectorXd>& SolverDDP::get_k() const { return k_; }
...
+void SolverDDP::set_reg_max(double regmax) {
+  if (0. > regmax) {
+    throw_pretty("Invalid argument: "
+                 << "regmax value has to be positive.");
+  }
+  reg_max_ = regmax;
 }
 
-void SolverDDP::set_regmax(const double& regmax) {
+void SolverDDP::set_regmax(double regmax) {
```
proyan, line 499: call non-depreccated function here.

src/core/solvers/ddp.cpp
```diff
@@ -438,28 +447,61 @@ const std::vector<Eigen::VectorXd>& SolverDDP::get_k() const { return k_; }
...
+void SolverDDP::set_regmin(double regmin) {
   if (0. > regmin) {
     throw_pretty("Invalid argument: "
                  << "regmin value has to be positive.");
   }
-  regmin_ = regmin;
+  reg_min_ = regmin;
+}
+
+void SolverDDP::set_reg_max(double regmax) {
```
proyan, line 491: const

src/core/solvers/ddp.cpp
```diff
@@ -482,31 +524,31 @@ void SolverDDP::set_alphas(const std::vector<double>& alphas) {
   alphas_ = alphas;
 }
 
-void SolverDDP::set_th_stepdec(const double& th_stepdec) {
+void SolverDDP::set_th_stepdec(double th_stepdec) {
```
proyan, line 527: const

src/core/solvers/ddp.cpp
```diff
@@ -482,31 +524,31 @@ void SolverDDP::set_alphas(const std::vector<double>& alphas) {
...
   if (0. >= th_stepdec || th_stepdec > 1.) {
     throw_pretty("Invalid argument: "
                  << "th_stepdec value should between 0 and 1.");
   }
   th_stepdec_ = th_stepdec;
 }
 
-void SolverDDP::set_th_stepinc(const double& th_stepinc) {
+void SolverDDP::set_th_stepinc(double th_stepinc) {
```
proyan, line 535: const

src/core/solvers/ddp.cpp
```diff
@@ -482,31 +524,31 @@ void SolverDDP::set_alphas(const std::vector<double>& alphas) {
...
   if (0. >= th_stepinc || th_stepinc > 1.) {
     throw_pretty("Invalid argument: "
                  << "th_stepinc value should between 0 and 1.");
   }
   th_stepinc_ = th_stepinc;
 }
 
-void SolverDDP::set_th_grad(const double& th_grad) {
+void SolverDDP::set_th_grad(double th_grad) {
```
proyan, line 543: const

src/core/solvers/ddp.cpp
```diff
@@ -482,31 +524,31 @@ void SolverDDP::set_alphas(const std::vector<double>& alphas) {
...
   if (0. > th_grad) {
     throw_pretty("Invalid argument: "
                  << "th_grad value has to be positive.");
   }
   th_grad_ = th_grad;
 }
 
-void SolverDDP::set_th_gaptol(const double& th_gaptol) {
+void SolverDDP::set_th_gaptol(double th_gaptol) {
```
proyan, line 551: const

src/core/solvers/fddp.cpp
```diff
@@ -229,7 +229,7 @@ void SolverFDDP::forwardPass(const double& steplength) {
 
 double SolverFDDP::get_th_acceptnegstep() const { return th_acceptnegstep_; }
 
-void SolverFDDP::set_th_acceptnegstep(const double& th_acceptnegstep) {
+void SolverFDDP::set_th_acceptnegstep(double th_acceptnegstep) {
```
proyan, line 232: I'll avoid putting the comment about const parameters for the rest of the PR.

src/core/solvers/kkt.cpp
```diff
@@ -13,26 +13,27 @@ namespace crocoddyl {
 
 SolverKKT::SolverKKT(boost::shared_ptr<ShootingProblem> problem)
     : SolverAbstract(problem),
-      regfactor_(10.),
-      regmin_(1e-9),
-      regmax_(1e9),
+      reg_incfactor_(10.),
+      reg_decfactor_(10.),
+      reg_min_(1e-16),
+      reg_max_(1e12),
```
proyan, line 19: follows https://github.com/loco-3d/crocoddyl/pull/893#discussion_r542122288

proyan (changes requested):

include/crocoddyl/multibody/friction-cone.hpp
```diff
@@ -29,30 +29,29 @@ class FrictionConeTpl {
...
-  const Scalar& get_mu() const;
-  const std::size_t& get_nf() const;
-  const bool& get_inner_appr() const;
-  const Scalar& get_min_nforce() const;
-  const Scalar& get_max_nforce() const;
+  Scalar get_mu() const;
+  std::size_t get_nf() const;
+  bool get_inner_appr() const;
+  Scalar get_min_nforce() const;
+  Scalar get_max_nforce() const;
```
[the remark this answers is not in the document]
cmastalli: I guess you referred only to
```c++
  Scalar get_min_nforce() const;
  Scalar get_max_nforce() const;
```
because the other cases are superfluous.
Indeed, if you add those `const` keyworks you will get a warning message.
See below for a more elaborated explanation:
https://stackoverflow.com/questions/1134237/pedantic-gcc-warning-type-qualifiers-on-function-return-type

src/core/solvers/ddp.cpp
cmastalli: Hi @proyan,

I don't agree with the suggestions in this thread.
I would explain you why, but if this doesn't yet convinced you then we need to discuss it in private.

>     1. I agree that we should start this value from zero. It doesn't make any sense to use such ridiculously low value (putting 1e-18 or 1e-16 is non-sense).

No, it is not a good idea to define the minimum regularization (or to start the algorithm) with a zero value.
First, values lower than `1e-16` (i.e., the double resolution) do not really affect the direction or do not produce any numerical effect. I have also tested this well-known fact across our problems. Please believe me or try yourself.
Second, if you use define the minimum regulation value to zero, you are allowing our algorithm to decrease superfluously  the regularization value. Furthermore, it would take useless steps to recover from it, i.e., to increase bigger than `1e-16`.
I have also validated this statement across our problems. Indeed, this suggestion increases the number of iteration in some cases.

Note that we have implemented a Levenberg-Marquardt algorithm. It is common across LM implementation to have a minimum regularization values different to zero. Indeed, this is clearly stated in the original paper that proposed the LM strategy I copied for you the text below:
image
 PDF: https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwiM3OHei7ztAhVhRhUIHXu1DcEQFjAAegQIARAC&url=https%3A%2F%2Fepubs.stfc.ac.uk%2Fmanifestation%2F6683%2FAERE_R_6799.pdf&usg=AOvVaw19LGQ4NFbIsBebeSX7BgPg

---

>     2\. For the reg_max, I don't understand why you decided to increase it to 1e12? 1e9 is already quite high, and if problem cannot be solved with 1e9 as the regularization, there is something wrong with the problem.
>         1e9 is already an arbitrary choice, and so is 1e12. Perhaps it would be worth setting the max reg factor as a value relative
>         to the problem Quu? But that would be too complicated.
This number indeed needs to be studied more deeply. I just found that in some very hard problems is good to allow a bigger value.
Your proposition can be done for a specific problem (at the user level). With a bigger value, I am jusrt making the default configuration more robust.

I hope this follow-up explanation can satisfy you. Otherwise, you really need to make contra-argument with evidence. It is not fair to insist without any support of your statement. I am more than open to get feedback but with evidence!


**Note**
In near future, I will conduct another study to potentially improve a bit more the regularization. For instance, it might be useful to always start with Gauss-Newton direction as KNITRO performs. Note that we always start with the stored regularization value. I have been spending quite a lot of my time in these tiny details, please do not think that I am taking this lightly.

bindings/python/crocoddyl/core/solvers/ddp.cpp
```diff
@@ -97,32 +98,39 @@ void exposeSolverDDP() {
...
+                    bp::make_function(&SolverDDP::get_reg_incfactor,
+                                      deprecated<>("Deprecated. Use reg_incfactor or reg_decfactor")),
+                    bp::make_function(&SolverDDP::set_reg_incfactor,
+                                      deprecated<>("Deprecated. Use reg_incfactor or reg_decfactor")),
+                    "regularization factor used for increasing or decreasing the damping value.")
+      .add_property("reg_min", bp::make_function(&SolverDDP::get_reg_min), bp::make_function(&SolverDDP::set_reg_min),
+                    "minimum regularization value.")
+      .add_property("reg_max", bp::make_function(&SolverDDP::get_reg_max), bp::make_function(&SolverDDP::set_reg_max),
+                    "maximum regularization value.")
+      .add_property("regMin", bp::make_function(&SolverDDP::get_reg_min, deprecated<>("Deprecated. Use reg_min")),
```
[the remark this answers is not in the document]
cmastalli: The idea is to follow the same convention used for `th_XXX` variables. I recall a few examples below:
  1. `th_stepDec`
  2. `th_stepInc`
  3. `th_grad`

I consider this conversation as resolved. Please open it if there is something else to discuss :)

bindings/python/crocoddyl/utils/__init__.py
```diff
@@ -1010,11 +1010,13 @@ def __init__(self, shootingProblem):
...
+        self.callbacks = None
         self.x_reg = 0
         self.u_reg = 0
-        self.regFactor = 10
-        self.regMax = 1e9
-        self.regMin = 1e-9
+        self.regIncFactor = 10
+        self.regDecFactor = 10
+        self.regMax = 1e12
+        self.regMin = 1e-16
```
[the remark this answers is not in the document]
cmastalli: > Also, if regMin, regMax, regIncFactor etc are deprecated, they should be changed here

Indeed, I just renamed to `reg_min`, `reg_max`, `reg_incFactor`, and `reg_decFactor` as done in the bindings. Note that this code is for unit-testing thus it is not strictly important to match the naming convention.
However, it is good for readiness purpose.





> I couldn't find where you had defined regIncFactor. Could you please point this out?

Just a few lines before, see here: https://github.com/cmastalli/crocoddyl/blob/topic/ddp-params/bindings/python/crocoddyl/utils/__init__.py#L1016-L1017

bindings/python/crocoddyl/utils/__init__.py
```diff
@@ -1047,8 +1049,8 @@ def tryStep(self, stepLength=1):
 
     def solve(self, init_xs=[], init_us=[], maxiter=100, isFeasible=False, regInit=None):
         self.setCandidate(init_xs, init_us, isFeasible)
-        self.x_reg = regInit if regInit is not None else self.regMin
-        self.u_reg = regInit if regInit is not None else self.regMin
+        self.x_reg = regInit if regInit is not None else 1e-9
+        self.u_reg = regInit if regInit is not None else 1e-9
```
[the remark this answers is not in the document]
cmastalli: This indeed is not a change. The previous value was `1e-9`, but now the minimum regularization value is not `1e-9`.
So, the short answer is `I didn't want to change this, otherwise, it will affect the log files`.

But actually, there is a better explanation.
Using the minimum regularization implies
   `the default regularization believes that starting from a Gauss-Newton direction is better`.
Instead, using `1e-9` implies that
   `the default regularization assumes that it is more robust to start somewhere between Gauss-Newton to steepest descent, however, it is good to be closed to Gauss-Newton one`.

This choice was designed by Nicolas, however, I am inclined to keep this strategy.

Other comments, note that it is important to keep the same values in these Python derived classes and c++ code for unit-testing purpose.

I consider this conversation as resolved, however, feel free to open it if this need further discussion.

bindings/python/crocoddyl/utils/__init__.py
```diff
@@ -1217,8 +1218,8 @@ def __init__(self, shootingProblem):
...
```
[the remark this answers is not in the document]
cmastalli: I have answered this question above.

include/crocoddyl/core/solver-base.hpp
```diff
@@ -76,8 +76,8 @@ class SolverAbstract {
    * @return A boolean that describes if convergence was reached.
    */
   virtual bool solve(const std::vector<Eigen::VectorXd>& init_xs = DEFAULT_VECTOR,
-                     const std::vector<Eigen::VectorXd>& init_us = DEFAULT_VECTOR, const std::size_t& maxiter = 100,
-                     const bool& is_feasible = false, const double& reg_init = 1e-9) = 0;
+                     const std::vector<Eigen::VectorXd>& init_us = DEFAULT_VECTOR, const std::size_t maxiter = 100,
+                     const bool is_feasible = false, const double reg_init = 1e-9) = 0;
```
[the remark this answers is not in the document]
cmastalli: As explained before, I wanted to keep the same initial regularization value. But it is also important to define robustness rather than performance. I also explained above that this design choice was done by Nicolas, and I am inclined to this strategy as well.

As in other comments, I will close this discussion as I consider done. Do not hesitate to re-open it.

include/crocoddyl/core/solvers/box-qp.hpp
```diff
@@ -89,8 +88,8 @@ class BoxQP {
    * @param[in] th_grad        Gradient tolerance threshold (default 1e-9)
    * @param[in] reg            Regularization value (default 1e-9)
    */
-  BoxQP(const std::size_t nx, std::size_t maxiter = 100, const double th_acceptstep = 0.1, const double th_grad = 1e-9,
-        const double reg = 1e-9);
+  BoxQP(const std::size_t nx, const std::size_t maxiter = 100, const double th_acceptstep = 0.1,
+        const double th_grad = 1e-9, const double reg = 1e-9);
```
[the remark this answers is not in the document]
cmastalli: same explanation for the Box-QP solver :)

include/crocoddyl/core/solvers/ddp.hpp
```diff
@@ -58,10 +58,10 @@ class SolverDDP : public SolverAbstract {
   virtual ~SolverDDP();
 
   virtual bool solve(const std::vector<Eigen::VectorXd>& init_xs = DEFAULT_VECTOR,
-                     const std::vector<Eigen::VectorXd>& init_us = DEFAULT_VECTOR, const std::size_t& maxiter = 100,
-                     const bool& is_feasible = false, const double& regInit = 1e-9);
-  virtual void computeDirection(const bool& recalc = true);
-  virtual double tryStep(const double& steplength = 1);
+                     const std::vector<Eigen::VectorXd>& init_us = DEFAULT_VECTOR, const std::size_t maxiter = 100,
+                     const bool is_feasible = false, const double regInit = 1e-9);
```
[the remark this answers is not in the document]
cmastalli: and here as well

include/crocoddyl/core/solvers/ddp.hpp
```diff
@@ -150,20 +150,32 @@ class SolverDDP : public SolverAbstract {
    */
   virtual void allocateData();
 
+  /**
+   * @brief Return the regularization factor used to increase the damping value
+   */
+  double get_reg_incfactor() const;
```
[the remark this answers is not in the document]
cmastalli: This is consistent with our standard for `reg_XXX` as explained above. You should notice that we do a bit differently between c++ and Python
 1. (c++) `th_stepdec`  /   (Python) `th_stepDec`
 2. (c++) `th_stepinc`  /   (Python) `th_stepInc`

You might disagree with having difference between c++ and Python.
If so, we could handle them but this will imply to change it everywhere and add a lot of deprecation functions.

I would avoid that, but you're free to tackle it in another PR.
So, I consider this discussion as closed. Please create an issue if you want to handle this.

include/crocoddyl/core/solvers/ddp.hpp
```diff
@@ -240,20 +252,32 @@ class SolverDDP : public SolverAbstract {
...
 
+  /**
+   * @brief Modify the regularization factor used to increase the damping value
+   */
+  void set_reg_incfactor(const double reg_factor);
+
+  /**
+   * @brief Modify the regularization factor used to decrease the damping value
+   */
+  void set_reg_decfactor(const double reg_factor);
```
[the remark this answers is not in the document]
cmastalli: it follows our standard as explained above :)

include/crocoddyl/core/solvers/ddp.hpp
```diff
@@ -240,20 +252,32 @@ class SolverDDP : public SolverAbstract {
...
-  void set_regmin(const double& regmin);
+  void set_reg_min(const double regmin);
+  DEPRECATED("Use set_reg_min()", void set_regmin(const double regmin));
 
   /**
    * @brief Modify the maximum regularization value
    */
-  void set_regmax(const double& regmax);
+  void set_reg_max(const double regmax);
+  DEPRECATED("Use set_reg_max()", void set_regmax(const double regmax));
```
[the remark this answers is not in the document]
cmastalli: I have answered this question above. Please write here if you don't manage to find it.

In the meantime, I consider this discussion closed as you explained in other comments.

include/crocoddyl/core/solvers/fddp.hpp
```diff
@@ -58,8 +58,8 @@ class SolverFDDP : public SolverDDP {
   virtual ~SolverFDDP();
 
   virtual bool solve(const std::vector<Eigen::VectorXd>& init_xs = DEFAULT_VECTOR,
-                     const std::vector<Eigen::VectorXd>& init_us = DEFAULT_VECTOR, const std::size_t& maxiter = 100,
-                     const bool& is_feasible = false, const double& regInit = 1e-9);
+                     const std::vector<Eigen::VectorXd>& init_us = DEFAULT_VECTOR, const std::size_t maxiter = 100,
+                     const bool is_feasible = false, const double regInit = 1e-9);
```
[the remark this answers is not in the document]
cmastalli: already answered above :)

bindings/python/crocoddyl/utils/__init__.py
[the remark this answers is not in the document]
cmastalli: I resolved the difference between the Python bindings.
So I believe this discussion is resolved :)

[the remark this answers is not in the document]
cmastalli: I forgot to mention that I have provided the answer to https://github.com/loco-3d/crocoddyl/pull/893#discussion_r542122288.
It explains why that suggestion is not a good idea.

src/core/solvers/kkt.cpp
cmastalli: I already answer this above.
Let's discuss there. I will keep it open only for my own use

bindings/python/crocoddyl/utils/__init__.py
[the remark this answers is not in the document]
cmastalli: Let's discuss above. I will keep it open only for my own use

cmastalli:
Thanks @proyan for realizing the missing `const` keywords in few functions.

Unfortunately, Justin and you are committing a mistake with the statement regarding the minimum regularization value.
I have conducted experiments, provided explanations, cited reference that sustain the decision I have made in this PR.
I hope we can proceed with it, and to avoid looping discussion with unsupported evidences :)

src/core/solvers/ddp.cpp
cmastalli: No, it has to be strictly equal / bigger than 1.
Note that we use this number to divide the current regularization value.

cmastalli:
@proyan any update from your side.
I need to move on this topic asap (to avoid extreme multitasking)

proyan:
Let's discuss this tomorrow

src/core/solvers/ddp.cpp
cmastalli: I have reverted the values as discussed with @nmansard.
Note that the values were `1e-9` and `1e9` for the `regMin` and `regMax`, respectively.

cmastalli:
@proyan -- This PR is ready to be merged as I reverted the values. Additionally, I did other actions:
  1. Rebased the code and fixed conflict issues
  2. Deprecated the `th_gaptol` in Python bindings (the only missed deprecation)

bindings/python/crocoddyl/utils/__init__.py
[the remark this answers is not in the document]
proyan: Let's replace this value by reg_min. This way, there is only one place where a custom value is defined. :)

include/crocoddyl/core/solver-base.hpp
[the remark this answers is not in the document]
proyan: We should set reg_init to reg_min here as well

include/crocoddyl/core/solvers/ddp.hpp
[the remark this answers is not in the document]
proyan: same here

include/crocoddyl/core/solvers/fddp.hpp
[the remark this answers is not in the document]
proyan: same here

include/crocoddyl/core/solvers/kkt.hpp
```diff
@@ -25,10 +25,10 @@ class SolverKKT : public SolverAbstract {
   virtual ~SolverKKT();
 
   virtual bool solve(const std::vector<Eigen::VectorXd>& init_xs = DEFAULT_VECTOR,
-                     const std::vector<Eigen::VectorXd>& init_us = DEFAULT_VECTOR, const std::size_t& maxiter = 100,
-                     const bool& is_feasible = false, const double& regInit = 1e-9);
-  virtual void computeDirection(const bool& recalc = true);
-  virtual double tryStep(const double& steplength = 1);
+                     const std::vector<Eigen::VectorXd>& init_us = DEFAULT_VECTOR, const std::size_t maxiter = 100,
+                     const bool is_feasible = false, const double regInit = 1e-9);
```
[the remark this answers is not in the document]
proyan: and here

proyan (changes requested):
Thanks a lot for taking care of the changes Carlos.
There are just a few minor modifications that I'll ask you to do, and should not take you more than two minutes :). 

Other than that, this PR is ready to go, and you could merge this PR 👍🏽 
Thanks for the effort.

proyan (changes requested):
Hi Carlos,

Thanks a lot for this effort!! 

I'll just ask you for a few more minor modifications, which should not take you more than 2 minutes to do. Apart from that, this PR looks ready from my side :). You could merge this at will

include/crocoddyl/core/solver-base.hpp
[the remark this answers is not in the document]
cmastalli: > We should set reg_init to reg_min here as well

I don't understand this comment. The value of reg_min is `1e-9` (as it was originally). We cannot neither write a code like this:
```cpp
                     const std::vector<Eigen::VectorXd>& init_us = DEFAULT_VECTOR, const std::size_t maxiter = 100,
                     const bool is_feasible = false, const double reg_init = reg_min_) = 0;
```
as `reg_min_` is a non-static data member.

bindings/python/crocoddyl/utils/__init__.py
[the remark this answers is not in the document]
cmastalli: done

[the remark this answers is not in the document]
proyan: cool thanks!

include/crocoddyl/core/solver-base.hpp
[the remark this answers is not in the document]
proyan: Ah yes. you are right, it has to be set in the Derived classes.

cmastalli:
Thanks @proyan for the review!
Please check the friction cone PR #907 as soon as you can (this PR needs to be merged before the residual one)

Issue #896, opened 2020-12-31
Error about omniORB.CORBA.BAD_PARAM?


ddliu365:
Hi, I tested the solver, and got this error:

```
---------------------------------------------------------------------------
BAD_PARAM                                 Traceback (most recent call last)
<ipython-input-7-8f02414781a8> in <module>()
      2 
      3 rolloverPlanner = DoubleSupportThrowPlanner(x0, nq, nv, na, 100, release_index=0, timeStep=ctrl_time_step, display=True)
----> 4 tauRolloverTraj,throwTime = rolloverPlanner.forward(m)

/home/jack/repos/throw/src/whole_body_planner.pyc in forward(self, model)
     50         gait = SimpleHumanoidGaitProblem(model.model_o, q0, torso, lFoot, rFoot, rGripper)
     51         problem = gait.createDSPrepThrowProblem(self.x0, timeStep, prepKnots, final=False)
---> 52         x_prep = self.solve(model.robot_o, model.model_o, problem, endEffectors)
     53         x_f_prep = x_prep[:,-1]
     54         x_f_prep = x_f_prep.reshape((x_f_prep.shape[0], 1))

/home/jack/repos/throw/src/planner.pyc in solve(self, robot, model, problem, endEffectors)
     59             for m, d in list(zip(self.ddp.problem.runningModels, self.ddp.problem.runningDatas))
     60         ]
---> 61         done = self.ddp.solve(xs, us, 500, False, 0.1)
     62         print(done)
     63         if self.display==True:

/opt/openrobots/lib/python2.7/site-packages/crocoddyl/__init__.pyc in __call__(self, solver)
    351         if (solver.iter + 1) % self.visualization.freq:
    352             return
--> 353         self.visualization.displayFromSolver(solver)
    354 
    355 

/opt/openrobots/lib/python2.7/site-packages/crocoddyl/__init__.pyc in displayFromSolver(self, solver, factor)
     44         models = solver.problem.runningModels.tolist() + [solver.problem.terminalModel]
     45         dts = [m.dt if hasattr(m, "differential") else 0. for m in models]
---> 46         self.display(solver.xs, fs, ps, dts, factor)
     47         if numpy_conversion:
     48             numpy_conversion = False

/opt/openrobots/lib/python2.7/site-packages/crocoddyl/__init__.pyc in display(self, xs, fs, ps, dts, factor)
[...]

PR #897, opened 2021-01-04
fix_corba_bad_param_issue

ddliu365:
@cmastalli 
This is a pr to address [issue](https://github.com/loco-3d/crocoddyl/issues/896#issue-777084374) #896

### DIFF
--- a/bindings/python/crocoddyl/__init__.py
+++ b/bindings/python/crocoddyl/__init__.py
@@ -310,13 +310,13 @@ class GepettoDisplay(DisplayAbstract):
             self.frictionMu[coneName] = mu
             coneGroup = self.frictionGroup + "/" + coneName
 
-            self.robot.viewer.gui.deleteNode(coneGroup + "/lines/0", "")
-            self.robot.viewer.gui.deleteNode(coneGroup + "/lines/1", "")
-            self.robot.viewer.gui.deleteNode(coneGroup + "/lines/2", "")
-            self.robot.viewer.gui.deleteNode(coneGroup + "/lines/3", "")
-            self.robot.viewer.gui.deleteNode(coneGroup + "/lines", "")
-            self.robot.viewer.gui.deleteNode(coneGroup + "/cone", "")
-            self.robot.viewer.gui.deleteNode(coneGroup, "")
+            self.robot.viewer.gui.deleteNode(coneGroup + "/lines/0", True)
+            self.robot.viewer.gui.deleteNode(coneGroup + "/lines/1", True)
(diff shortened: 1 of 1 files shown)


hrp2-14:
Hi ! This project doesn't usually accept pull requests on master. If this wasn't intentionnal, you can change the base branch of this pull request to devel (No need to close it for that). Best, a bot.

cmastalli (approved):
Thanks @ddliu365 for your contribution :)
Thanks @jmirabel for pointing out quickly the error :)

Looking forward for having more contributions from your part.
I apologize if this was not sorted out earlier.

bindings/python/crocoddyl/__init__.py
```diff
@@ -310,13 +310,13 @@ def _setConeMu(self, coneName, mu):
...
-            self.robot.viewer.gui.deleteNode(coneGroup + "/lines", "")
-            self.robot.viewer.gui.deleteNode(coneGroup + "/cone", "")
-            self.robot.viewer.gui.deleteNode(coneGroup, "")
+            self.robot.viewer.gui.deleteNode(coneGroup + "/lines/0", True)
+            self.robot.viewer.gui.deleteNode(coneGroup + "/lines/1", True)
+            self.robot.viewer.gui.deleteNode(coneGroup + "/lines/2", True)
+            self.robot.viewer.gui.deleteNode(coneGroup + "/lines/3", True)
+            self.robot.viewer.gui.deleteNode(coneGroup + "/lines", True)
+            self.robot.viewer.gui.deleteNode(coneGroup + "/cone", True)
+            self.robot.viewer.gui.deleteNode(coneGroup, True)
```
jmirabel, line 319: The boolean value toggle resistivity. As node *coneGroup* seems to be a group, calling `deleteNode(coneGroup, True)` automatically deletes child nodes so the last line would be sufficient here.

cmastalli: What should I do in the previous lines? should I add `False` or do not include any argument?

jmirabel: You can remove everything but the last line. `deleteNode(coneGroup, True)` will delete all nodes inside the group too.

This is not mandatory. You may prefer to keep the explicit deletion.

ddliu365: I checked the animation in Gepetto, it is same with these codes and without theses codes. For clarity, I just deleted the first six lines in the new pr #898, but I am not sure which one is better.

cmastalli: I will mark as resolved this discussion since the reviews are done in the PR.
Thanks guys :)

PR #899, opened 2021-01-06
Created residual models to be shared between cost and constraint models

cmastalli:
This PR creates the *residual-model* concept to be shared posteriori in our constraint models.
Currently, this is an ongoing work. However, I create this PR to give you time for providing any feedback.
Probably, the most important point is the *reference API*, however, I am following what we have agreed.
@proyan indeed, we mightn't need to work at the cost level for the reference. Let's discuss this in private :)

Coming back to the PR, the main handled points are:
  1. An abstract class for residual model and data (with Python bindings)
  2. Define the residual model and extra constructor functions in cost base (do not break API)
  3. Added the residual data object in cost base data (it will replace `r`, `Rx`, `Ru` in cost base data; the only change in the API)
  4. Created the frame velocity residual with bindings, note that it follows our recent convention regarding the reference
  5. Created residual factory, and numdiff
  6. Incorporated in the residual into its cost model
  7. Deprecated the very old getter/setter functions in frame velocity cost
  8. Included this residual into its factory for c++ unittesting
  9. Adapted the Python derived cost for unittesting

Before merging this PR, I plan to do point 6-9 for the remaining residuals:
- [x] centroidal momentum
- [x] com position
- [x] contact cop position
- [x] contact force
- [x] contact friction cone
- <del>[x] contact impulse</del>
- [x] contact wrench cone
- [x] frame placement
- [x] frame rotation
- [x] frame translation
- [x] impulse com
- <del>[ ] impulse cop position</del>
- <del>[ ] impulse friction cone</del>
- <del>[x] impulse wrench cone</del>
- [x] state
- [x] control
- [x] control gravity
- [x] contact gravity contact 

Additionally, we need to revise the API for:
- [x] residuals in action models
- [x] residuals in diff-action models

which are only used for numdiff routines.

I will work locally in the remaining residual for at least a couple days, thus, I give you time to focus on this particular residual. However, it is very important a prompt feedback as it is already a big effort that I am handling alone.

### DIFF
--- a/bindings/python/crocoddyl/core/core.cpp
+++ b/bindings/python/crocoddyl/core/core.cpp
@@ -28,6 +28,7 @@ void exposeCore() {
   exposeCostAbstract();
   exposeResidualControl();
   exposeCostSum();
+  exposeCostResidual();
   exposeCostControl();
   exposeActionNumDiff();
   exposeDifferentialActionNumDiff();
--- a/bindings/python/crocoddyl/deprecated.py
+++ b/bindings/python/crocoddyl/deprecated.py
@@ -1,5 +1,6 @@
 import functools
 import warnings
+from types import FunctionType
 
 
 class DeprecatedWarning(UserWarning):
@@ -30,3 +31,36 @@ def deprecated(instructions):
         return wrapper
 
     return decorator
+
+
+def depr_wrapper(f, warning):
+    def new(*args, **kwargs):
+        if not args[0].warned:
+            print("Deprecated Warning: %s" % warning)
+            args[0].warned = True
+        return f(*args, **kwargs)
+
+    return new
+
+
+def DeprecatedObject(o, warning):
+    class temp(o):
+        pass
+
+    temp.__name__ = "Deprecated_%s" % o.__class__.__name__
+    output = temp.__new__(temp, o)
+
+    output.warned = True
+    wrappable_types = (type(int.__add__), type(zip), FunctionType)
+    unwrappable_names = ("__str__", "__unicode__", "__repr__", "__getattribute__", "__setattr__")
+
+    for method_name in dir(temp):
+        if not type(getattr(temp, method_name)) in wrappable_types:
+            continue
+        if method_name in unwrappable_names:
+            continue
+        if method_name != "__class__":
+            setattr(temp, method_name, depr_wrapper(getattr(temp, method_name), warning))
+
+    output.warned = False
+    return output
--- a/bindings/python/crocoddyl/deprecated.py
+++ b/bindings/python/crocoddyl/deprecated.py
@@ -33,9 +33,9 @@ def deprecated(instructions):
 
 
 class DeprecationHelper(object):
-    def __init__(self, new_target, warning_str):
+    def __init__(self, new_target, old_name):
         self.new_target = new_target
-        self.warning_str = warning_str
+        self.warning_str = '%s is deprecated: Use %s' % (old_name, new_target.__name__)
 
     def _warn(self):
         warnings.warn(self.warning_str)
--- a/bindings/python/crocoddyl/deprecated.py
+++ b/bindings/python/crocoddyl/deprecated.py
@@ -1,6 +1,5 @@
 import functools
 import warnings
-from types import FunctionType
 
 
 class DeprecatedWarning(UserWarning):
@@ -33,34 +32,18 @@ def deprecated(instructions):
     return decorator
 
 
-def depr_wrapper(f, warning):
-    def new(*args, **kwargs):
-        if not args[0].warned:
-            print("Deprecated Warning: %s" % warning)
-            args[0].warned = True
-        return f(*args, **kwargs)
+class DeprecationHelper(object):
+    def __init__(self, new_target, warning_str):
+        self.new_target = new_target
+        self.warning_str = warning_str
 
-    return new
+    def _warn(self):
+        warnings.warn(self.warning_str)
 
+    def __call__(self, *args, **kwargs):
+        self._warn()
+        return self.new_target(*args, **kwargs)
 
-def DeprecatedObject(o, warning):
-    class temp(o):
-        pass
-
-    temp.__name__ = "Deprecated_%s" % o.__class__.__name__
-    output = temp.__new__(temp, o)
-
-    output.warned = True
-    wrappable_types = (type(int.__add__), type(zip), FunctionType)
-    unwrappable_names = ("__str__", "__unicode__", "__repr__", "__getattribute__", "__setattr__")
-
-    for method_name in dir(temp):
-        if not type(getattr(temp, method_name)) in wrappable_types:
-            continue
-        if method_name in unwrappable_names:
-            continue
-        if method_name != "__class__":
-            setattr(temp, method_name, depr_wrapper(getattr(temp, method_name), warning))
-
-    output.warned = False
-    return output
+    def __getattr__(self, attr):
+        self._warn()
+        return getattr(self.new_target, attr)
--- a/bindings/python/crocoddyl/multibody/residuals/com-position.cpp
+++ b/bindings/python/crocoddyl/multibody/residuals/com-position.cpp
@@ -19,7 +19,7 @@ void exposeResidualCoMPosition() {
       "ResidualModelCoMPosition",
       "This residual function defines the CoM tracking as r = c - cref, with c and cref as the current and reference "
       "CoM position, respectively.",
-      bp::init<boost::shared_ptr<StateMultibody>, Eigen::Vector3d, int>(
+      bp::init<boost::shared_ptr<StateMultibody>, Eigen::Vector3d, std::size_t>(
           bp::args("self", "state", "cref", "nu"),
           "Initialize the CoM position residual model.\n\n"
           ":param state: state of the multibody system\n"
--- a/include/crocoddyl/core/activations/quadratic-barrier.hpp
+++ b/include/crocoddyl/core/activations/quadratic-barrier.hpp
@@ -48,10 +48,10 @@ struct ActivationBoundsTpl {
         }
       }
       // Assign the maximum value for infinity/nan values
-      if (!std::isfinite(lb(i))) {
+      if (!isfinite(lb(i))) {
         lb(i) = -std::numeric_limits<Scalar>::max();
       }
-      if (!std::isfinite(ub(i))) {
+      if (!isfinite(ub(i))) {
         ub(i) = std::numeric_limits<Scalar>::max();
       }
     }
--- a/include/crocoddyl/core/activations/quadratic-barrier.hpp
+++ b/include/crocoddyl/core/activations/quadratic-barrier.hpp
@@ -40,6 +40,7 @@ struct ActivationBoundsTpl {
       throw_pretty("Invalid argument: "
                    << "The range of beta is between 0 and 1");
     }
+    using std::isfinite;
     for (std::size_t i = 0; i < static_cast<std::size_t>(lb.size()); ++i) {
       if (isfinite(lb(i)) && isfinite(ub(i))) {
         if (lb(i) - ub(i) > 0) {
--- a/include/crocoddyl/core/activations/quadratic-barrier.hpp
+++ b/include/crocoddyl/core/activations/quadratic-barrier.hpp
@@ -40,7 +40,6 @@ struct ActivationBoundsTpl {
       throw_pretty("Invalid argument: "
                    << "The range of beta is between 0 and 1");
     }
-    using std::isfinite;
     for (std::size_t i = 0; i < static_cast<std::size_t>(lb.size()); ++i) {
       if (isfinite(lb(i)) && isfinite(ub(i))) {
         if (lb(i) - ub(i) > 0) {
@@ -48,11 +47,18 @@ struct ActivationBoundsTpl {
                        << "The lower and upper bounds are badly defined; ub has to be bigger / equals to lb");
         }
       }
+      // Assign the maximum value for infinity/nan values
+      if (!std::isfinite(lb(i))) {
+        lb(i) = -std::numeric_limits<Scalar>::max();
+      }
+      if (!std::isfinite(ub(i))) {
+        ub(i) = std::numeric_limits<Scalar>::max();
+      }
     }
 
     if (beta >= Scalar(0) && beta <= Scalar(1.)) {
-      VectorXs m = Scalar(0.5) * (lower + upper);
-      VectorXs d = Scalar(0.5) * (upper - lower);
+      VectorXs m = Scalar(0.5) * (lb + ub);
+      VectorXs d = Scalar(0.5) * (ub - lb);
       lb = m - beta * d;
       ub = m + beta * d;
     } else {
--- a/include/crocoddyl/core/cost-base.hpp
+++ b/include/crocoddyl/core/cost-base.hpp
@@ -68,8 +68,7 @@ class CostModelAbstractTpl {
    * @param[in] activation  Activation model
    * @param[in] residual    Residual model
    */
-  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
-		       boost::shared_ptr<ActivationModelAbstract> activation,
+  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state, boost::shared_ptr<ActivationModelAbstract> activation,
                        boost::shared_ptr<ResidualModelAbstract> residual);
 
   /**
@@ -79,8 +78,7 @@ class CostModelAbstractTpl {
    * @param[in] activation  Activation model
    * @param[in] nu          Dimension of control vector
    */
-  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
-		       boost::shared_ptr<ActivationModelAbstract> activation,
+  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state, boost::shared_ptr<ActivationModelAbstract> activation,
                        const std::size_t nu);
 
   /**
@@ -91,8 +89,7 @@ class CostModelAbstractTpl {
    * @param[in] state       State of the dynamical system
    * @param[in] activation  Activation model
    */
-  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
-		       boost::shared_ptr<ActivationModelAbstract> activation);
+  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state, boost::shared_ptr<ActivationModelAbstract> activation);
 
   /**
    * @copybrief CostModelAbstractTpl()
@@ -102,8 +99,7 @@ class CostModelAbstractTpl {
    * @param[in] state     State of the dynamical system
    * @param[in] residual  Residual model
    */
-  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
-		       boost::shared_ptr<ResidualModelAbstract> residual);
+  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state, boost::shared_ptr<ResidualModelAbstract> residual);
 
   /**
    * @copybrief CostModelAbstractTpl()
@@ -114,9 +110,7 @@ class CostModelAbstractTpl {
    * @param[in] nr     Dimension of residual vector
    * @param[in] nu     Dimension of control vector
    */
-  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
-		       const std::size_t nr,
-		       const std::size_t nu);
+  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state, const std::size_t nr, const std::size_t nu);
 
   /**
    * @copybrief CostModelAbstractTpl()
--- a/include/crocoddyl/core/cost-base.hpp
+++ b/include/crocoddyl/core/cost-base.hpp
@@ -68,7 +68,8 @@ class CostModelAbstractTpl {
    * @param[in] activation  Activation model
    * @param[in] residual    Residual model
    */
-  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state, boost::shared_ptr<ActivationModelAbstract> activation,
+  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
+		       boost::shared_ptr<ActivationModelAbstract> activation,
                        boost::shared_ptr<ResidualModelAbstract> residual);
 
   /**
@@ -78,7 +79,8 @@ class CostModelAbstractTpl {
    * @param[in] activation  Activation model
    * @param[in] nu          Dimension of control vector
    */
-  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state, boost::shared_ptr<ActivationModelAbstract> activation,
+  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
+		       boost::shared_ptr<ActivationModelAbstract> activation,
                        const std::size_t nu);
 
   /**
@@ -89,7 +91,8 @@ class CostModelAbstractTpl {
    * @param[in] state       State of the dynamical system
    * @param[in] activation  Activation model
    */
-  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state, boost::shared_ptr<ActivationModelAbstract> activation);
+  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
+		       boost::shared_ptr<ActivationModelAbstract> activation);
 
   /**
    * @copybrief CostModelAbstractTpl()
@@ -99,7 +102,8 @@ class CostModelAbstractTpl {
    * @param[in] state     State of the dynamical system
    * @param[in] residual  Residual model
    */
-  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state, boost::shared_ptr<ResidualModelAbstract> residual);
+  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
+		       boost::shared_ptr<ResidualModelAbstract> residual);
 
   /**
    * @copybrief CostModelAbstractTpl()
@@ -110,7 +114,9 @@ class CostModelAbstractTpl {
    * @param[in] nr     Dimension of residual vector
    * @param[in] nu     Dimension of control vector
    */
-  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state, const std::size_t nr, const std::size_t nu);
+  CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
+		       const std::size_t nr,
+		       const std::size_t nu);
 
   /**
    * @copybrief CostModelAbstractTpl()
--- a/include/crocoddyl/core/cost-base.hpp
+++ b/include/crocoddyl/core/cost-base.hpp
@@ -205,7 +205,7 @@ class CostModelAbstractTpl {
    * @brief Return the cost reference
    */
   template <class ReferenceType>
-  ReferenceType get_reference() const;
+  ReferenceType get_reference();
 
  protected:
   /**
@@ -216,7 +216,7 @@ class CostModelAbstractTpl {
   /**
    * @copybrief get_reference()
    */
-  virtual void get_referenceImpl(const std::type_info&, void*) const;
+  virtual void get_referenceImpl(const std::type_info&, void*);
 
   boost::shared_ptr<StateAbstract> state_;                 //!< State description
   boost::shared_ptr<ActivationModelAbstract> activation_;  //!< Activation model
--- a/include/crocoddyl/core/cost-base.hpp
+++ b/include/crocoddyl/core/cost-base.hpp
@@ -232,6 +232,7 @@ struct CostDataAbstractTpl {
   typedef _Scalar Scalar;
   typedef MathBaseTpl<Scalar> MathBase;
   typedef ActivationDataAbstractTpl<Scalar> ActivationDataAbstract;
+  typedef ResidualDataAbstractTpl<Scalar> ResidualDataAbstract;
   typedef DataCollectorAbstractTpl<Scalar> DataCollectorAbstract;
   typedef typename MathBase::VectorXs VectorXs;
   typedef typename MathBase::MatrixXs MatrixXs;
--- a/include/crocoddyl/core/cost-base.hxx
+++ b/include/crocoddyl/core/cost-base.hxx
@@ -122,14 +122,14 @@ void CostModelAbstractTpl<Scalar>::set_referenceImpl(const std::type_info&, cons
 
 template <typename Scalar>
 template <class ReferenceType>
-ReferenceType CostModelAbstractTpl<Scalar>::get_reference() const {
+ReferenceType CostModelAbstractTpl<Scalar>::get_reference() {
   ReferenceType ref;
   get_referenceImpl(typeid(ref), &ref);
   return ref;
 }
 
 template <typename Scalar>
-void CostModelAbstractTpl<Scalar>::get_referenceImpl(const std::type_info&, void*) const {
+void CostModelAbstractTpl<Scalar>::get_referenceImpl(const std::type_info&, void*) {
   throw_pretty("It has not been implemented the set_referenceImpl() function");
 }
 
--- a/include/crocoddyl/core/costs/control.hxx
+++ b/include/crocoddyl/core/costs/control.hxx
@@ -1,7 +1,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2019-2021, LAAS-CNRS, University of Edinburgh
+// Copyright (C) 2019-2021, LAAS-CNRS, University of Edinburgh, INRIA
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
@@ -68,14 +68,6 @@ CostModelControlTpl<Scalar>::CostModelControlTpl(boost::shared_ptr<typename Base
 
 template <typename Scalar>
 CostModelControlTpl<Scalar>::~CostModelControlTpl() {}
-
-template <typename Scalar>
-void CostModelControlTpl<Scalar>::calc(const boost::shared_ptr<CostDataAbstract>& data,
-                                       const Eigen::Ref<const VectorXs>& x, const Eigen::Ref<const VectorXs>& u) {
-  residual_->calc(data->residual, x, u);
-  activation_->calc(data->activation, data->residual->r);
-  data->cost = data->activation->a_value;
-}
 
 template <typename Scalar>
 void CostModelControlTpl<Scalar>::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
--- a/include/crocoddyl/core/costs/control.hxx
+++ b/include/crocoddyl/core/costs/control.hxx
@@ -6,6 +6,7 @@
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
+#include <iostream>
 #include "crocoddyl/core/utils/exception.hpp"
 
 namespace crocoddyl {
@@ -14,6 +15,7 @@ CostModelControlTpl<Scalar>::CostModelControlTpl(boost::shared_ptr<typename Base
                                                  boost::shared_ptr<ActivationModelAbstract> activation,
                                                  const VectorXs& uref)
     : Base(state, activation, boost::make_shared<ResidualModelControl>(state, uref)), uref_(uref) {
+  std::cerr << "Deprecated CostModelControl: Use ResidualModelControl with CostModelResidual" << std::endl;
   if (activation_->get_nr() != nu_) {
     throw_pretty("Invalid argument: "
                  << "nr is equals to " + std::to_string(nu_));
@@ -25,6 +27,7 @@ CostModelControlTpl<Scalar>::CostModelControlTpl(boost::shared_ptr<typename Base
                                                  boost::shared_ptr<ActivationModelAbstract> activation)
     : Base(state, activation, boost::make_shared<ResidualModelControl>(state)),
       uref_(VectorXs::Zero(activation->get_nr())) {
+  std::cerr << "Deprecated CostModelControl: Use ResidualModelControl with CostModelResidual" << std::endl;
   if (activation_->get_nr() != nu_) {
     throw_pretty("Invalid argument: "
                  << "nr is equals to " + std::to_string(nu_));
@@ -36,6 +39,7 @@ CostModelControlTpl<Scalar>::CostModelControlTpl(boost::shared_ptr<typename Base
                                                  boost::shared_ptr<ActivationModelAbstract> activation,
                                                  const std::size_t nu)
     : Base(state, activation, boost::make_shared<ResidualModelControl>(state, nu)), uref_(VectorXs::Zero(nu)) {
+  std::cerr << "Deprecated CostModelControl: Use ResidualModelControl with CostModelResidual" << std::endl;
   if (activation_->get_nr() != nu_) {
     throw_pretty("Invalid argument: "
                  << "nr is equals to " + std::to_string(nu_));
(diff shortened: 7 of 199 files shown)


bindings/python/crocoddyl/core/core.cpp
```diff
@@ -1,7 +1,8 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2018-2020, University of Edinburgh, LAAS-CNRS
+// Copyright (C) 2018-2021, University of Edinburgh
```
jcarpent, line 4: The copyright is not correct for Edinburgh. 2018 is not correct.

jcarpent:
Why there is so many cone (contact, impulse, etc.). One formalism is enough.

cmastalli:
> Why there is so many cone (contact, impulse, etc.). One formalism is enough.

Yes, indeed you might be correct. I will remove the extra action points later (if this the case)

cmastalli:
> Why there is so many cone (contact, impulse, etc.). One formalism is enough.

The problem with fusing both is data duplication. Contact and impulse are stored in different data structures as they contain different information. If we create a residual function that contains both, then one of these data pointers will be empty (we could use an internal flag to indicate whether is a contact or impulse).

This data duplication is not a major computational pitfall as it is only a shared pointer. But it will create a lot of confusion from the user side.

Having two classes doesn't help to maintain the project, having two data objects doesn't help users.
What is the best approach? (I am inclined for having two classes as written in the initial message)

proyan (changes requested):
I've reviewed about 40% of the PR, and most of the comments apply to the majority of the cost-models in this PR. I'll let you handle it while I review the rest.

It is better to remove the cost models by simple typedefs which we can deprecate. 
In our previous discussion, you had pointed out that calculation of derivatives for for an independent x or u vector would lead to zero matrix multiplication.

A simple way out would be to store this dependency on a state vector and control vector in the residual itself. 
This can be denoted by 
```c++
bool x_dependent, u_dependent;
Eigen::Vector2i x_segment, dx_segment, u_segment;

```
in the residual models. 

For example, for the case of the CoMResidual, this means that in the constructor, we set 
```c++
x_dependent=true;
u_dependent=false;
x_segment = Vector2i(0,state->pinocchio->nq);
dx_segment = Vector2i(0,state->pinocchio->nv);
u_segment = Vector2i(0, state->nu);
```

here, x_dependent and u_dependent are flags which tell you which fn r(x, u) or r(u) or r(x) to apply, and x_segment, u_segment tell you which continuous segment of state or control is active

In the cost-model, the application of calc-diff would be as follows:

```c++
if(residual->x_dependent) {
Arr_Rx.cols(residual->dx_segment[0], residual->dx_segment[1]).noalias() = activation->Arr * Rx.segment(residual.dx_segment[0], residual->dx_segment[1]);

Lx.segment(residual->dx_segment[0], residual->dx_segment[1]).noalias() = Rx.segment(residual.dx_segment[0], residual->dx_segment[1]) * activation->Ar;

Lxx.block(<dx segment col and rows>). noalias() = Rx.segment() * Arr_Rx.cols();
}
if(residual->u_dependent) {
//same here
}
```

Once you do this, all cost models can be removed by a simple deprecated typedef

include/crocoddyl/core/numdiff/cost.hpp
```diff
@@ -167,9 +167,7 @@ struct CostDataNumDiffTpl : public CostDataAbstractTpl<_Scalar> {
   using Base::Lx;
   using Base::Lxu;
   using Base::Lxx;
-  using Base::r;
-  using Base::Ru;
-  using Base::Rx;
+  using Base::residual;
```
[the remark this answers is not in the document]
cmastalli: I am planning to remove `r`, `Rx`, and `Ru` from the base class.
This is not done as I first need to completed all the migration.

Note that we are using a shared_ptr (named residual) to store all these quantities.
It is the similar design done for the activation classes.

I believe this conversation can be marked as resolved.
However, please feel free to re-open it if you think there is something more to discuss

include/crocoddyl/multibody/costs/centroidal-momentum.hxx
```diff
@@ -38,49 +39,43 @@ CostModelCentroidalMomentumTpl<Scalar>::CostModelCentroidalMomentumTpl(
...
-  }
-
-  d->Arr_Rx.noalias() = data->activation->Arr * data->Rx;
-  data->Lx.noalias() = data->Rx.transpose() * data->activation->Ar;
-  data->Lxx.noalias() = data->Rx.transpose() * d->Arr_Rx;
+  activation_->calcDiff(data->activation, data->residual->r);
+  residual_->calcDiff(d->residual, x, u);
+
+  // Compute the derivatives of the cost function based on a Gauss-Newton approximation
+  Eigen::Ref<Matrix6xs> Rx(data->residual->Rx);
```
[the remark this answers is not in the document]
cmastalli: The idea was to perform the above operations with a fixed number of rows.
I though this was more efficient, isn't it?

As far I understand, this code doesn't allocate data.

**Let me know if you still advice to do not use this instance**

include/crocoddyl/multibody/costs/com-position.hxx
```diff
@@ -36,39 +37,42 @@ CostModelCoMPositionTpl<Scalar>::CostModelCoMPositionTpl(boost::shared_ptr<State
...
-  data->Rx.leftCols(nv) = d->pinocchio->Jcom;
-  data->Lx.head(nv).noalias() = d->pinocchio->Jcom.transpose() * data->activation->Ar;
-  d->Arr_Jcom.noalias() = data->activation->Arr * d->pinocchio->Jcom;
-  data->Lxx.topLeftCorner(nv, nv).noalias() = d->pinocchio->Jcom.transpose() * d->Arr_Jcom;
+  residual_->calcDiff(d->residual, x, u);
+  activation_->calcDiff(data->activation, data->residual->r);
+
+  // Compute the derivatives of the cost function based on a Gauss-Newton approximation
+  const std::size_t nv = state_->get_nv();
+  Eigen::Ref<Matrix3xs> Jcom(data->residual->Rx.leftCols(nv));
```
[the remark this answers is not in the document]
cmastalli: There are a few places with this pattern.
I will propagate the decision in https://github.com/loco-3d/crocoddyl/pull/899#discussion_r556497770 to the remaining cases (including this one)

I will close this conversation as this is the same point described before (this is for me to track easily the needed modifications)

include/crocoddyl/core/numdiff/cost.hpp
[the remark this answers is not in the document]
proyan: Don't remove r, Rx and Ru. Make them references to residual->r, residual->Rx and residual-Ru, and deprecate them. There are users who are using costdata->r values

include/crocoddyl/multibody/costs/centroidal-momentum.hxx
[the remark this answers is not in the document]
proyan: No, it is more efficient to use a block reference

include/crocoddyl/core/numdiff/cost.hpp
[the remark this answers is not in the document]
proyan: see my comment here https://github.com/loco-3d/crocoddyl/pull/899#discussion_r556473762

cmastalli:
@proyan thanks for the review. I don't expect you to review the whole thing now.

Please make comments only on the frame-velocity case (if there is still something missing)
I will propagate each point in the other cases.
(This will help us to make this review process easier/faster)

When the PR is finished, I will let you know.
Then, it will be helpful if you do a full review (just to check is I missed something).

include/crocoddyl/core/cost-base.hxx
```diff
@@ -1,33 +1,71 @@
...
 template <typename Scalar>
 CostModelAbstractTpl<Scalar>::CostModelAbstractTpl(boost::shared_ptr<StateAbstract> state,
                                                    boost::shared_ptr<ActivationModelAbstract> activation,
                                                    const std::size_t& nu)
-    : state_(state), activation_(activation), nu_(nu), unone_(VectorXs::Zero(nu)) {}
+    : state_(state),
+      activation_(activation),
+      residual_(boost::make_shared<ResidualModelAbstract>(state, activation->get_nr(), nu)),
+      nu_(nu),
+      unone_(VectorXs::Zero(nu)) {}
```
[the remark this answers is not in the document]
cmastalli: Yes, of course

bindings/python/crocoddyl/core/core.cpp
cmastalli: I have discussed with @nim65s about the copyright information.
This file, as each one defined in the c++ library, needs to start from 2019 (including LAAS-CNRS).

As @nim65s explained me, these files were not a copy-pasted from 2018 library as this library was purely written in Python.
It is the same for the code that defines the algorithms (e.g. SolverDDP) as they were not copy-pasted neither (I did rewritten them).

Remember that I worked in 2019 for LAAS and Edinburgh.
So, there are many files where both institutions share the copy-holders from 2019.

This requires to modify the copyright information for the entire library, not only the files in this PR.
Thus, I will create another PR only focuses on this topic.
Then I will rebase this PR and check again the information for the modified files.

cmastalli: I forgot to say that the c++ library was written in 2019 (summertime).
And in autumn, I moved to Edinburgh where I kept working quite a lot on the library.

cmastalli: done!

jcarpent:
@cmastalli Are you still sure to need Impulse and Contact wrench cones? Because, from a mathematical viewpoint, they are totally equivalent.

cmastalli:
> @cmastalli Are you still sure to need Impulse and Contact wrench cones? Because, from a mathematical viewpoint, they are totally equivalent.

Each wrench cone (for the impulses or contacts) uses a specific data. If you have a clever idea how to solve this issue, then I would happily integrated in this PR.

cmastalli:
@jcarpent  I have figured out a solution for combining both residuals without affecting performance.

include/crocoddyl/multibody/costs/centroidal-momentum.hxx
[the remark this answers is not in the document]
cmastalli: Could you clarify more your suggestion?

I don't understand why you suggested to use block reference.

If we don't want to use `Eigen::Ref<Matrix6xs>`, then we do need to use `data->residual->Rx` instead of `Rx` in the above multiplication. Is this what do you suggest?

include/crocoddyl/core/cost-base.hpp
```diff
@@ -233,15 +263,16 @@ struct CostDataAbstractTpl {
...
   Scalar cost;
   VectorXs Lx;
   VectorXs Lu;
   MatrixXs Lxx;
   MatrixXs Lxu;
   MatrixXs Luu;
-  VectorXs r;
-  MatrixXs Rx;
-  MatrixXs Ru;
+  VectorXs r;   // TODO(cmastalli) remove after full integration of residual data
```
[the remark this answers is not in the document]
cmastalli: done

cmastalli:
@proyan -- as discussed in private, this PR is ready to be reviewed.
It includes the `CostModelResidual` as proposed by you in this thread; however, it doesn't deprecate all cost models.
I would push the deprecation work (to work together) after merging this PR.

include/crocoddyl/multibody/costs/centroidal-momentum.hxx
```diff
@@ -93,15 +86,19 @@ template <typename Scalar>
...
 template <typename Scalar>
-void CostModelCentroidalMomentumTpl<Scalar>::get_referenceImpl(const std::type_info& ti, void* pv) const {
+void CostModelCentroidalMomentumTpl<Scalar>::get_referenceImpl(const std::type_info& ti, void* pv) {
   if (ti == typeid(Vector6s)) {
     Eigen::Map<Vector6s> ref_map(static_cast<Vector6s*>(pv)->data());
+    ResidualModelCentroidalMomentum* residual = static_cast<ResidualModelCentroidalMomentum*>(residual_.get());
+    href_ = residual->get_reference();
     ref_map[0] = href_[0];
     ref_map[1] = href_[1];
     ref_map[2] = href_[2];
```
proyan, line 104: It looks like you are 
-> first copying residual->href_ into href_,
-> and then copying href_ into ref_map.

Instead, I suggest to make href_ as a reference that binds to residual ->href_ at construction time.
And to assign this to the pv pointer, use (syntax may not be correct)
```c++
pv = &residual->href_(0); // get the address storing the data for href_
Eigen::Map<Vector6s> ref_map (pv, 6); 
```

include/crocoddyl/multibody/costs/com-position.hxx
```diff
@@ -36,39 +36,41 @@ CostModelCoMPositionTpl<Scalar>::CostModelCoMPositionTpl(boost::shared_ptr<State
...
 template <typename Scalar>
 CostModelCoMPositionTpl<Scalar>::CostModelCoMPositionTpl(boost::shared_ptr<StateMultibody> state, const Vector3s& cref)
-    : Base(state, 3), cref_(cref) {}
+    : Base(state, boost::make_shared<ResidualModelCoMPosition>(state, cref)), cref_(cref) {}
 
 template <typename Scalar>
 CostModelCoMPositionTpl<Scalar>::~CostModelCoMPositionTpl() {}
 
 template <typename Scalar>
 void CostModelCoMPositionTpl<Scalar>::calc(const boost::shared_ptr<CostDataAbstract>& data,
```
proyan, line 49: If we move the implementation of costmodelresidual to costmodelbase, we can get rid of the most of the duplication of code in `calc`, `calcdiff`

include/crocoddyl/multibody/costs/com-position.hxx
```diff
@@ -81,15 +83,19 @@ template <typename Scalar>
...
     throw_pretty("Invalid argument: incorrect type (it should be Vector3s)");
   }
 }
 
 template <typename Scalar>
-void CostModelCoMPositionTpl<Scalar>::get_referenceImpl(const std::type_info& ti, void* pv) const {
+void CostModelCoMPositionTpl<Scalar>::get_referenceImpl(const std::type_info& ti, void* pv) {
   if (ti == typeid(Vector3s)) {
     Eigen::Map<Vector3s> ref_map(static_cast<Vector3s*>(pv)->data());
+    ResidualModelCoMPosition* residual = static_cast<ResidualModelCoMPosition*>(residual_.get());
```
proyan, line 97: This is a duplicate of the previous comment in centroidal-momentum-cost

include/crocoddyl/multibody/costs/centroidal-momentum.hpp
```diff
@@ -184,9 +168,7 @@ struct CostDataCentroidalMomentumTpl : public CostDataAbstractTpl<_Scalar> {
   using Base::Lx;
   using Base::Lxu;
   using Base::Lxx;
-  using Base::r;
-  using Base::Ru;
-  using Base::Rx;
+  using Base::residual;
```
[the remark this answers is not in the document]
proyan: For example, cost.ref_c is bound to residual.ref_b in the following example:

```c++
#include <iostream>                                                                           
struct Residual {                                                                             
                                                                                              
  Residual(int r) : ref_b(r) {}                                                               
  int ref_b;                                                                                  
};                                                                                            
                                                                                              
struct Cost {                                                                                 
                                                                                              
  Cost(int r):                                                                                
    residual(r),                                                                              
    ref_c(residual.ref_b) {}                                                                  
                                                                                              
  Residual residual;                                                                          
  int & ref_c;                                                                                
};                                                                                            
                                                                                              
int main() {                                                                                  
                                                                                              
                                                                                              
  Cost c(4);                                                                                  
  std::cerr<<c.ref_c<<std::endl;                                                              
  c.residual.ref_b = 10;                                                                      
  std::cerr<<c.ref_c<<std::endl;                                                              
  c.ref_c = 20;                                                                               
  std::cerr<<c.residual.ref_b<<std::endl;                                                     
  return 0;                                                                                   
}
```

proyan (changes requested):
Thanks Carlos for this massive effort.
I'll leave a partial review so that we can start discussing while I keep reviewing.
Mostly, there are two changes that I think can be made in this implementation: 
1) CostModelResidual can be merged with CostModelBase.
2) Redundant calc and calcDiff computations can be removed from most cost models
3) The components of the residual data can be used as a reference in the cost data.

I will continue to review, so please take your time.

include/crocoddyl/multibody/residuals/frame-placement.hxx
```diff
@@ -0,0 +1,88 @@
...
+void ResidualModelFramePlacementTpl<Scalar>::calc(const boost::shared_ptr<ResidualDataAbstract>& data,
+                                                  const Eigen::Ref<const VectorXs>&,
+                                                  const Eigen::Ref<const VectorXs>&) {
+  Data* d = static_cast<Data*>(data.get());
+
+  // Compute the frame placement w.r.t. the reference frame
+  pinocchio::updateFramePlacement(*pin_model_.get(), *d->pinocchio, id_);
+  d->rMf = oMf_inv_ * d->pinocchio->oMf[id_];
+  d->r = pinocchio::log6(d->rMf);
+  data->r = d->r;  // this is needed because we overwrite it
```
proyan, line 44: Why do we need to overwrite it? Sorry I don't understand.

include/crocoddyl/multibody/residuals/frame-rotation.hpp
```diff
@@ -0,0 +1,181 @@
...
+    if (d == NULL) {
+      throw_pretty("Invalid argument: the shared data should be derived from DataCollectorMultibody");
+    }
+
+    // Avoids data casting at runtime
+    pinocchio = d->pinocchio;
+  }
+
+  pinocchio::DataTpl<Scalar>* pinocchio;  //!< Pinocchio data
+  Vector3s r;
```
proyan, line 163: Same here. Why are we redefining r here?

include/crocoddyl/core/costs/residual.hxx
```diff
@@ -0,0 +1,87 @@
...
+// All rights reserved.
+///////////////////////////////////////////////////////////////////////////////
+
+#include "crocoddyl/core/utils/exception.hpp"
+
+namespace crocoddyl {
+template <typename Scalar>
+CostModelResidualTpl<Scalar>::CostModelResidualTpl(boost::shared_ptr<typename Base::StateAbstract> state,
+                                                   boost::shared_ptr<ResidualModelAbstract> residual,
+                                                   boost::shared_ptr<ActivationModelAbstract> activation)
```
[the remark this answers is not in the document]
cmastalli: I think it is better to keep the `CostModelAbstract` class, as 1) it allows us to develop any cost function without the `residual` assumption (i.e., Newton-Gauss approximation), and 2) it does not break the API.

On the other hand, we could define `calc` and `calcDiff` as nonpure virtual functions (as an action to handle the above-mentioned points), but then the code becomes tricky as it is hard to spot an issue if the users implement only one of these functions.



> Eventually, the idea is to remove all children of CostModelBase.

I have already solved this issue. But my plan is to create a follow-up PR with this deprecation code (the idea is to do not make this PR more complex).

Should you agree to mark this conversation as resolved?

include/crocoddyl/multibody/costs/centroidal-momentum.hxx
cmastalli: This code will be deprecated in the follow-up PR (very soon).

I will avoid any unnecessary effort on improving the tailored cost functions.

include/crocoddyl/multibody/costs/com-position.hxx
cmastalli: Same comment: this code will be deprecated soon.

I am marking these conversation as resolved, unfortunately, I don't think we should waste our time by improving a code that will be deprecated :)

include/crocoddyl/multibody/residuals/frame-rotation.hpp
cmastalli: Same explanation (see https://github.com/loco-3d/crocoddyl/pull/899#discussion_r604258217)

include/crocoddyl/multibody/costs/com-position.hxx
proyan: I am okay with keeping the current implementation, since it would need be deprecated soon. 

In that case, the only thing to be done is to bind the references, r, rx, ru to residual.ref, r, rx, ru, so that there is no break in the API.

include/crocoddyl/multibody/costs/centroidal-momentum.hxx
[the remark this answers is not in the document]
proyan: Do you have compilation issues with this?

include/crocoddyl/core/cost-base.hpp
```diff
@@ -73,7 +86,7 @@ class CostModelAbstractTpl {
    *
    * The default `nu` value is obtained from `StateAbstractTpl::get_nv()`.
    *
-   * @param[in] state       State of the multibody system
+   * @param[in] state       State of the dynamical system
```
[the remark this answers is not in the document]
cmastalli: The proper name is *dynamical* system as it refers to a system that involves dynamics. In contrast, *dynamics* system refers to a system that is changing (not necessary changing dynamically). Indeed, Wikipedia uses *dynamical* system as you can see here: https://en.wikipedia.org/wiki/Dynamical_system

I will close this issue :)

include/crocoddyl/multibody/costs/centroidal-momentum.hxx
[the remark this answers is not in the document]
cmastalli: done

include/crocoddyl/multibody/costs/com-position.hxx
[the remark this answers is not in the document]
cmastalli: done

include/crocoddyl/core/cost-base.hpp
```diff
@@ -211,37 +241,37 @@ struct CostDataAbstractTpl {
...
     Lxx.setZero();
     Lxu.setZero();
     Luu.setZero();
-    r.setZero();
-    Rx.setZero();
-    Ru.setZero();
   }
   virtual ~CostDataAbstractTpl() {}
 
+  DEPRECATED("Use residual.r", const VectorXs& get_r() const { return residual->r; };)
```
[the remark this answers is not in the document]
cmastalli: I would copy your comments for more clarity. You have written in https://github.com/loco-3d/crocoddyl/pull/899#discussion_r556545744 :

> Don't remove r, Rx and Ru. Make them references to residual->r, residual->Rx and residual-Ru, and deprecate them. There are users who are using costdata->r values

It is clear to me how to use `Eigen::Ref` for retrieving the values from `residual->XX` (i.e., "make them references of"). What I don't see possible (or reasonable) from your comment is the deprecation of these member objects. Let me explain you this statement. A way of deprecating this object would be
```cpp
DEPRECATED("Use residual.r", Eigen::Ref<VectorXs> r);
```
However, it is not useful as this deprecation code will always print a deprecation message. In short, unfortunately, there is no a reasonable way to deprecate member objects in a class.

On the other hand, I believe, in this case, it is more informative for users to have a compilation error.
Indeed, it should be straightforward for users to realize how the fixing code looks by inspecting the internal code, e.g.,
```cpp
DEPRECATED("Use residual.r", const VectorXs& get_r() const { return residual->r; };)
```

**Closing remarks**:
I believe this is the best approach that we can do. If you have a solution to the deprecation code issue, which I am not aware of, then let me know. Otherwise, I would suggest to mark this conversation as resolved

include/crocoddyl/core/costs/residual.hxx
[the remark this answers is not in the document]
cmastalli: > I think the way forward would be to do the class hierarchy CostModelAbstract -> CostModelResidual -> CostModelCom..., ... ,

This could be an approach to do. However, I would prefer to avoid this task. Note that each `CostModelCom,...` class will be deprecated. I have already developed this code, but I don't want to push here as it modifies many more files.



> I am okay if you push your implementation in this PR. It would provide a more coherent idea of what you are proposing.

Thank you for the availability (despite the dimension of this PR). But I really prefer to do it in another PR, so you can easily spot any issue. (I want to have this PR merged asap)


I believe I can mark this discussion as resolved. Please feel free to re-open it if you do not believe so!

include/crocoddyl/multibody/costs/centroidal-momentum.hpp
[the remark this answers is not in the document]
cmastalli: My issue is not with the reference, it is with the deprecation code. In short, I have explained here https://github.com/loco-3d/crocoddyl/pull/899#discussion_r605054503 that we could not create a method for deprecating these objects.

I will close this message, as we should keep the discussion in https://github.com/loco-3d/crocoddyl/pull/899#discussion_r605054503 (easier for everybody to track the discussion thread)

cmastalli:
@proyan -- a short comment about last push!

I have changed the definition for boundless barriers. Now, `inf` has to be used to define no bound. This protocol has more logic as *the maximum floating* number shouldn't be considered as no bound. This is important for the constraint development that I am doing right now.

include/crocoddyl/multibody/residuals/frame-placement.hxx
cmastalli: Interesting ... I see where was my mistake. We need to write
```cpp
data->r = pinocchio::log6(d->rMf).toVector();
```
instead of
```cpp
data->r = pinocchio::log6(d->rMf);
```

If we do not use `toVector()`, then Pinocchio expects a vector with fixed dimension.

**In short, I have adapted both cases: frame-placement and frame-rotation. Thanks for the tip!**

include/crocoddyl/multibody/residuals/frame-rotation.hpp
cmastalli: solved

include/crocoddyl/multibody/costs/com-position.hxx
cmastalli: I will mark this conversation as resolved because we agree to keep current implementation (which it will be deprecated soon).

include/crocoddyl/core/costs/residual.hxx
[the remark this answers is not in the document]
proyan: In that case, let's remove this costmodel from this PR. You could add it in the next PR, which would give a better understanding of the final structure.

include/crocoddyl/core/cost-base.hpp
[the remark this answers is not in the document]
proyan: C++ reference would be more efficient than Eigen::Ref here. 
We are already deprecating the constructors of the cost models. Once we finalize on the structure for costmodelresidual, we would add a deprecation to the individual cost models as well.

It is safe to assume that we don't need to deprecate each variable of the costmodel once that happens.
So if we simply replace `VectorXs r` by `VectorXs& r`, I think it should be fine.

include/crocoddyl/multibody/costs/frame-placement.hpp
```diff
@@ -164,32 +161,10 @@ struct CostDataFramePlacementTpl : public CostDataAbstractTpl<_Scalar> {
...
-    // Avoids data casting at runtime
-    pinocchio = d->pinocchio;
   }
 
-  pinocchio::DataTpl<Scalar>* pinocchio;
-  Vector6s r;
-  pinocchio::SE3Tpl<Scalar> rMf;
-  Matrix6xs J;
-  Matrix6s rJf;
-  Matrix6xs fJf;
```
proyan, line 192: The "change to reference" strategy should apply to these quantities as well. For example, CosttDataFramePlacement->rMf is now broken.

include/crocoddyl/core/costs/residual.hxx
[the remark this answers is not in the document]
cmastalli: > In that case, let's remove this costmodel from this PR

What do you mean with *costmodel*? Do you mean `CostModelResidual`? (it is the only cost model that has been included in this PR since you've requested in first revision)

I know it is very stressful to review a large PR; however, removing code that then I will include in another PR it is an unnecessary work. If there is something that concerns you, then I am happy to clarify these bits in the final structure ;)

Note that I will create a follow-up PR (just immediately after merging this PR), with the deprecation code. Indeed, this PR is not changing any structure, instead it is adding a new `CostModelResidual` option.

I believe this discussion is resolved. But I am very keen to answer any question that you might still have about the structure.

include/crocoddyl/core/fwd.hpp
```diff
@@ -143,6 +155,11 @@ class CostModelSumTpl;
 template <typename Scalar>
 struct CostDataSumTpl;
 
+template <typename Scalar>
+class CostModelResidualTpl;
+template <typename Scalar>
+struct CostDataResidualTpl;
```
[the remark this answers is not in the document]
cmastalli: This is related with https://github.com/loco-3d/crocoddyl/pull/899#discussion_r605469885

I will close as these won't be removed

include/crocoddyl/core/activations/quadratic-barrier.hpp
```diff
@@ -40,19 +40,25 @@ struct ActivationBoundsTpl {
       throw_pretty("Invalid argument: "
                    << "The range of beta is between 0 and 1");
     }
-    using std::isfinite;
     for (std::size_t i = 0; i < static_cast<std::size_t>(lb.size()); ++i) {
       if (isfinite(lb(i)) && isfinite(ub(i))) {
```
proyan, line 44: I wonder how this is able to compile if we remove `using std::isfinite`.

include/crocoddyl/core/activations/quadratic-barrier.hpp
```diff
@@ -40,19 +40,25 @@ struct ActivationBoundsTpl {
...
         if (lb(i) - ub(i) > 0) {
           throw_pretty("Invalid argument: "
                        << "The lower and upper bounds are badly defined; ub has to be bigger / equals to lb");
         }
       }
+      // Assign the maximum value for infinity/nan values
+      if (!std::isfinite(lb(i))) {
```
proyan, line 51: Anyways, I remember that there was an issue with the definition of isfinite for codegen classes. As a result, I had fixed the implementation here: https://github.com/stack-of-tasks/pinocchio/blob/master/src/codegen/cppadcg.hpp#L67. 
This was to let the compiler choose which `isfinite` to use depending on the scalar type.
So It would be better to not specify `std::isfinite`, but rather use `isfinite` instead.

include/crocoddyl/multibody/costs/frame-placement.hpp
cmastalli: I would ask you to wait for the next PR, that will deprecate this classes. This is the most pragmatical approach, however, if you believe this is needed, then I would ask you if you can handle it. Unfortunately, I cannot afford to handle this alone.

include/crocoddyl/multibody/force-base.hpp
```diff
@@ -0,0 +1,58 @@
...
+        df_du(model->get_nc(), model->get_nu()) {
+    Jc.setZero();
+    df_dx.setZero();
+    df_du.setZero();
+  }
+  virtual ~ForceDataAbstractTpl() {}
+
+  typename pinocchio::DataTpl<Scalar>* pinocchio;  //!< Pinocchio data
+  pinocchio::JointIndex joint;                     //!< Joint index of the contact frame
+  pinocchio::FrameIndex frame;                     //!< Frame index of the contact frame
```
proyan, line 47: Good idea to create a ForceDataAbstract class. 

Saving both the jointindex and frameindex seems redundant and error-prone. Once you have the frame_id, it is simple to get the parent joint id. I suggest to remove the jointindex from here.

include/crocoddyl/core/activations/quadratic-barrier.hpp
cmastalli: Perhaps, we were not compiling this class in our codegen code (for benchmarking and unittesting)

include/crocoddyl/multibody/force-base.hpp
cmastalli: Done!

include/crocoddyl/core/cost-base.hpp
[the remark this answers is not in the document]
cmastalli: >  So if we simply replace VectorXs r by VectorXs& r in CostDataAbstract, I think it should be fine.

Yes, using `Eigen::Ref` of this approach is not the issue.

What I mean we could not deprecate this code in c++:
```cpp
VectorXs& r
```
as it will always print a deprecation message, EVEN if the user doesn't include this code.


This is the main issue, we cannot deprecate in c++ attribute in a structure (or public attribute in a class). At least there is no solution that I am aware of, note that I have spent time to find a solution.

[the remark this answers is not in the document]
cmastalli: This is a study limitation of c++ (at least c++98). A workaround might not be trivial, and I am clueless here.

include/crocoddyl/core/activations/quadratic-barrier.hpp
cmastalli: done

cmastalli:
@proyan -- do you know why `isfinite` doesn't work in Ubuntu 16.04? (see https://travis-ci.org/github/loco-3d/crocoddyl/jobs/765491428)

include/crocoddyl/core/activations/quadratic-barrier.hpp
```diff
@@ -48,10 +48,10 @@ struct ActivationBoundsTpl {
         }
       }
       // Assign the maximum value for infinity/nan values
-      if (!std::isfinite(lb(i))) {
```
jcarpent: You should use `using namespace std`.

jcarpent: In any case, this code is not compatible with codegen because of the if statements !!!

proyan: I think just keeping the `using std::isfinite` (https://github.com/loco-3d/crocoddyl/pull/899#discussion_r605476858) is enough here.
I agree about the if statements, they need to be handled separately. For now, I don't' think this maximum value setting would work in codegen.

cmastalli: Are you sure this code won't work for codegen?

I am asking it as this if-statement is in the class constructor.

(I have solved this issue by adding `using std::isfinite`)

jcarpent: It depends if you follow or not the design pattern of Pinocchio with the cast methods.

cmastalli: I am not aware of these details. @proyan -- could you let us know?

(I will mark this conversation as resolved)

proyan:
Hi @cmastalli 
In the commits the changes are:
1) Fix the constructor argument order of costmodelresidual to match that of costmodelbase (state, actuation, residual)
2) make costmodelresidual as the base class for all relavant cost classes
3) Remove the costdata classes.
4) add INRIA as the support and copyright holder.

I'm planning to add the deleted CostData classes in the deprecated.py for the python users, and a typedef for the c++ users. However, there would still be one place where the API breaks: for the custom member variables like Arr_J etc, where we don't follow the same nomenclature as CostDataResidual.

I don't think _any_ user is going to be using these internal variables, so I'm going to let this break. Apart from that, I think the changes are done.

proyan:
@nim65s I'm not that strong with using dunder methods in Python. Could you please have a look at the last commit https://github.com/loco-3d/crocoddyl/pull/899/commits/2ea9248a672e6933c291520ba5b5a8f7991bc7de to verify that this Deprecated structure would work for python classes in python3? You could have a look at https://stackoverflow.com/questions/922550/how-to-mark-a-global-as-deprecated-in-python for my reference

cmastalli (commented):
Nice work! the code indeed has higher quality.
There is nothing that concerns me, with a tiny exception regarding `calcDiff` in `CostModelControl`.

Again, thanks @proyan for this big effort from your side

proyan:
@nim65s Could you please have a look at why the ROS-CI is failing?
https://pipelines.actions.githubusercontent.com/LJAzhgat4cbGQ3d5xS3ZS5E7MIAIm5ZTSBxhpqmGiHfevgbNif/_apis/pipelines/1/runs/30/signedlogcontent/4?urlExpires=2021-04-28T10%3A01%3A12.4284031Z&urlSigningMethod=HMACV1&urlSignature=8G%2BazK5e2B1cah%2FcPnygtHRQRwY2qD15ksteDQpNcaI%3D

with the error message
```2021-04-28T09:44:22.8517092Z /usr/bin/tar: ../../../../../github: Cannot mkdir: Permission denied```

File: bindings/python/crocoddyl/deprecated.py [hunk 22]
```diff
@@ -30,3 +30,20 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+class DeprecationHelper(object):
+    def __init__(self, new_target, old_name):
+        self.new_target = new_target
+        self.warning_str = f'{old_name} is deprecated: Use {new_target.__name__}'
```
nim65s, line 38:
```suggestion
        self.warning_str = '%s is deprecated: Use %s' % (old_name, new_target.__name__)
```

cmastalli:
> @nim65s Could you please have a look at why the ROS-CI is failing?
> https://pipelines.actions.githubusercontent.com/LJAzhgat4cbGQ3d5xS3ZS5E7MIAIm5ZTSBxhpqmGiHfevgbNif/_apis/pipelines/1/runs/30/signedlogcontent/4?urlExpires=2021-04-28T10%3A01%3A12.4284031Z&urlSigningMethod=HMACV1&urlSignature=8G%2BazK5e2B1cah%2FcPnygtHRQRwY2qD15ksteDQpNcaI%3D
> 
> with the error message
> `2021-04-28T09:44:22.8517092Z /usr/bin/tar: ../../../../../github: Cannot mkdir: Permission denied`

I think @wxmerkt might understand the reason.

bindings/python/crocoddyl/deprecated.py
nim65s: Sorry, but the f-strings are working only for Python >= 3.6

proyan: thanks for the update

include/crocoddyl/core/costs/control.hxx
```diff
@@ -39,50 +49,30 @@ CostModelControlTpl<Scalar>::CostModelControlTpl(boost::shared_ptr<typename Base
...
-                 << "u has wrong dimension (it should be " + std::to_string(nu_) + ")");
-  }
-
-  data->r = u - uref_;
-  activation_->calc(data->activation, data->r);
-  data->cost = data->activation->a_value;
-}
-
 template <typename Scalar>
 void CostModelControlTpl<Scalar>::calcDiff(const boost::shared_ptr<CostDataAbstract>& data,
```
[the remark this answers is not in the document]
proyan: Because Ru is identity for control costs. So the generic formulation isn't the best one here.

wxmerkt:
I dont see failures of the ROS-CI - they passed for the previous commits and the most recent commit is still building

proyan:
btw, this code now uses type aliases, which is a c++11 feature.

https://github.com/cmastalli/crocoddyl/blob/topic/residual/include/crocoddyl/multibody/fwd.hpp#L196

proyan:
> I dont see failures of the ROS-CI - they passed for the previous commits and the most recent commit is still building

It may be because of the force push. If this commit fails to build as well, then perhaps we really do have an issue.

include/crocoddyl/core/costs/control.hxx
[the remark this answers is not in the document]
cmastalli: I see, thx

bindings/python/crocoddyl/multibody/residuals/com-position.cpp
```diff
@@ -0,0 +1,83 @@
...
+namespace python {
+
+void exposeResidualCoMPosition() {
+  bp::register_ptr_to_python<boost::shared_ptr<ResidualModelCoMPosition> >();
+
+  bp::class_<ResidualModelCoMPosition, bp::bases<ResidualModelAbstract> >(
+      "ResidualModelCoMPosition",
+      "This residual function defines the CoM tracking as r = c - cref, with c and cref as the current and reference "
+      "CoM position, respectively.",
+      bp::init<boost::shared_ptr<StateMultibody>, Eigen::Vector3d, int>(
```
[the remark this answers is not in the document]
proyan: I've updated it

proyan (approved):

proyan:
@nim65s @wxmerkt I am surprised that the commit https://github.com/loco-3d/crocoddyl/pull/899/commits/84004cf4d0b628b5820b7f7084ed43dd92fb4a14 was passing in ROS-CI, when the pybinds_costs test case should have been failing, like it did in `gitlab` and `travis`. I've fixed the bug now in https://github.com/loco-3d/crocoddyl/pull/899/commits/579e830ace9fc0c620797a84cdb1828b42eeafd2, but could you please investigate why ROS-CI wasn't giving a failure?

cmastalli:
It would be also nice to have a clear way to inspect when ROS CI Jobs could fail.

@proyan thanks for pushing the latest bit to finalize this large PR ;)

wxmerkt:
I don't think ROS-CI runs unittests yet. I can look into it next week

proyan:
Thanks Wolfgang

On May 1, 2021 4:37:48 PM GMT+02:00, Wolfgang Merkt ***@***.***> wrote:
>I don't think ROS-CI runs unittests yet. I can look into it next week
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/loco-3d/crocoddyl/pull/899#issuecomment-830642800

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #902, opened 2021-01-13
Bugfix: Created properly the data in flt-exp/log activations

cmastalli:
This issue was causing problems in 16.04 jobs.
For some magical reason, this code was passing in the other jobs (@nim65s -- can you explain me why?)

@edantec this PR is important for you as it fixes issues in the flat-log/exp activation models.

### DIFF
--- a/include/crocoddyl/core/activations/quadratic-flat-exp.hpp
+++ b/include/crocoddyl/core/activations/quadratic-flat-exp.hpp
@@ -103,9 +103,8 @@ class ActivationModelQuadFlatExpTpl : public ActivationModelAbstractTpl<_Scalar>
    * @return the activation data
    */
   virtual boost::shared_ptr<ActivationDataAbstract> createData() {
-    boost::shared_ptr<ActivationDataAbstract> data =
-        boost::allocate_shared<ActivationDataAbstract>(Eigen::aligned_allocator<ActivationDataAbstract>(), this);
-    data->Arr.diagonal().fill((Scalar)1.);
+    boost::shared_ptr<Data> data =
+        boost::allocate_shared<Data>(Eigen::aligned_allocator<Data>(), this);
     return data;
   };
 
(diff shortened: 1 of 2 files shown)


include/crocoddyl/core/activations/quadratic-flat-exp.hpp
```diff
@@ -103,9 +103,8 @@ class ActivationModelQuadFlatExpTpl : public ActivationModelAbstractTpl<_Scalar>
    * @return the activation data
    */
   virtual boost::shared_ptr<ActivationDataAbstract> createData() {
-    boost::shared_ptr<ActivationDataAbstract> data =
-        boost::allocate_shared<ActivationDataAbstract>(Eigen::aligned_allocator<ActivationDataAbstract>(), this);
-    data->Arr.diagonal().fill((Scalar)1.);
+    boost::shared_ptr<Data> data =
```
proyan, line 106: there are missing noalias() in the calcdiffs of the two activations. Could you please add them here as well?

proyan (approved):

include/crocoddyl/core/activations/quadratic-flat-exp.hpp
cmastalli: `a1`, `alpha_` are scalars.
Those lines are performing a scalar-vector multiplication.
As far I understand, scalars multiplication don't allocate data; therefore, it is not needed `noalias()`

cmastalli: @proyan do you agree with that?
(I could modify the lines, but I want to be sure that it is really needed)

cmastalli: image

This is comment from Eigen. "Scalar are harmless"
This is why I understand that `noalias()` is not needed in those cases.

proyan:
Yes, scalar vector doesn't need noalias.

Issue #866, opened 2020-10-07
Changing API of costModel.reference


proyan:
Dear Crocoddyl developers, 

It looks like the `get_reference` function in the costmodel is creating a copy of the object in python ([here](https://github.com/loco-3d/crocoddyl/blob/devel/include/crocoddyl/core/cost-base.hxx#L84)) and returning that copy. We don't want that. We need to return an internal reference.

Unfortunately, the returning of internal references is being made complicated by the current design choice, where the cost reference is not a `VectorX` object, but instead a wrapper around this vector (`FrameForce`, `FrameMotion`, `FramePlacement`) etc. 
Thus, we need to somehow imitate a templated virtual function, which I don't feel is needed.

I'm proposing here to change this and simplify the c++ API to make `costmodel.reference` into a vector object, as it was earlier in the  original python version. This would move the frame-id information into the respective cost classes, and remove the wrappers FrameMotion, FrameForce, FrameWrenchCone etc, which are simply duplicate wrappers on pinocchio classes.
Also, this would rename the individual costmodel references (`fref`, `cref`, `Mref`) into a single name reference.

After the depreciation phase, this API change would affect all users from the future releases, and thus, your opinions are welcome here.

jcarpent:
Everything that makes Crocoddyl closer to the Pinocchio API is always, both in terms of maintainability and extensions. I'm in favour of this change.

cmastalli:
Hi @proyan,

Each reference cannot fit into a vector, e.g. SE(3) placement.
Furthermore, the reference are also describing the frame index.
So, if you define a vector (and put everything there) then you will need to convert it to the right object.
Note that this conversion is strictly needed since `calc` and `calcDiff`use it.

On the other hand, having to create an object indeed is not desired, I agree with you.
But again, the vector conversion that you propose doesn't seem to be a better idea.
It also sacrifices the usability for a few nanoseconds.

In short, I believe it is better to keep the current implementation.
IMHO, your solution creates too much user-side complexity for a neglectable runtime improvement.

Carlos

proyan:
Regarding your points, I'll try to go one by one.

> Each reference cannot fit into a vector, e.g. SE(3) placement.

Yes it can.

>  Furthermore, the reference are also describing the frame index. So, if you define a vector (and put everything there) then you will need to convert it to the right object.

I think you are misunderstanding what the issue is. We should not be defining the frame indices in the reference object in the first place. Currently, the API is 
[...]

PR #907, opened 2021-01-26
Included CoP support + improvements in cones (friction & wrench)

cmastalli:
This PR includes a structure that computes the inequality matrix and bounds for a CoP support. The idea is to use this structures in the residual APIs (#899) as we are doing for the friction/wrench cones.

Additionally, it improves the computation of the inequality matrices for the friction and wrench cones (exploits the structure during rotation transformation). There are other modifications regarding the cones, I enlist next:
  1. `update` function (in both cones) do not require to pass any argument in `update` function (deprecated original one)
  2. Wrench cone handles an arbitrary number of friction-cone facets (before it was hard-code -> buggy for an arbitrary number)
  3. Friction cone is constructed with the rotation matrix as we want to encode yaw rotations (deprecate constructor based on surface normal)
  4. Wrench cone models both inner/outer approximations of the cone (before it was only the inner one)
  5. Included documentation of both cones
  6. Extended the set of unit-tests for both cones
  7. Updated the needed version of eigenpy as it solves an important bug for binding 4x6 matrices
  8. Updated the gepetto displayer (it uses the rotation matrix, instead of the surface normal)


cmastalli:
I had to rebase and resolve conflicts. This PR is now ready to be reviewed!

proyan:
Would you agree that FrictionCone, CoPSupport and WrenchCone are subsets, or duplications of each other?

proyan:
In the discussion following https://github.com/loco-3d/crocoddyl/issues/866, it was decided to get rid of the special classes that contain metadata about cost models. Following that, I see that the PR https://github.com/loco-3d/crocoddyl/pull/899 does indeed gets rid of the frame classes. Thank you for that. 
However, it still keeps the FrictionCone, CoPSupport and WrenchCone classes in order to define residuals. Following the discussion that we had, this should be removed as well.

cmastalli:
> However, it still keeps the FrictionCone, CoPSupport and WrenchCone classes in order to define residuals. Following the discussion that we had, this should be removed as well.

This classes are not just containing data (as for the `FrameXXX` ones). I see them equals to the `pinocchio::Motion`. Perhaps we should discuss this topic in our next meeting.

cmastalli:
Further comment, the inequality constraint matrices are updated only when it is called `update()`. This matrix changes according meta parameters such as rotation matrix, inner appr., amount others.

Btw, as far I remember, for instance, we decided to deprecate `FrameFrictionCone` but still keep `FrictionCone`.

cmastalli:
Status:

I have discussed in private with @proyan about his concerns (which are not clearly reflected in this thread). His concern is about embedding both cones (friction and wrench) through an abstraction. He sustains that we should pursue it in this way; however, as I explained him, it affects the performance as the inequality matrices cannot be defined with fixed dimensions.

Further, despite that his suggestion is valid for discussion (reduce the code), it is orthogonal to the objective of this PR. It will also require extensive development: change of API, deprecation, and adaptation of unitt-test and cost functions, which affects significantly my ongoing agenda.

@proyan -- I suggest you to prepare your proposal with clear details. If this is accepted, then you need to proactively provide the solution (unfortunately, I do not have time).
In any case, thank you for raising your point, but, again, I hope you understand that your request is out of the scope of this PR.

In future, I would kindly ask you to only raise points that are within the scope of a PR. This is the most productive way for everyone.

PR #908, opened 2021-02-06
Fix deprecated copy for frames

wxmerkt:
Since we are explicitly declaring the copy constructor for the frames, we need to also explicitly declare the assignment operator. This becomes a warning in more recent versions of gcc/C++ standards.

### DIFF
--- a/include/crocoddyl/core/activations/quadratic-barrier.hpp
+++ b/include/crocoddyl/core/activations/quadratic-barrier.hpp
@@ -2,6 +2,7 @@
 // BSD 3-Clause License
 //
 // Copyright (C) 2019-2020, LAAS-CNRS, University of Edinburgh
+// Copyright (C) 2021, University of Oxford
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
@@ -59,8 +60,15 @@ struct ActivationBoundsTpl {
       beta = Scalar(1.);
     }
   }
-  ActivationBoundsTpl(const ActivationBoundsTpl& bounds) : lb(bounds.lb), ub(bounds.ub), beta(bounds.beta) {}
+  ActivationBoundsTpl(const ActivationBoundsTpl& other) : lb(other.lb), ub(other.ub), beta(other.beta) {}
   ActivationBoundsTpl() : beta(Scalar(1.)) {}
+
+  ActivationBoundsTpl& operator=(const ActivationBoundsTpl& other) {
+    lb = other.lb;
+    ub = other.ub;
+    beta = other.beta;
+    return *this;
+  }
 
   VectorXs lb;
   VectorXs ub;
--- a/include/crocoddyl/core/activations/quadratic-barrier.hpp
+++ b/include/crocoddyl/core/activations/quadratic-barrier.hpp
@@ -1,8 +1,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2019-2020, LAAS-CNRS, University of Edinburgh
-// Copyright (C) 2021, University of Oxford
+// Copyright (C) 2019-2021, LAAS-CNRS, University of Edinburgh, University of Oxford
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
--- a/include/crocoddyl/core/activations/quadratic-barrier.hpp
+++ b/include/crocoddyl/core/activations/quadratic-barrier.hpp
@@ -63,9 +63,11 @@ struct ActivationBoundsTpl {
   ActivationBoundsTpl() : beta(Scalar(1.)) {}
 
   ActivationBoundsTpl& operator=(const ActivationBoundsTpl& other) {
-    lb = other.lb;
-    ub = other.ub;
-    beta = other.beta;
+    if (this != &other) {
+      lb = other.lb;
+      ub = other.ub;
+      beta = other.beta;
+    }
     return *this;
   }
 
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -2,6 +2,7 @@
 // BSD 3-Clause License
 //
 // Copyright (C) 2019-2020, LAAS-CNRS, University of Edinburgh
+// Copyright (C) 2021, University of Oxford
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -424,7 +424,7 @@ template <typename Scalar>
 void ShootingProblemTpl<Scalar>::set_x0(const VectorXs& x0_in) {
   if (x0_in.size() != x0_.size()) {
     throw_pretty("Invalid argument: "
-                 << "invalid size of x0 provided.");
+                 << "invalid size of x0 provided: Expected " << x0_.size() << ", received " << x0_in.size());
   }
   x0_ = x0_in;
 }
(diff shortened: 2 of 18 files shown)


proyan (approved):

proyan:
Cool. Thanks @wxmerkt!

include/crocoddyl/core/activations/quadratic-barrier.hpp
```diff
@@ -2,6 +2,7 @@
 // BSD 3-Clause License
 //
 // Copyright (C) 2019-2020, LAAS-CNRS, University of Edinburgh
+// Copyright (C) 2021, University of Oxford
```
cmastalli, line 5: Would you write this in the same line? (Same for the other files)

This is just to simplify the copyright signature. Indeed, I did this simplification a few PR ago.

include/crocoddyl/core/optctrl/shooting.hxx
```diff
@@ -2,6 +2,7 @@
 // BSD 3-Clause License
 //
 // Copyright (C) 2019-2020, LAAS-CNRS, University of Edinburgh
+// Copyright (C) 2021, University of Oxford
```
cmastalli, line 5: Same here

include/crocoddyl/multibody/frames.hpp
```diff
@@ -3,6 +3,7 @@
 // BSD 3-Clause License
 //
 // Copyright (C) 2019-2020, LAAS-CNRS, University of Edinburgh
+// Copyright (C) 2021, University of Oxford
```
cmastalli, line 6: And here

include/crocoddyl/multibody/friction-cone.hpp
```diff
@@ -2,6 +2,7 @@
 // BSD 3-Clause License
 //
 // Copyright (C) 2019-2021, University of Edinburgh
+// Copyright (C) 2021, University of Oxford
```
cmastalli, line 5: Here

include/crocoddyl/multibody/friction-cone.hxx
```diff
@@ -2,6 +2,7 @@
 // BSD 3-Clause License
 //
 // Copyright (C) 2019-2021, University of Edinburgh
+// Copyright (C) 2021, University of Oxford
```
cmastalli, line 5: Here

include/crocoddyl/multibody/wrench-cone.hpp
```diff
@@ -2,6 +2,7 @@
 // BSD 3-Clause License
 //
 // Copyright (C) 2020-2021, University of Edinburgh
+// Copyright (C) 2021, University of Oxford
```
cmastalli, line 5: Here

include/crocoddyl/multibody/wrench-cone.hxx
```diff
@@ -2,6 +2,7 @@
 // BSD 3-Clause License
 //
 // Copyright (C) 2020-2021, University of Edinburgh
+// Copyright (C) 2021, University of Oxford
```
cmastalli, line 5: Here

cmastalli (changes requested):
Thank you so much @wxmerkt.

I have two main requests:
 1. Could you add unittests for the FrictionCone and WrenchCone in the 'test_constructor' functions? (To check the new operator introduced in this PR)
  2. Could you add this constructor to the CoMSupport class? (Including unittest as in 1)

There is also another minimal request (yet not crucial). It would be nice if you add the copyright in the same line with other copyholders (just to simplify the code).

wxmerkt:
Thanks for the feedback, I've addressed the comments and added the unit tests. I've also added code to explicitly ignore deprecation warnings in the Python bindings. The project is now free of warnings.

cmastalli (approved):
Thank you so much! also for the ignore deprecation macros!

@proyan --- thank you so much for also reviewing the code.

proyan:
@wxmerkt The tests are failing,
```
14/42 Test #14: test_diff_actions .............................***Failed    0.94 sec

Running test_DifferentialActionModelLQR

Running test_DifferentialActionModelLQRDriftFree

Running test_DifferentialActionModelFreeFwdDynamics_TalosArm

Running test_DifferentialActionModelFreeFwdDynamics_TalosArm_Squashed

Running test_DifferentialActionModelContactFwdDynamics_TalosArm

Running test_DifferentialActionModelContactFwdDynamics_HyQ

Running test_DifferentialActionModelContactFwdDynamics_Talos

Running test_DifferentialActionModelContactFwdDynamicsWithFriction_TalosArm

Running test_DifferentialActionModelContactFwdDynamicsWithFriction_HyQ

Running test_DifferentialActionModelContactFwdDynamicsWithFriction_Talos

Running 50 test cases...

/home/travis/build/loco-3d/crocoddyl/unittest/test_diff_actions.cpp(127): error in "boost::bind(&test_partial_derivatives_against_numdiff, action_type)": check (data->Fx - data_num_diff->Fx).isMuchSmallerThan(1.0, tol) failed

*** 1 failure detected in test suite "Master Test Suite"
```
Could you please take a look?

jcarpent:
You should use isZero instead of isMuchSmallerThan.

cmastalli:
@wxmerkt  -- I should probably handle the issue with the unit-tests as this is not affected by the work on this PR.
The only issue is that I could not work now (but asap) on it.

cmastalli:
> I am aware that likely there is an issue with my code in this PR - it does not yet handle self-assignment checks for the assignment operator (`if (this != &other)`).

 I'll let you solve this issue. Then, I'll fix any issue in the unit-test (if they are still happening after handling this)

> You should use isZero instead of isMuchSmallerThan.

@jcarpent -- it makes sense, and I happy to handle this in another PR as all the unit-test is written in this way.
Any suggestion regarding the tolerance? using the same one?

wxmerkt:
I did some more digging and switched back to gcc - the test failure is already present on `devel`. I added an output of the norm between the Fx and num-diff-Fx as:
```cpp
std::cout << action_type << " - DIFF - " <<  (data->Fx - data_num_diff->Fx).norm() << std::endl;
```
in `test_diff_actions.cpp`, L127.

The output we get is then:
```
DifferentialActionModelContactFwdDynamics_Talos - DIFF - 0.00111288
crocoddyl/unittest/test_diff_actions.cpp(128): error: in "test_DifferentialActionModelContactFwdDynamics_Talos/boost__bind(&test_partial_derivatives_against_numdiff, action_type)": check (data->Fx - data_num_diff->Fx).isMuchSmallerThan(1.0, tol) has failed
```

I'll get on fixing the self-assignment issue in the PR and update it with that.

jcarpent:
The tolerance is `tol = sqrt(eps)` where eps is the increment of the finite diff. Please see Pinocchio for further examples.

wxmerkt:
This PR should now be ready:
- I went back to `devel` and made sure I can get the unittests to pass there with gcc 9.3.0 in Debug mode: This required a modification to the tolerance for Fx, Fu in `test_diff_actions`. I based the general tolerance on `sqrt(eps)`, but had to adapt for Fx and Fu.
- I added the self-assignment check for the assignment operator. I also found another missing copy constructor in `CoPSupport`.

unittest/test_friction_cone.cpp
```diff
@@ -52,6 +52,37 @@ void test_constructor() {
   BOOST_CHECK(static_cast<std::size_t>(cone.get_A().rows()) == nf + 1);
   BOOST_CHECK(static_cast<std::size_t>(cone.get_lb().size()) == nf + 1);
   BOOST_CHECK(static_cast<std::size_t>(cone.get_ub().size()) == nf + 1);
+
+  // Create the friction cone from a reference
+  {
+    crocoddyl::FrictionCone cone_reference(cone);
+
+    BOOST_CHECK(cone.get_nf() == cone_reference.get_nf());
+    BOOST_CHECK((cone.get_A() - cone_reference.get_A()).isMuchSmallerThan(1.0, 1e-9));
```
jcarpent, line 61: `isMuchSmallerThan` is not very elegant.

wxmerkt:
@nim65s Gepgitlab is failing due to "out of harddrive space"

cmastalli:
@wxmerkt and @jcarpent  -- I have fixed the unit-test request in #910. However, the contact-dynamics tests do not pass with `sqrt(eps)`. Instead, we need to use a factor of `10`, i.e., `10*sqrt(eps)`.

I suggest to do not work more on the unittests in this PR.

wxmerkt:
This PR is complete and can be merged. The unit tests have also been fixed. I suggest to merge this one before #910

PR #910, opened 2021-02-15
Used isZero in unit-tests

cmastalli:
This PR replace `isMuchSmallerThan` by `isZero` in all the unit-tests. Additionally, it changes the tolerance in numdiff tests by `10 * sqrt(eps)` where `eps` is the numdiff disturbance. We need to include `10` as a factor since the Talos contact-dynamics unit-tests needed. Finally, this PR increases the tolerance by `1e-9` in some extra unit-tests.

### DIFF
--- a/unittest/test_actions.cpp
+++ b/unittest/test_actions.cpp
@@ -88,9 +88,9 @@ void test_partial_derivatives_against_numdiff(ActionModelTypes::Type action_mode
   model_num_diff.calcDiff(data_num_diff, x, u);
 
   // Checking the partial derivatives against NumDiff
-  double tol = NUMDIFF_MODIFIER * sqrt(model_num_diff.get_disturbance());
-  BOOST_CHECK((data->Fx - data_num_diff->Fx).isZero(tol));
-  BOOST_CHECK((data->Fu - data_num_diff->Fu).isZero(tol));
+  double tol = sqrt(model_num_diff.get_disturbance());
+  BOOST_CHECK((data->Fx - data_num_diff->Fx).isZero(NUMDIFF_MODIFIER * tol));
+  BOOST_CHECK((data->Fu - data_num_diff->Fu).isZero(NUMDIFF_MODIFIER * tol));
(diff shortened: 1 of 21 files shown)


unittest/test_actions.cpp
```diff
@@ -88,19 +88,19 @@ void test_partial_derivatives_against_numdiff(ActionModelTypes::Type action_mode
   model_num_diff.calcDiff(data_num_diff, x, u);
 
   // Checking the partial derivatives against NumDiff
-  double tol = NUMDIFF_MODIFIER * model_num_diff.get_disturbance();
-  BOOST_CHECK((data->Fx - data_num_diff->Fx).isMuchSmallerThan(1.0, tol));
-  BOOST_CHECK((data->Fu - data_num_diff->Fu).isMuchSmallerThan(1.0, tol));
-  BOOST_CHECK((data->Lx - data_num_diff->Lx).isMuchSmallerThan(1.0, tol));
-  BOOST_CHECK((data->Lu - data_num_diff->Lu).isMuchSmallerThan(1.0, tol));
+  double tol = NUMDIFF_MODIFIER * sqrt(model_num_diff.get_disturbance());
```
wxmerkt, line 91: `NUMDIFF_MODIFIER` is 3e4. Note I used `sqrt(epsilon)` in #908, and only increased tol by a factor of 10 where required (Fx, Fu) and left it tighter for the other ones.

cmastalli: Yes, I have observed this as well.
I wanted to standardize the code; however, I might be better to only use 10 for the contact-fwd tests (as you pointed out).

Let me revert the other cases. I think this is the most tuned approach! Thank you for raising this point :)

cmastalli: done.

Let me know if you have other comments!

proyan (approved):

proyan:
Great! thanks Carlos for this PR. Is this ready to merge?

cmastalli:
> Great! thanks Carlos for this PR. Is this ready to merge?

Yes, it is. I am merging it now.

Thanks @proyan and @wxmerkt for the code review :)

Issue #895, opened 2020-12-28
ActuationModelFull with FreeFlyer


hammoudbilal:
Hi, 

I'm trying to create a simple demo for a fully actuated sphere (i.e. my model is only an SE3 joint with some mass) just to move it around and control the orientation too, however ActuationModelFull prevents this
 `Invalid argument: the first joint cannot be free-flyer`
is there a workout around this issue ?

proyan:
Hi BIlal,

This is a bug that should not be part of the code. I think this was done to make sure a user does not use ActuationModelFull when controlling a floating base robot. But perhaps we we need to trust our users a bit more. 

We are already using the StateMultibody in our ActuationModel class. So the only thing to do is to remove this throw. Do you think you can make a PR with this, and a unittest of your example in `unittest/factory/actuation`?

hammoudbilal:
Hi @proyan 

alright, I will leave the issue open till i get to doing it sometime later this week! thanks

cmastalli:
solved in #912

PR #912, opened 2021-02-18
Enable free-flyer joint in full actuation model #895

cmastalli:
This PR handles #895 and contains the following actions:
  1. Used state-base (instead of state multi-body) as we don't need to check if the system is not free-flyer.
  2. Included more actuation unit-tests
  3. Removed deprecated headers for control cost


cmastalli:
I will also need to move these files to the core folder. But I will handle this later as I need to merge first #911.

proyan (approved):

proyan:
Thanks Carlos for handling this issue

PR #918, opened 2021-03-01
Expose ActionModelUnicycle::dt_

jmirabel:


cmastalli (approved):
Thanks @jmirabel :)

include/crocoddyl/core/actions/unicycle.hxx
```diff
@@ -97,4 +97,16 @@ void ActionModelUnicycleTpl<Scalar>::set_cost_weights(const typename MathBase::V
   cost_weights_ = weights;
 }
 
+template <typename Scalar>
+const Scalar& ActionModelUnicycleTpl<Scalar>::get_dt() const {
+  return dt_;
+}
+
+template <typename Scalar>
+void ActionModelUnicycleTpl<Scalar>::set_dt(const Scalar& dt) {
```
wxmerkt, line 106: Do we want const-ref on primitive types? Same for the return in `get_dt`

cmastalli: @wxmerkt - thanks for spot it.
We need to also adapt the Python bindings

cmastalli:
@jmirabel - the format needs to be adjusted.
This is a very simple task as you need to run a command explained here: https://github.com/loco-3d/crocoddyl/wiki/Developer-Guide

include/crocoddyl/core/actions/unicycle.hxx
jmirabel: You want primitive types to be copied ?

wxmerkt: Yes - for primitive types it's often as expensive to create the reference and pass and resolve it than to copy its value. There was some discussion on this here in the past as well and all const-ref for primitives was changed to pass-by-value

jmirabel: Ok, thanks for the explanation. I'll update the code.

include/crocoddyl/core/actions/unicycle.hpp
```diff
@@ -42,6 +42,9 @@ class ActionModelUnicycleTpl : public ActionModelAbstractTpl<_Scalar> {
   const Vector2s& get_cost_weights() const;
   void set_cost_weights(const Vector2s& weights);
 
+  Scalar get_dt() const;
+  void set_dt(Scalar dt);
```
cmastalli, line 46: It would be good if this is `const`, i.e.,
```cpp
void set_dt(const Scalar dt);
```

I apologize for not being clear enough

include/crocoddyl/core/actions/unicycle.hxx
jmirabel: Do you have the link to the discussion ?

wxmerkt: - Mostly here in comments by myself and Justin: https://github.com/loco-3d/crocoddyl/pull/642
- More uses of it were removed in https://github.com/loco-3d/crocoddyl/pull/893

cmastalli (approved):
Thanks again @jmirabel (and also @wxmerkt for reviewing this PR)

PR #920, opened 2021-03-05
Multi-threading support was not running the desired number of threads

cmastalli:
This PR fixes an issue regarding the parallel procedure implemented for retrieving the derivatives in solvers. This issue was spotted by @wxmerkt yesterday, when he observed that the multi-threading support was using all the cores (making conflict with other components such as the Gazebo simulator).

Additionally, the PR includes a few extra parallelization routines.

@wxmerkt -- please try this PR from your side ;)

### DIFF
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -8,7 +8,6 @@
 
 #ifdef CROCODDYL_WITH_MULTITHREADING
 #include <omp.h>
-#define NUM_THREADS CROCODDYL_WITH_NTHREADS
 #endif  // CROCODDYL_WITH_MULTITHREADING
 
 namespace crocoddyl {
@@ -141,7 +140,7 @@ Scalar ShootingProblemTpl<Scalar>::calc(const std::vector<VectorXs>& xs, const s
   }
 
 #ifdef CROCODDYL_WITH_MULTITHREADING
-#pragma omp parallel for num_threads(NUM_THREADS)
+#pragma omp parallel for num_threads(CROCODDYL_WITH_NTHREADS)
 #endif
   for (std::size_t i = 0; i < T_; ++i) {
     const std::size_t nu = running_models_[i]->get_nu();
@@ -176,7 +175,7 @@ Scalar ShootingProblemTpl<Scalar>::calcDiff(const std::vector<VectorXs>& xs, con
   }
 
 #ifdef CROCODDYL_WITH_MULTITHREADING
-#pragma omp parallel for num_threads(NUM_THREADS)
+#pragma omp parallel for num_threads(CROCODDYL_WITH_NTHREADS)
 #endif
   for (std::size_t i = 0; i < T_; ++i) {
     if (running_models_[i]->get_nu() != 0) {
@@ -249,7 +248,7 @@ void ShootingProblemTpl<Scalar>::quasiStatic(std::vector<VectorXs>& us, const st
   }
 
 #ifdef CROCODDYL_WITH_MULTITHREADING
-#pragma omp parallel for num_threads(NUM_THREADS)
+#pragma omp parallel for num_threads(CROCODDYL_WITH_NTHREADS)
 #endif
   for (std::size_t i = 0; i < T_; ++i) {
     const std::size_t nu = running_models_[i]->get_nu();
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -394,9 +394,13 @@ const typename MathBaseTpl<Scalar>::VectorXs& ShootingProblemTpl<Scalar>::get_x0
 
 template <typename Scalar>
 void ShootingProblemTpl<Scalar>::allocateData() {
+  running_datas_.resize(T_);
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#pragma omp parallel for num_threads(NUM_THREADS)
+#endif
   for (std::size_t i = 0; i < T_; ++i) {
     const boost::shared_ptr<ActionModelAbstract>& model = running_models_[i];
-    running_datas_.push_back(model->createData());
+    running_datas_[i] = model->createData();
   }
   terminal_data_ = terminal_model_->createData();
 }
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -141,7 +141,7 @@ Scalar ShootingProblemTpl<Scalar>::calc(const std::vector<VectorXs>& xs, const s
   }
 
 #ifdef CROCODDYL_WITH_MULTITHREADING
-#pragma omp parallel for
+#pragma omp parallel for num_threads(NUM_THREADS)
 #endif
   for (std::size_t i = 0; i < T_; ++i) {
     const std::size_t nu = running_models_[i]->get_nu();
@@ -173,7 +173,7 @@ Scalar ShootingProblemTpl<Scalar>::calcDiff(const std::vector<VectorXs>& xs, con
   }
 
 #ifdef CROCODDYL_WITH_MULTITHREADING
-#pragma omp parallel for
+#pragma omp parallel for num_threads(NUM_THREADS)
 #endif
   for (std::size_t i = 0; i < T_; ++i) {
     if (running_models_[i]->get_nu() != 0) {
@@ -243,7 +243,7 @@ void ShootingProblemTpl<Scalar>::quasiStatic(std::vector<VectorXs>& us, const st
   }
 
 #ifdef CROCODDYL_WITH_MULTITHREADING
-#pragma omp parallel for
+#pragma omp parallel for num_threads(NUM_THREADS)
 #endif
   for (std::size_t i = 0; i < T_; ++i) {
     const std::size_t nu = running_models_[i]->get_nu();
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -155,7 +155,7 @@ Scalar ShootingProblemTpl<Scalar>::calc(const std::vector<VectorXs>& xs, const s
 
   cost_ = Scalar(0.);
 #ifdef CROCODDYL_WITH_MULTITHREADING
-#pragma omp simd reduction(+:cost_)
+#pragma omp simd reduction(+ : cost_)
 #endif
   for (std::size_t i = 0; i < T_; ++i) {
     cost_ += running_datas_[i]->cost;
@@ -190,7 +190,7 @@ Scalar ShootingProblemTpl<Scalar>::calcDiff(const std::vector<VectorXs>& xs, con
 
   cost_ = Scalar(0.);
 #ifdef CROCODDYL_WITH_MULTITHREADING
-#pragma omp simd reduction(+:cost_)
+#pragma omp simd reduction(+ : cost_)
 #endif
   for (std::size_t i = 0; i < T_; ++i) {
     cost_ += running_datas_[i]->cost;
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -154,6 +154,9 @@ Scalar ShootingProblemTpl<Scalar>::calc(const std::vector<VectorXs>& xs, const s
   terminal_model_->calc(terminal_data_, xs.back());
 
   cost_ = Scalar(0.);
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#pragma omp simd reduction(+:cost_)
+#endif
   for (std::size_t i = 0; i < T_; ++i) {
     cost_ += running_datas_[i]->cost;
   }
@@ -186,6 +189,9 @@ Scalar ShootingProblemTpl<Scalar>::calcDiff(const std::vector<VectorXs>& xs, con
   terminal_model_->calcDiff(terminal_data_, xs.back());
 
   cost_ = Scalar(0.);
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#pragma omp simd reduction(+:cost_)
+#endif
   for (std::size_t i = 0; i < T_; ++i) {
     cost_ += running_datas_[i]->cost;
   }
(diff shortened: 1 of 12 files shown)


src/core/solvers/box-ddp.cpp
```diff
@@ -36,6 +41,9 @@ void SolverBoxDDP::allocateData() {
   const std::size_t T = problem_->get_T();
   Quu_inv_.resize(T);
   const std::size_t nu = problem_->get_nu_max();
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#pragma omp parallel for num_threads(NUM_THREADS)
```
wxmerkt, line 45: Does this have a practical benefit? This requires spawning & closing threads for allocating a few matrices (whereas the other parallelisations focused on expensive operations)

src/core/solvers/box-fddp.cpp
```diff
@@ -36,6 +41,9 @@ void SolverBoxFDDP::allocateData() {
   const std::size_t T = problem_->get_T();
   Quu_inv_.resize(T);
   const std::size_t nu = problem_->get_nu_max();
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#pragma omp parallel for num_threads(NUM_THREADS)
```
wxmerkt, line 45: As above, does this provide a benefit?

src/core/solvers/ddp.cpp
```diff
@@ -172,6 +181,9 @@ double SolverDDP::calcDiff() {
     const std::size_t T = problem_->get_T();
     const std::vector<boost::shared_ptr<ActionModelAbstract> >& models = problem_->get_runningModels();
     const std::vector<boost::shared_ptr<ActionDataAbstract> >& datas = problem_->get_runningDatas();
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#pragma omp parallel for num_threads(NUM_THREADS)
```
wxmerkt, line 185: Does `could_be_feasible` here get reduced correctly as it's shared (not declared private) across threads?

src/core/solvers/ddp.cpp
```diff
@@ -369,6 +381,9 @@ void SolverDDP::allocateData() {
   const std::size_t ndx = problem_->get_ndx();
   const std::size_t nu = problem_->get_nu_max();
   const std::vector<boost::shared_ptr<ActionModelAbstract> >& models = problem_->get_runningModels();
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#pragma omp parallel for num_threads(NUM_THREADS)
```
wxmerkt, line 385: This here allocates a lot more than the two cases above, so there might be a benefit. I guess there is?

include/crocoddyl/core/optctrl/shooting.hxx
```diff
@@ -141,7 +141,7 @@ Scalar ShootingProblemTpl<Scalar>::calc(const std::vector<VectorXs>& xs, const s
   }
 
 #ifdef CROCODDYL_WITH_MULTITHREADING
-#pragma omp parallel for
+#pragma omp parallel for num_threads(NUM_THREADS)
```
wxmerkt, line 144: Nice, well spotted :-).

wxmerkt (commented):
Thank you Carlos - this fixes the observed issue. I left comments for clarification of understanding, e.g. if parallelising allocation makes a difference?

src/core/solvers/box-ddp.cpp
proyan: I agree with Wolf here. It doesn't make sense to use multithreading at memory allocation stage.

include/crocoddyl/core/optctrl/shooting.hxx
```diff
@@ -154,6 +154,9 @@ Scalar ShootingProblemTpl<Scalar>::calc(const std::vector<VectorXs>& xs, const s
   terminal_model_->calc(terminal_data_, xs.back());
 
   cost_ = Scalar(0.);
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#pragma omp simd reduction(+ : cost_)
```
proyan, line 158: Could you please point me to a reference where I could see the documentation of simd reduction clause?

src/core/solvers/ddp.cpp
proyan: We can remove the `if(could_be_feasible)` from inside the `for loop`, and do the check separately afterwards. The multithreaded loop would then only handle the statement `model->get_state()->diff(xs_[t + 1], d->xnext, fs_[t + 1]);
`
@cmastalli could you do these changes here?

src/core/solvers/ddp.cpp
proyan: Since there is no additional computational benefit, I think we can keep the memory allocation without multithreading everywhere.

include/crocoddyl/core/optctrl/shooting.hxx
```diff
@@ -388,9 +394,13 @@ const typename MathBaseTpl<Scalar>::VectorXs& ShootingProblemTpl<Scalar>::get_x0
...
 template <typename Scalar>
 void ShootingProblemTpl<Scalar>::allocateData() {
+  running_datas_.resize(T_);
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#pragma omp parallel for num_threads(NUM_THREADS)
+#endif
   for (std::size_t i = 0; i < T_; ++i) {
     const boost::shared_ptr<ActionModelAbstract>& model = running_models_[i];
-    running_datas_.push_back(model->createData());
+    running_datas_[i] = model->createData();
```
proyan, line 403: Same comment about memory allocation and multithreading here

proyan (changes requested):

include/crocoddyl/core/optctrl/shooting.hxx
wxmerkt: - https://docs.microsoft.com/en-us/cpp/parallel/openmp/openmp-simd?view=msvc-160
- https://www.ibm.com/support/knowledgecenter/en/SSXVZZ_16.1.0/com.ibm.xlcpp161.lelinux.doc/compiler_ref/prag_omp_simd.html

src/core/solvers/box-ddp.cpp
cmastalli: I removed the multi-threading routine in all cases where we allocate data

src/core/solvers/ddp.cpp
cmastalli: Done!

include/crocoddyl/core/optctrl/shooting.hxx
cmastalli: I see a little of improvement with these routines.
Perhaps, you could try it for double checking this.

jcarpent: You cannot use SIMD here, this is not aligned in memory the cost.

src/core/solvers/box-ddp.cpp
```diff
@@ -1,14 +1,19 @@
...
+// Copyright (C) 2019-2021, CNRS-LAAS, University of Edinburgh
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
 #include <iostream>
-#include "crocoddyl/core/utils/exception.hpp"
+#ifdef CROCODDYL_WITH_MULTITHREADING
+#include <omp.h>
+#define NUM_THREADS CROCODDYL_WITH_NTHREADS
```
jcarpent: I think NUM_THREADS is too generic and can conflict with other libs. Always use CROCODDYL prefix

jcarpent (commented):
My general overview of multithreading is: use it only when necessary, not for all the operations because it might slow down other parallel applications for nothing.

include/crocoddyl/core/optctrl/shooting.hxx
proyan: From what I understand after reading the documentation is, SIMD directives parallelize the execution intra-thread, by using vectorization of the associated variables. 

What I don't understand is, how this worked for a scalar where such a vectorization isn't possible. Could you please post the timings that you saw here?

src/core/solvers/box-ddp.cpp
proyan: Do we need to redefine `NUM_THREADS` here? We can directly use `CROCODDYL_WITH_NTHREADS` and skip re-definition of the macro.

include/crocoddyl/core/optctrl/shooting.hxx
cmastalli: > You cannot use SIMD here, this is not aligned in memory the cost.

Cost is an scalar, why do you think we cannot use it? (I did not observe any issue despite that I spend all day running this code with a MPC code.

> What I don't understand is, how this worked for a scalar where such a vectorization isn't possible.

This is indeed not explained, but I got inspiration by this tutorial: https://bisqwit.iki.fi/story/howto/openmp/#SimdConstructOpenmp%204%200 (see The reduction clause).

> Could you please post the timings that you saw here?

I observed a time reduction of 0.02-0.09 ms for a single iteration of the FDDP solver.
I run this benchmark for the ANYmal case with 8 threads.

src/core/solvers/box-ddp.cpp
cmastalli: Yes, indeed, let's use `CROCODDYL_WITH_NTHREADS`. However, I was thinking that it might be useful to create a mechanism were users can change the number of threads per object, e.g.,
```python
problem = crocoddyl.ShootingProblem(x0, runningModels, terminalModel)
problem.set_nthreads(10) # otherwise use CROCODDYL_WITH_NTHREADS
solver = crocoddyl.SolverDDP(problem)
solver.set_nthreads(10) # same here
```

I am keen to implement this approach in this PR. What do you think?

wxmerkt: I had this in preparation as a PR to set/get the threads for shooting problem (ifdef'd wrapper around `omp_set_num_threads`) - I'll let you handle it then or happy to add to this PR.

cmastalli: Removed `NUM_THREADS` as requested by @jcarpent and @proyan 

> I had this in preparation as a PR to set/get the threads for shooting problem (ifdef'd wrapper around `omp_set_num_threads`) - I'll let you handle it then or happy to add to this PR.

Yes, let me handle it. I had started to work on it. Soon I will push the code :)
Thanks in any case!

cmastalli: I have pushed the code that allows user to modify the nthreads.
I will close this issue as I believe it is resolved :)

include/crocoddyl/core/optctrl/shooting.hxx
jcarpent: > Cost is a scalar, why do you think we cannot use it? (I did not observe any runtime issue despite that I spend all day running this code with a MPC code.

The vectorisation works for 2D vectors and more.

cmastalli: > The vectorisation works for 2D vectors and more.

Yes, that's correct! But in this case, we still have multiple data.
And this macro works for such cases as described in the this link:  https://bisqwit.iki.fi/story/howto/openmp/#SimdConstructOpenmp%204%200

include/crocoddyl/core/optctrl/shooting.hxx
```diff
@@ -470,6 +482,21 @@ void ShootingProblemTpl<Scalar>::set_terminalModel(boost::shared_ptr<ActionModel
   terminal_data_ = terminal_model_->createData();
 }
 
+template <typename Scalar>
```
proyan, line 485: Thanks for this addition.
Please expose `set_nthreads`,`get_nthreads` and the associated bindings only when `CROCODDYL_WITH_MULTITHREADING` is defined.

proyan (commented):
Thanks Carlos for exposing `nthreads` to the user.
I don't like the fact that there are two separate definitions of nthreads in solver class and shooting problem class. It would be better to remove one definition. I suggest that we use only the definition in solver.

include/crocoddyl/core/optctrl/shooting.hxx
cmastalli: > Please expose `set_nthreads`,`get_nthreads` and the associated bindings only when `CROCODDYL_WITH_MULTITHREADING` is defined.

I though about this option but I did convince myself that it is not good approach.
For instance, a Python code that uses `nthreads` will be broken if the user doesn't compile crocoddyl with multi-threading support. To avoid that, I still generate the bindings but prints a warning message that tells user *this is useless*.

@proyan  -- after this explanation, do you still think it is better to generate the bindings only when has been defined `CROCODDYL_WITH_MULTITHREADING`?

what is the point of view of others (e.g., @jcarpent @nim65s @wxmerkt )?

wxmerkt: My approach would have been the same as what you included in this PR to avoid breaking Python code but making it clear that it won't change behaviour.

include/crocoddyl/core/optctrl/shooting.hxx
proyan: Since there is a minor performance improvement, it would be worthwhile to keep the simd construct here. Thanks for this addition carlos. I'll resolve this discussion, please feel free to re-open it if some issue still remains.

proyan:
For me the PR is good to go. The last thing remaining is to remove the duplicate definitions of `nthreads` in shooting problem, and use only the `nthreads` from solver. Other than that, it is ready to merge

cmastalli:
> For me the PR is good to go. The last thing remaining is to remove the duplicate definitions of `nthreads` in shooting problem, and use only the `nthreads` from solver. Other than that, it is ready to merge

I also think it is better (i.e., more flexible) to have different `nthreads`. With this, the user can enable a `Ns` number of threads for `simd` (or potentially forward and backward passes), and `Np` number of threads for computing derivatives, rollouts, quasiStatic, etc.

What is your final though about this?

proyan:
I don't see any reason for separating the two. It is better to have a simpler API for a user. Multiple definitions of the same feature is an exposure that would provoke more faqs than user benefit.

cmastalli:
> I don't see any reason for separating the two. It is better to have a simpler API for a user. Multiple definitions of the same feature is an exposure that would provoke more faqs than user benefit.

Ok, we can do that.
However, we still need to keep `nthreads_` internally in the shooting problem class.
There are two options that I see:
  1. Remove `SolverAbstract::nthreads_` and use `Shooting::nthreads_` in each solver routine.
  2. Keep both `nthreads_` (for solver and shooting problem) and then call `ShootingProblem::set_nthreads` when we call `SolverAbstract::set_nthreads`.

The option 1 seems to be more aligned with what you prefer to do.
The option 2 also makes the work that you request, but still give the freedom to have different `nthreads` (for advanced users). But this option is also a bit more complex than current one.

proyan:
Option 1 sounds good to me

cmastalli:
> Option 1 sounds good to me

done!

cmastalli:
Thank everyone for reviewing this PR

PR #930, opened 2021-04-22
Registered shared pointers of model objects (Python bindings) + state-multibody documentation

cmastalli:
@PepMS -- this is important for you.
Now you don't need to register Crocoddyl stuffs in your MPC library.


proyan (approved):

proyan:
THanks Carlos for the update

PR #932, opened 2021-04-23
ostream operators, MeshcatVisualizer fix, removal of ineffective SIMD reduction

wxmerkt:
This PR includes:
1. `operator<<` for shooting problem, action model, and differential action model to enable `print(ddp.problem)` or `print(ddp.running_models[t])` which is useful in debugging to get information on the problem
2. Removes the ineffective OpenMP simd instructions previously pointed out by Justin and also notified by the clang compiler as not optimised
3. Fixes the `MeshcatVisualizer` when using the latest Pinocchio devel branch (not sure when it broke)

### DIFF
--- a/include/crocoddyl/core/action-base.hxx
+++ b/include/crocoddyl/core/action-base.hxx
@@ -1,10 +1,14 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2019-2020, LAAS-CNRS, University of Edinburgh
+// Copyright (C) 2019-2021, LAAS-CNRS, University of Edinburgh, University of Oxford
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
+
+#include <iostream>                 // std::ostream
+#include <typeinfo>                 // typeid()
+#include <boost/core/demangle.hpp>  // boost::core::demangle
 
 #include "crocoddyl/core/utils/exception.hpp"
 
@@ -143,6 +147,12 @@ void ActionModelAbstractTpl<Scalar>::set_u_ub(const VectorXs& u_ub) {
 template <typename Scalar>
 void ActionModelAbstractTpl<Scalar>::update_has_control_limits() {
   has_control_limits_ = isfinite(u_lb_.array()).any() && isfinite(u_ub_.array()).any();
+}
+
+template <typename Scalar>
+std::ostream& operator<<(std::ostream& os, const ActionModelAbstractTpl<Scalar>& action_model) {
+  os << "ActionModel type '" << boost::core::demangle(typeid(action_model).name()) << "'" << action_model;
+  return os;
 }
 
 }  // namespace crocoddyl
--- a/include/crocoddyl/core/action-base.hxx
+++ b/include/crocoddyl/core/action-base.hxx
@@ -6,9 +6,9 @@
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
-#include <iostream>                 // std::ostream
-#include <typeinfo>                 // typeid()
-#include <boost/core/demangle.hpp>  // boost::core::demangle
+#include <iostream>
+#include <typeinfo>
+#include <boost/core/demangle.hpp>
 
 #include "crocoddyl/core/utils/exception.hpp"
 
--- a/include/crocoddyl/core/action-base.hxx
+++ b/include/crocoddyl/core/action-base.hxx
@@ -151,7 +151,7 @@ void ActionModelAbstractTpl<Scalar>::update_has_control_limits() {
 
 template <typename Scalar>
 std::ostream& operator<<(std::ostream& os, const ActionModelAbstractTpl<Scalar>& action_model) {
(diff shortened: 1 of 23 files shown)


include/crocoddyl/core/action-base.hxx
```diff
@@ -1,11 +1,15 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2019-2020, LAAS-CNRS, University of Edinburgh
+// Copyright (C) 2019-2021, LAAS-CNRS, University of Edinburgh, University of Oxford
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
+#include <iostream>                 // std::ostream
```
cmastalli, line 9: Could you remove the commented lines?

include/crocoddyl/core/diff-action-base.hxx
```diff
@@ -1,11 +1,15 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2019-2020, LAAS-CNRS, The University of Edinburgh
+// Copyright (C) 2019-2021, LAAS-CNRS, The University of Edinburgh, University of Oxford
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
+#include <iostream>                 // std::ostream
```
cmastalli, line 9: Same here

include/crocoddyl/core/integrator/euler.hxx
```diff
@@ -1,12 +1,14 @@
...
 // BSD 3-Clause License
 //
-// Copyright (C) 2019-2020, LAAS-CNRS, University of Edinburgh
+// Copyright (C) 2019-2021, LAAS-CNRS, University of Edinburgh, University of Oxford
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
-#include <iostream>
+#include <iostream>                 // std::ostream
```
cmastalli, line 9: And here

include/crocoddyl/core/optctrl/shooting.hpp
```diff
@@ -235,6 +235,12 @@ class ShootingProblemTpl {
    */
   std::size_t get_nthreads() const;
 
+  /**
+   * @brief Print information on the ShootingProblem
```
cmastalli, line 239: I think to get a proper hyperlink in doxygen we should change ShootingProblem by 'ShootingProblem'

include/crocoddyl/core/optctrl/shooting.hxx
```diff
@@ -520,11 +521,24 @@ std::size_t ShootingProblemTpl<Scalar>::get_nu_max() const {
...
 #endif
   return nthreads_;
 }
 
+template <typename Scalar>
+std::ostream& operator<<(std::ostream& os, const ShootingProblemTpl<Scalar>& problem) {
+  os << "ShootingProblem (T=" << problem.get_T() << ", nx=" << problem.get_nx() << ", ndx=" << problem.get_ndx()
+     << ", nu_max=" << problem.get_nu_max() << ") " << std::endl;
+  os << "  Models:" << std::endl;
+  auto runningModels = problem.get_runningModels();
```
cmastalli, line 536: Please do not use 'auto' as it is not supported by c++98 (we did not decide yet to move to c++11)

cmastalli (changes requested):
Thank your for this work. It is quite clean and nice! I just left a few comments, and the most important point is about 'auto' keyword.

Regarding the displayers, we should create unittests. Feel free to include them in this PR (or create an issue if you don't have bandwidth now) :)

include/crocoddyl/core/optctrl/shooting.hxx
wxmerkt: Thanks for the feedback - I'll make the amendments throughout. With 16.04 support ending next month it may be worth discussing moving to a newer C++ standard unless there is a requirement to support C++98?

cmastalli: Yes, indeed. Thank you for opening the issue ;)

proyan: This conversation continues in #934

proyan (approved):
This looks good to me, Thanks Carlos for the review.

Regarding the CI failure, it seems to be an `apt` issue, perhaps the new packages aren't available fully. Could you please run again @wxmerkt ?

wxmerkt:
I restarted Travis but robotpkg is still a bit in a weird sync state so it won't pass. As long as gepgitlab passed we should be good?

proyan:
From my side it should be okay to merge. I'll let Carlos do the merging, since he provided the review

proyan:
Thanks @wxmerkt  for this nice work

cmastalli:
@wxmerkt  I have pushed a commit that fixes the code format (which includes c++98 compatibility). Note that the code format is *hardly* enforced in Travis and *softly* enforced in gitlab-ci. It was also needed to satisfies the 16.04 job (which is failing in gitlab-ci).

I will merge it after the current jobs finish (I will ignore Travis to avoid further delays).

cmastalli (approved):

PR #931, opened 2021-04-23
Add stop-watch class (with Python bindings) for profiling code.

andreadelprete:
I've added a stop-watch class which is very useful for doing accurate profiling of the code. This is important for the work that I plan to do to optimize the performance of Crocoddyl with high-order integration schemes. Profiling is off by default so that it does not affect computation times.

PR #935, opened 2021-04-23
Add GitHub Actions CI with ROS dependency resolution

wxmerkt:
This adds a GitHub Actions-based CI on 18.04 and 20.04 resolving Pinocchio/EigenPy via ROS and using catkin tools to build and run tests on Crocoddyl. Part of the motivation is that a) Travis is being phased out, b) Crocoddyl can and is being used within ROS workspaces e.g. for robot deployment, and c) that RobotPkg without staging fails CI jobs every now and then when packages are out of sync (e.g. in #932 and #931 right now). This augments the more extensive coverage and set of configurations Gepgitlab is providing and is more of a stand-in replacement for Travis.

### DIFF
--- a/.github/workflows/industrial_ci_action.yml
+++ b/.github/workflows/industrial_ci_action.yml
@@ -1,7 +1,7 @@
 # This config uses industrial_ci (https://github.com/ros-industrial/industrial_ci.git).
 # For troubleshooting, see readme (https://github.com/ros-industrial/industrial_ci/blob/master/README.rst)
 
-name: CI
+name: ROS-CI
 
 # This determines when this workflow is run
 on: [push, pull_request] # on all pushes and PRs
--- a/.github/workflows/industrial_ci_action.yml
+++ b/.github/workflows/industrial_ci_action.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          submodules: true
+          submodules: recursive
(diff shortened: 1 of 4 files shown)


cmastalli (approved):
This is really nice and useful. Thanks @wxmerkt ;)

nim65s (changes requested):
Thanks for this !

This is currently failing because the submodule for example-robot-data is missing, but I guess this is easy to fix.

jcarpent:
@wxmerkt in your current setup, there is currently no build. Am I missing something?

jcarpent:
Additionally, it would be nice to also do it on Pinocchio, hpp-fcl and eigenpy.

wxmerkt:
Thank you for the feedback and comments. I do plan to add it to eigenpy, pinocchio, hpp-fcl once this one here succeeds. I've added recursive submodule checkouts now so hope the build will pass (https://github.com/wxmerkt/crocoddyl/actions/runs/780608212).

@jcarpent It uses [industrial-ci](https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst) which runs the build in a Docker - as thus I don't need to manually specify the build step manually

nim65s (approved):
LGTM, thanks !

Plus 30min is a nice build time :)

proyan:
With this new CI check, we could allow travis to fail. Could you please do this here @wxmerkt ?

PR #937, opened 2021-04-26
Improving backward pass efficiency

wxmerkt:
This PR follows a conversation from last Thursday's meeting on potentially improving the efficiency of the backward pass. @nmansard mentioned that @jcarpent identified potential improvements. I am happy to test and look into some of these this week - under guidance and advice.

To start off, I checked the backward-pass (`backwardPass`, `tryStep`, `computeGains`) for missing `noalias` but could not find any. Second, I looked through the equations to identify left-hand side terms that could potentially be moved to row-major storage order. These are: `K_`, `FxTVxx_p_`, and `FuTVxx_p_`. A draft to change this is included in this PR - I am expressly looking for feedback whether this is an appropriate step and which other areas we could look into for performance improvement of the serial backward pass.

Initial results _without_ vectorisation on the quadruped benchmark did not show any improvement. 

With vectorisation enabled, using the `all-robots` benchmark, we observe a 13% reduction in time for the backward pass on Talos and a 17% reduction for Anymal. Full outputs are attached and summary table below. Note: Despite the reduction of ~20% in the backward pass benchmark of "all-robots", we only see a ~3.2% reduction in the computation time for a single DDP.solve iteration in "quadrupedal-gaits".

image

All experiments were run with EigenPy, Pinocchio, Crocoddyl compiled with "-O3 -march=native"/ENABLE_VECTORIZATION=ON. CPU set to use the performance governor (i7-10875H @ ~3.8GHz, 16MB cache). Multi-threading disabled.

[ALL_ROBOTS_DEVEL.txt](https://github.com/loco-3d/crocoddyl/files/6378118/ALL_ROBOTS_DEVEL.txt)
[ALL_ROBOTS_ROWMAJOR.txt](https://github.com/loco-3d/crocoddyl/files/6378119/ALL_ROBOTS_ROWMAJOR.txt)
[QUADRUPEDAL_DEVEL.txt](https://github.com/loco-3d/crocoddyl/files/6378121/QUADRUPEDAL_DEVEL.txt)
[QUADRUPEDAL_ROWMAJOR.txt](https://github.com/loco-3d/crocoddyl/files/6378122/QUADRUPEDAL_ROWMAJOR.txt)

Any advice or insights highly appreciated :-).


jcarpent:
Very nice attempt. Seems this is the first milestone. 
I also mentioned last time that vector contained in the structures were not correctly aligned in memory, mostly when using std::vector. This means that some stuff is not aligned for Pinocchio.

I would also suggest activating mavx2 (which for most platform is not active by default).

wxmerkt:
Thank you for the comments, I'll look into aligned vectors next. mavx2 was activated by default through march=native.

src/core/solvers/ddp.cpp
```diff
@@ -336,7 +336,7 @@ void SolverDDP::computeGains(const std::size_t t) {
     }
     K_[t].topRows(nu).noalias() = Qxu_[t].leftCols(nu).transpose();
 
-    Eigen::Block<Eigen::MatrixXd> K = K_[t].topRows(nu);
+    auto K = K_[t].topRows(nu);
```
wxmerkt, line 339: Replacing this `auto` with the supposedly correct type (RowsBlockXpr / `Eigen::Block<MatrixXdRowMajor, Eigen::Dynamic, Eigen::internal::traits<MatrixXdRowMajor>::ColsAtCompileTime, true>`) results in a 3% performance loss across the board. Perhaps the compiler won't make optimisations if the type is explicitly specified? Did I select the wrong return type?

jcarpent: if you have turbo boost active? It may impact the timings for no obvious reason.

jcarpent: by the way, line 337, noalias is not useful.

wxmerkt: Thank you - I just removed the noalias that were unnecessary in f69676c24af5ff3171979a2bd2122a6e937fad8b

wxmerkt:
I've just added a commit that uses a `aligned_vector` container based on Pinocchio's and used it across all files (https://github.com/loco-3d/crocoddyl/pull/937/commits/d394eadcec685a0d18501c663cd4872eaa4abaf2). However, in the benchmarks this disimproved the benchmarks for `all-robots` (cf. attached). I'll revert the commit in the branch for now but would love your insights if I used this incorrectly. Should I only selectively add alignment for K, Quu, etc.?

[ALL_ROBOTS_ROWMAJOR_NOT_ALIGNED.txt](https://github.com/loco-3d/crocoddyl/files/6378656/ALL_ROBOTS_ROWMAJOR.txt)
[ALL_ROBOTS_ROWMAJOR_ALIGNED.txt](https://github.com/loco-3d/crocoddyl/files/6378657/ALL_ROBOTS_ROWMAJOR_ALIGNED_2.txt)

.

src/core/solvers/ddp.cpp
jcarpent: Do you enable turbo boost on your machine.
For stable bench, you should disable it ;)

jcarpent:
@wxmerkt When watching to your timings, they do not seem to be stable. it is hard to say which one is better.

include/crocoddyl/core/solvers/ddp.hpp
```diff
@@ -16,6 +16,8 @@
 
 namespace crocoddyl {
 
+typedef Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> MatrixXdRowMajor;
```
cmastalli, line 19: Could you add this typedef in 'MathTpl'?

src/core/solvers/ddp.cpp
```diff
@@ -456,7 +457,7 @@ const std::vector<Eigen::VectorXd>& SolverDDP::get_Qx() const { return Qx_; }
 
 const std::vector<Eigen::VectorXd>& SolverDDP::get_Qu() const { return Qu_; }
 
-const std::vector<Eigen::MatrixXd>& SolverDDP::get_K() const { return K_; }
+const std::vector<MatrixXdRowMajor>& SolverDDP::get_K() const { return K_; }
```
cmastalli, line 460: Could you verify that this is properly bind by EigenPy? (@jcarpent -- you might comment here)

cmastalli (commented):
@wxmerkt  -- nice that you're handling this important topic.

Could you also handle the BoxQP? @jcarpent -- could you provide us any obvious improvement in that class?

wxmerkt:
> Could you also handle the BoxQP?

I have some separate improvements for BoxQP that I've been using for a while that give a significant boost for cases where constraints are active. I wanted to do this in a separate PR. 

Changing to RowMajor will cause an API breaking change for BoxQP - I am happy with it but it won't be API or ABI compatible (or we can have internal temporaries that are being copied to). Would it be okay to break API/ABI for BoxQP (it's mostly internally)?

wxmerkt:
New numbers are in with Turbo-Boost=OFF, CPU-Governor=Performance - i.e. these are at 2.3 GHz, 16 MB Cache - top shows change between `devel` and moving to RowMajor in this PR, bottom row shows difference between RowMajor and RowMajor and aligned container (
d394ead):

- Using RowMajor as above gives a speed boost of about 16% on a quadruped, 12% on a humanoid
- Moving to aligned-vector containers does not result in a benefit (and would break API right now). This may be because the default alignment to 16 bytes isn't helpful for the odd sizes considered? Did I do something wrong? 

image

I'll go ahead and address the comments above - otherwise, are there additional points to consider or debug for the backward pass?

src/core/solvers/ddp.cpp
wxmerkt: Should work since Feb 2020, cf. https://github.com/stack-of-tasks/eigenpy/blob/b55c10dde71f81aaa51e199f4b64071d31e882eb/src/matrix-double.cpp#L12

cmastalli:
> I have some separate improvements for BoxQP that I've been using for a while that give a significant boost for cases where constraints are active. I wanted to do this in a separate PR.

Sounds a good idea. Looking forward for reviewing that PR :)

> Changing to RowMajor will cause an API breaking change for BoxQP - I am happy with it but it won't be API or ABI compatible (or we can have internal temporaries that are being copied to). Would it be okay to break API/ABI for BoxQP (it's mostly internally)?

Yes, we can break the API if this leads to a computational benefit. What it is important for our users is to adding a deprecate procedure. We could discuss it when you create that PR.

src/core/solvers/ddp.cpp
cmastalli: Thanks for the answer :)

cmastalli:
> I'll go ahead and address the comments above - otherwise, are there additional points to consider or debug for the backward pass?

Perhaps, we could improve the computational time by using `triangularView` of Hessian terms. @jcarpent -- is this could lead to a computation-time reduction? If so, could you provide a guidance in this topic?

jcarpent:
This is the line in question: https://github.com/loco-3d/crocoddyl/blob/cbf81a329e3abaf4ce1b4a8fab1431f93cd9a5c8/src/core/solvers/ddp.cpp#L248
There is memory allocation. For quick fix, a tmp variable should be allocated once for all.

wxmerkt:
Thank you for noticing the heap allocation. I fixed this now. I spent some time on triangularView etc. but the improvements and changes have been hard to quantify. I believe we are now at the level where more time spent on optimisation is likely to outweigh the benefits and would consider this PR as complete.

src/core/solvers/ddp.cpp
```diff
@@ -258,7 +258,8 @@ void SolverDDP::backwardPass() {
       }
       Vxx_[t].noalias() -= Qxu_[t].leftCols(nu) * K_[t].topRows(nu);
     }
-    Vxx_[t] = 0.5 * (Vxx_[t] + Vxx_[t].transpose()).eval();
+    Vxx_tmp_ = Vxx_[t] + Vxx_[t].transpose();
+    Vxx_[t].noalias() = 0.5 * Vxx_tmp_;
```
jcarpent, line 262: Putting 0.5 in the above line will conduct to even better performances I imagine.
Another way to improve this would be to do the operation in two passes. But I do agree, improving timings takes times.

cmastalli: See also my comment in https://github.com/loco-3d/crocoddyl/pull/937#commitcomment-50062484

wxmerkt: Thank you, amended :-)

cmastalli (approved):
Everything looks fine for me
Thank you for this nice work :)

proyan:
Thanks @wxmerkt and @jcarpent for your efforts in this PR 👍🏽

PR #938, opened 2021-04-26
Activate all warnings; turn off Wsign-conversion

wxmerkt:
We have a lot of conversions between `Eigen::Index` and `std::size_t` (e.g. `.head(nu)` etc.) which pollute the compiler output. Fixing those would make part of the code quite unreadable. Instead, this PR activates all warnings and then only turns off `Wsign-conversion`. The remaining warning is fixed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,8 +34,7 @@ ENDIF()
 SET(DOXYGEN_USE_MATHJAX YES)
 SET(DOXYGEN_USE_TEMPLATE_CSS YES)
 
-# Disable -Werror on Unix for now.
-SET(CXX_DISABLE_WERROR True)
+SET(CXX_DISABLE_WERROR False)
 SET(PROJECT_USE_CMAKE_EXPORT True)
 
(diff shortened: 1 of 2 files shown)


cmastalli:
Are you referring to clang output? (GCC doesn't show these messages)

wxmerkt:
With the `-Wall` and the other settings in this PR it's likely that gcc shows them as well. After this PR I get 0 warnings from clang (I fixed the unused variable ones in the earlier PR) which is nice and means all new warnings would be introduced by new code/code changes and can be fixed immediately - to enforce this we may want to turn on "Werror" in a build.

cmastalli:
> which is nice and means all new warnings would be introduced by new code/code changes and can be fixed immediately - to enforce this we may want to turn on "Werror" in a build.

Yes, I like the idea to turn on `Werror`. It will help us to avoid cleaning-up future PRs. Could you include it in this PR?

cmastalli (approved):
Thanks @wxmerkt :)

PR #941, opened 2021-04-27
Fix quadrotor examples; update log files (that changed)

wxmerkt:
This PR is mostly to investigate log files that have changed significantly. It also includes a fix for the nominal configuration of models that do not have an SRDF (which now returns `None` rather than `pinocchio.neutral(model)`. The logs that aren't included only had some numerical tolerance differences. Computed using -O3 and vectorisation


examples/log/bipedal_walk.log
```diff
@@ -1,98 +1,119 @@
...
+   1  8.16841e+08  4.30532e+12  1.62891e+09  1.00000e+01  1.00000e+01   0.0020     0
+   2  6.46220e+07  4.40688e+12  1.62520e+09  1.00000e+00  1.00000e+00   1.0000     1
+   3  3.21119e+07  7.21526e+14  1.28839e+08  1.00000e+00  1.00000e+00   0.5000     1
+   4  3.08284e+07  6.45283e+14  6.40505e+07  1.00000e+00  1.00000e+00   0.0625     1
+   5  2.87518e+07  7.19990e+14  6.14803e+07  1.00000e+00  1.00000e+00   0.1250     1
+   6  1.52045e+07  8.06352e+14  5.73367e+07  1.00000e-01  1.00000e-01   1.0000     1
+   7  1.44841e+07  5.61935e+14  3.03400e+07  1.00000e-01  1.00000e-01   0.2500     1
+   8  1.11486e+07  6.71913e+14  2.89160e+07  1.00000e-01  1.00000e-01   0.2500     1
+   9  6.03658e+06  5.76574e+14  *** SOLVE walking ***
+*** SOLVE walking ***
```
wxmerkt, line 100: This is a new change in output order / not flushing properly that I haven't seen before

cmastalli:
I think enabling vectorization changes the log files. It would be useful to generate them without vectorization as well.

wxmerkt:
The most notable changes are in the bipedal examples (up to an order of magnitude difference in initial cost). I also had to fix flushing of print output to avoid out of order printing, which seems odd. I'll rebase on devel and try with and without vectorisation next.

proyan:
>  I also had to fix flushing of print output to avoid out of order printing, which seems odd

Are you compiling with multi-threading support?

wxmerkt:
> Are you compiling with multi-threading support?
No, this is all single core but vectorisation is enabled. Unlike usually when I'm using stock 20.04, this (and all dependencies) is in a fresh conda environment - however I dont think that this matters.

cmastalli:
@wxmerkt and @proyan -- I think the latest values of the log files belong to the case when we use `1e-15` as a minimum regularization value. You can observe that `1e-9` produces more iterations in many examples (e.g., bipedal walk).

Before merging this PR, I would like to compute again this log files.

wxmerkt:
Do you suggest to change the regularisation value here in this branch or rebase on master and recompute based on the commits in master?

proyan:
@wxmerkt We decided to use 1e-9 as the cutoff regularization. A very low regularization means that the solver spends multiple iterations in order to arrive at a value that is substantial enough to affect the problem. I don't think that we should change the default regularization value

wxmerkt:
To double check my understanding: I should change the cut-off regularisation in this PR to 1e-9 and then recreate the logs. Note, these logs were created with devel as of last week.

wxmerkt:
I rebased on devel, reran the logs and force pushed this branch. 

I also locally changed `reg_min_` back to 1e-15 and these are the changes (not committed):

- **bipedal_walk_ubound**: one of the phases converges in 18 rather than 29 iterations, another phase is 1 iteration faster, the final phase is 31 instead of 28 iterations currently.
- **humanoid_taichi**: takes 122 instead of 116 iterations currently
- **quadrotor_ubound**: takes 1 more iteration

The other logs only have minor numerical changes

examples/quadrotor.py
```diff
@@ -52,7 +52,8 @@
 
 # Creating the shooting problem and the FDDP solver
 T = 33
-problem = crocoddyl.ShootingProblem(np.concatenate([hector.q0, np.zeros(state.nv)]), [runningModel] * T, terminalModel)
+q0 = pinocchio.neutral(hector.model)
```
cmastalli, line 55: we need to revert back the changes in the quadrotor example. The problem was at the example-robot-data repo. See this merged PR: https://github.com/Gepetto/example-robot-data/pull/80

examples/quadrotor_ubound.py
```diff
@@ -54,7 +54,8 @@
 
 # Creating the shooting problem and the boxddp solver
 T = 33
-problem = crocoddyl.ShootingProblem(np.concatenate([hector.q0, np.zeros(state.nv)]), [runningModel] * T, terminalModel)
+q0 = pinocchio.neutral(hector.model)
+problem = crocoddyl.ShootingProblem(np.concatenate([q0, np.zeros(state.nv)]), [runningModel] * T, terminalModel)
```
cmastalli, line 58: same here

wxmerkt:
I've reverted the changes to the quadrotor examples - this will require a new release of example-robot-data for those to work.

wxmerkt:
Is anything else missing to get this merged?

examples/bipedal_walk.py
```diff
@@ -48,6 +48,7 @@
 
     # Added the callback functions
     print('*** SOLVE ' + key + ' ***')
+    sys.stdout.flush()
```
cmastalli, line 51: We think we should remove all the `sys.stdout.flush()` in the example scripts. The reason is we just need them in the callback function.

cmastalli: I won't mark all the cases as they are easy to spot.

examples/log/arm_manipulation.log
```diff
@@ -1,8 +1,16 @@
 iter 	 cost 	      stop 	    grad 	  xreg 	      ureg 	 step 	 feas
-   0  4.28053e-01  1.91916e-08  1.88433e-01  1.00000e-10  1.00000e-10   1.0000     1
-   1  3.76491e-01  4.30981e-08  3.02072e-01  1.00000e-10  1.00000e-10   0.5000     1
-   2  2.75460e-01  3.03653e-08  2.73822e-01  1.00000e-10  1.00000e-10   0.5000     1
-   3  2.46128e-01  1.12361e-08  1.00743e-01  1.00000e-11  1.00000e-11   1.0000     1
-   4  1.97554e-01  1.18552e-08  1.02998e-01  1.00000e-12  1.00000e-12   1.0000     1
-   5  1.92186e-01  1.02679e-09  9.51238e-03  1.00000e-13  1.00000e-13   1.0000     1
-   6  1.89358e-01  2.41019e-10  2.31141e-03  1.00000e-14  1.00000e-14   1.0000     1
+   0  5.15359e-01  3.65308e-08  1.81450e-01  1.00000e-09  1.00000e-09   1.0000     1
```
cmastalli, line 2: I want to check myself all these logs, I will do it asap.

wxmerkt:
I reverted the `sys.stdout.flush` as requested but it breaks things again when using conda-python-3.8:
image

cmastalli:
> I reverted the `sys.stdout.flush` as requested but it breaks things again when using conda-python-3.8:

We should have at least another `std::flush` after calling "print variables name" (i.e., in here https://github.com/wxmerkt/crocoddyl/blob/topic/update-log-files/src/core/utils/callbacks.cpp#L36).

However, to be honest, I am not sure if we need to add `std::flush` for each line (i.e., starting before the new `std::cout`).

I believe this should solve this issue in conda-python-3.8. Could you try it?

wxmerkt:
> I believe this should solve the issue in conda-python-3.8. Could you try it?

I've tried it but alas it did not fix it :-(

cmastalli:
> I've tried it but alas it did not fix it :-(

It seems to be related with that setup. You should be able to solve it by running the script as `python -u my_script.py`. For more information see: https://stackoverflow.com/questions/230751/how-can-i-flush-the-output-of-the-print-function

wxmerkt:
> It seems to be related with that setup. You should be able to solve it by running the script as python -u my_script.py. For more information see: https://stackoverflow.com/questions/230751/how-can-i-flush-the-output-of-the-print-function

Thanks, that solved it! The PR should now be ready for review & merge

cmastalli:
I will merge this PR gitlab-ci is enough

PR #942, opened 2021-04-28
Release v1.7.0

cmastalli:
Hi @nim65s,

This new release incorporates the following points:
 1. Updated the examples based on new API in example-robot-data
 2. Removed reference in `std::sized_t` (and other primitives)
 3. Improved computation and handled richer conditions in friction and wrench cone (e.g., inner/outer apprx. in wrench cone, and rotation matrix in friction cone)
 4. Added more unit-tests for cones
 5. Included the CoP support notion
 6. Updated minimal version of EigenPy as it fixes a bug with 4x6 matrices
 7. Developed a gravity-based cost function for both free and in contact conditions (included its unit-test code)
 8. Added assignment operator in FrameXXX structures
 9. Replaced `isMuchSmallerThan` by `isZero` in all the unit-tests
 10. Enabled free-flyer joint in full actuation model
 11. Improved multicopter actuation API + unit-tests
 12. Exposed in Python ActionModelUnicyle::dt_
 13. Fixed multithreading support: running the correct number of threads
 14. Enabled that the user can set the number of threads (also in Python)
 15. Added publication list and updated README file
 16. Registered in Python the shared pointers of all the model classes
 17. Fixed Meshcat visualizer after update
 18. Added Github Action CI with ROS dependency resolution
 19. Improved efficiency of backward pass by defining properly RowMajor matrices
 20. Activated all warnings and Werror
 21. Improved documentation

PR #943, opened 2021-04-28
Hotfix for exposing std::vector<MatrixXdRowMajor>

wxmerkt:
Required before #942

### DIFF
--- a/bindings/python/crocoddyl/crocoddyl.cpp
+++ b/bindings/python/crocoddyl/crocoddyl.cpp
@@ -1,7 +1,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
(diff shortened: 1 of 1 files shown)


proyan:
It would be good if we try to keep close to pinocchio nomenclature. Could you please rename the RowMajor Matrix to `RowMatrixXs`?

https://github.com/stack-of-tasks/pinocchio/blob/master/src/multibody/data.hpp#L77

wxmerkt:
Thank you for the pointer @proyan. I've fixed it for the binding here and will amend in the next PR I'll open with more RowMajor changes.

proyan (approved):
Thanks for the PR @wxmerkt

cmastalli (approved):
Thanks @wxmerkt and @proyan.
Let's merge this PR before the residual one :)

PR #946, opened 2021-05-06
Fixed GV display after changes in #911

cmastalli:
Note that this error was not triggered by running the examples with `display` option.

### DIFF
--- a/bindings/python/crocoddyl/__init__.py
+++ b/bindings/python/crocoddyl/__init__.py
@@ -140,8 +140,8 @@ class GepettoDisplay(DisplayAbstract):
                         forcePose = pinocchio.SE3ToXYZQUATtuple(pinocchio.SE3(R, pose.translation))
                         forceMagnitud = np.linalg.norm(wrench.linear) / self.totalWeight
                         forceName = self.forceGroup + "/" + key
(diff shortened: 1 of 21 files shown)


cmastalli:
I have pushed a commit that includes another feature!
Now, the Python users can also define the type of dependency in the residual function.

jcarpent (commented):
I would say that these properties should not change after construction.

cmastalli:
> I would say that these properties should not change after construction.

Yes, I agree with you. I will change this bit asap.

proyan:
@cmastalli This PR should be okay to merge once you could take care of Justin's concerns

cmastalli:
This is the list of changes:
 1. I have included the definition of the dependencies in the constructor API (the only way to avoid that users modify it).
 2. I have also used this API for constructing all the residuals (important for resolving nondeterminism when nu is equals to 0)
 3. I have defined the q,v,u_dependent flags for the control residual (this bit was missed in previous PR).
 4. I have fixed a tiny issue in the computation of u_dependent terms in the CostModelResidual class.

cmastalli:
@proyan and @jcarpent -- let me know if you are happy with the provided solution. Otherwise, I will merge this PR tomorrow afternoon.

proyan (approved):
Thanks Carlos for this fix

PR #949, opened 2021-05-12
[C++14] Move default C++ standard support to 14

proyan:
Since the residual PR https://github.com/loco-3d/crocoddyl/pull/899, we now use the template aliases https://en.cppreference.com/w/cpp/language/type_alias, which is a C++11 feature. Thus, the devel has now unofficially moved to the C++11 standard. 

Following internal discussions, it was decided to switch the support to C++14 directly, since that is the default C++ standard in the currently supported Ubuntu LTS versions (18.04<), and is supported in the last unsupported Ubuntu LTS version 16.04.

This PR should officially make this switch to C++14 standard for crocoddyl


CMakeLists.txt
```diff
@@ -55,6 +55,8 @@ PROJECT(${PROJECT_NAME} ${PROJECT_ARGS})
 # If needed, fix CMake policy for APPLE systems
 APPLY_DEFAULT_APPLE_CONFIGURATION()
 
+CHECK_MINIMAL_CXX_STANDARD(14 ENFORCE)
```
wxmerkt, line 58: If I set my environment to 17, will this macro cause a conflict or will it gracefully accept it since 17>14?

wxmerkt (approved):

CMakeLists.txt
proyan: Based on @nim65s's logic here: https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/cxx-standard.cmake#L64, I would say that 14< should be allowed.

cmastalli (approved):

cmastalli:
I will also merge this PR

PR #952, opened 2021-05-21
Only build Python-examples if bindings are built

wxmerkt:
Currently we have a configuration option that can break build: `cmake .. -DBUILD_PYTHON_INTERFACE=OFF` will still try to build the Python examples. This PR fixes this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,7 +183,11 @@ ENDIF(BUILD_BENCHMARK)
 
 # Build the examples
 IF(BUILD_EXAMPLES)
-  ADD_SUBDIRECTORY(examples)
+  IF(BUILD_PYTHON_INTERFACE)
+    ADD_SUBDIRECTORY(examples)
+  ELSE(BUILD_PYTHON_INTERFACE)
+    MESSAGE(WARNING "Python interface is not built, hence cannot build examples.")
+  ENDIF(BUILD_PYTHON_INTERFACE)
 ENDIF(BUILD_EXAMPLES)
 
 # PkgConfig packaging of the project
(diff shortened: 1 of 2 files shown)


cmastalli:
Well catched @wxmerkt!

I think it is better if we do not include the Cmake file inside the examples folder. We could also add the rule that example-robot-data dependency is not needed if we don't want to generate the Python bindings.

nim65s:
that's right. example-robot-data is required only if `DBUILD_PYTHON_INTERFACE && BUILD_TESTING`.

proyan:
@wxmerkt ping :)
Could you please implement @cmastalli's suggestion here? Other than removing the examples, I think this PR is good to go

wxmerkt:
> I think it is better if we do not include the Cmake file inside the examples folder.

Every other folder (benchmarks, bindings) has their own CMakeLists that gets included via `add_subdirectory`. I take it the suggestion is to move the example build logic to the main CMakeLists? Or to conditionally include based on `BUILD_PYTHON_INTERFACE`? The latter only makes sense if we are not adding C++ examples at some point in the future. Happy to do the former but it breaks with the logic in the other directories (benchmarks, bindings). Which option should I pursue?

cmastalli:
> Or to conditionally include based on `BUILD_PYTHON_INTERFACE`? The latter only makes sense if we are not adding C++ examples at some point in the future.

Yes, my suggestion was this later point. Personally, I am not particularly interested in adding c++ examples, and do not think someone else will do it.

wxmerkt:
Thanks for the clarification, now amended

cmastalli (approved):
Thanks @wxmerkt :)

PR #954, opened 2021-05-28
fix include sequence after pinocchio commit c1ff8821c437e2401b80b7aa7f3f118464e2c051

proyan:
Pinocchio commit c1ff8821c437e2401b80b7aa7f3f118464e2c051 forces pinocchio inclusion before boost.
So crocoddyl now fails to compile with pinocchio devel.
This PR is meant to change the header include order based on the following logic:

1) Non-boost system headers
2) pinocchio headers
3) crocoddyl headers
4) pinocchio binding headers
5) crocoddyl binding headers

### DIFF
--- a/bindings/python/crocoddyl/multibody/frames.cpp
+++ b/bindings/python/crocoddyl/multibody/frames.cpp
@@ -11,12 +11,10 @@
 #include <eigenpy/memory.hpp>
 #include <eigenpy/eigen-to-python.hpp>
 
-
 #include <pinocchio/bindings/python/utils/std-aligned-vector.hpp>
 #include "python/crocoddyl/multibody/multibody.hpp"
 #include "python/crocoddyl/utils/printable.hpp"
 #include "python/crocoddyl/utils/deprecate.hpp"
-
 
 EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(crocoddyl::FrameTranslation)
 EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(crocoddyl::FrameRotation)
--- a/bindings/python/crocoddyl/multibody/frames.cpp
+++ b/bindings/python/crocoddyl/multibody/frames.cpp
@@ -6,15 +6,17 @@
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
+#include "crocoddyl/multibody/frames.hpp"
+
 #include <eigenpy/memory.hpp>
 #include <eigenpy/eigen-to-python.hpp>
 
+
+#include <pinocchio/bindings/python/utils/std-aligned-vector.hpp>
 #include "python/crocoddyl/multibody/multibody.hpp"
-#include "crocoddyl/multibody/frames.hpp"
 #include "python/crocoddyl/utils/printable.hpp"
 #include "python/crocoddyl/utils/deprecate.hpp"
 
-#include "pinocchio/bindings/python/utils/std-aligned-vector.hpp"
 
 EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(crocoddyl::FrameTranslation)
 EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(crocoddyl::FrameRotation)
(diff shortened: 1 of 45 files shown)


bindings/python/crocoddyl/multibody/frames.cpp
```diff
@@ -6,15 +6,17 @@
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
+#include "crocoddyl/multibody/frames.hpp"
+
 #include <eigenpy/memory.hpp>
 #include <eigenpy/eigen-to-python.hpp>
 
+
+#include "pinocchio/bindings/python/utils/std-aligned-vector.hpp"
```
cmastalli, line 15: Could you use `<>` simbol (as it is an external header)?

include/crocoddyl/core/activation-base.hpp
```diff
@@ -1,7 +1,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2019-2021, LAAS-CNRS, University of Edinburgh
+// Copyright (C) 2019-2020, LAAS-CNRS, University of Edinburgh
```
cmastalli, line 4: Please revert the change on the copyright year.

include/crocoddyl/core/activation-base.hpp
```diff
@@ -12,12 +12,13 @@
 #include <stdexcept>
 #include <boost/shared_ptr.hpp>
 #include <boost/make_shared.hpp>
-#include <boost/core/demangle.hpp>
 
 #include "crocoddyl/core/fwd.hpp"
 #include "crocoddyl/core/mathbase.hpp"
 #include "crocoddyl/core/utils/to-string.hpp"
 
+#include <boost/core/demangle.hpp>
```
cmastalli, line 20: Why do we need to include this header after Crocoddyl stuff?

The common rule is to include external headers first.

include/crocoddyl/core/integrator/euler.hxx
```diff
@@ -6,12 +6,12 @@
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
+#include "crocoddyl/core/utils/exception.hpp"
```
cmastalli, line 9: Same here.

cmastalli (changes requested):
Thanks Rohan for taking care of this tiny, yet important, details.

I have left a few comments which I believe they need to be addressed.

bindings/python/crocoddyl/multibody/frames.cpp
proyan: sure

bindings/python/crocoddyl/multibody/frames.cpp
```diff
@@ -6,15 +6,17 @@
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
 
+#include "crocoddyl/multibody/frames.hpp"
```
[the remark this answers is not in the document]
proyan: eigenpy uses boost, so no. Similarly, pinocchio bindings use boost

include/crocoddyl/core/activation-base.hpp
proyan: Sure

include/crocoddyl/core/integrator/euler.hxx
proyan: Sure

proyan:
Thanks for the review Carlos. This isn't a tiny detail, unfortunately. Crocoddyl devel doesn't compile with pinocchio devel because of the aforementioned commit. This is meant to fix that.

cmastalli:
> Thanks for the review Carlos. This isn't a tiny detail, unfortunately. Crocoddyl devel doesn't compile with pinocchio devel because of the aforementioned commit. This is meant to fix that.

Thanks for the explanation, I wasn't aware.

cmastalli (approved):
All alright from my side. However, the code might need to be formatted.

jcarpent:
> Thanks for the review Carlos. This isn't a tiny detail, unfortunately. Crocoddyl devel doesn't compile with pinocchio devel because of the aforementioned commit. This is meant to fix that.

I've removed some macro def which were not aimed to stay within Pinocchio. To avoid client breacks, I may restore one commit.

proyan:
> I've removed some macro def which were not aimed to stay within Pinocchio. To avoid client breacks, I may restore one commit.

Ok, I think that would be better for other downstream packages. For now Ill merge this

PR #958, opened 2021-06-04
[cmake] use OPTION to set BUILD_TESTING

### LINKED ISSUE #955: BUILD_TESTING OFF still builds unittests
We are not using BUILD_TESTING in the correct way in crocoddyl, and as a result unittests are built even when it is turned off. 

The issue is assigned to @nim65s , but if someone has the time to fix it first, please change the assignment and go ahead.

proyan:
Closes #955


cmastalli (approved):
Thanks @proyan and @nim65s for taking care of this issue ;)

PR #961, opened 2021-06-08
Align the arguments

jviereck:
Align the arguments based on the documentation `def __init__(self, nq, nv, nu):` in `class DifferentialActionModelAbstract:`[1]

[1] https://github.com/loco-3d/crocoddyl/blob/0985b92098ba70051ec3e5440526a7ec96d7aeca/unittest/python/crocoddyl/differential_action.py#L8


jviereck:
Looks like GitHub is referencing the old legacy code, which has different constructor arguments. The example looks fine when seeing the newer code here:

https://github.com/loco-3d/crocoddyl/blob/cbf81a329e3abaf4ce1b4a8fab1431f93cd9a5c8/bindings/python/crocoddyl/core/diff-action-base.hpp

PR #948, opened 2021-05-12
[travis][CI] use travis untill #944 is resolved

proyan:

PR #963, opened 2021-06-11
Run unittests in ROS-CI, deactivate Travis-CI

### LINKED ISSUE #944: ROS-CI does not yet run unittests
We use `catkin_tools` as the `BUILDER` to avoid dependency resolution issues in the `upstream_ws` which contains `example-robot-data` (`colcon` would not resolve this correctly in CI). Alas, `catkin_tools` requires the `run_tests` target which we currently do not have. `colcon` would use `ctest` correctly on the other hand and works when manually sourcing the `upstream_ws` before building the `target_ws`. As such, right now there is no quick fix to run the unittests on the ROS-CI on GitHub Actions. This issue is to track this and be resolved at some point in the future. If anyone has a fix or suggestion, happy to go with it :-).

wxmerkt:
This PR addresses two items:

1. Execution of unit tests on the ROS-CI running on GitHub Actions. In turn, allows failures on Travis which is prone to spurious failures (thus reverts #948)
2. Updates the cmake submodule to include a packaging fix (https://github.com/jrl-umi3218/jrl-cmakemodules/pull/479) important for use of Crocoddyl in downstream projects 

Resolves #944

### DIFF
--- a/.github/workflows/industrial_ci_action.yml
+++ b/.github/workflows/industrial_ci_action.yml
@@ -14,9 +14,18 @@ jobs:
           - {ROS_DISTRO: melodic}
           - {ROS_DISTRO: noetic}
     env:
-      CCACHE_DIR: /github/home/.ccache # Enable ccache
-      UPSTREAM_WORKSPACE: dependencies.rosinstall
+      CCACHE_DIR: /github/home/.ccache             # Enable ccache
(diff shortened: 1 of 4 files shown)


wxmerkt:
Tests now run successfully - cf. https://github.com/loco-3d/crocoddyl/pull/963/checks?check_run_id=2803711733#step:4:1019 for example test output

I just added another commit to fix a previous failure in Python formatting to also make Gitlab-CI pass.

cmastalli (approved):
Nice! thanks @wxmerkt :)

PR #965, opened 2021-06-21
Devel static

edantec:
This PR aims at providing an unittest for pair collision residual. Direct comparison of pair collision residual derivatives with numerical differentiation will not work by design, because we are only interested in the derivatives of the distance between two objects, and not the derivatives of the witness segment between these two objects. When combined with 2norm-barrier activation, pair collision residual produces the norm of the distance and this can be compared with numerical differentiation. 

This is what is implemented in this PR.


cmastalli:
Hi @edantec -- it seems you should rebase your branch (or merge the devel one into your branch) to make the review process easy for us. I see changes that do not belong to this effort.

cmastalli:
Could you rebase or merge before I start to review this PR?

proyan:
@edantec Before you create a PR, you should always look at the changes across the branches that you are proposing. Github gives you the functionality to verify the commits, and the changes across the files.

edantec:
I merged loco3d/devel into edantec/devel_static, hope it solves the problem.
Sorry for that!

edantec:
It didn't seem to have solve it... Maybe I should close this PR and open it again?

proyan:
> It didn't seem to be solved... Maybe I should close this PR and open it again?

I think it would be best. Rebase on top of the current devel, and then you should only see the changes related to the unittest. If not, there is some other issue. I'll close this PR :)

Issue #925, opened 2021-04-08
Error in computation of Value function's Hessian?


andreadelprete:
Looking at the code in [solver DDP](https://github.com/loco-3d/crocoddyl/blob/master/src/core/solvers/ddp.cpp#L246) it seems that the value function's Hessian is computed as:
```
Vxx = Qxx - Qxu*K 
```
As far as I know, this formula is correct when the regularization (on the control and the state) is null. However, when it's not null, this formula should be used:
```
Vxx = Qxx - Qxu*K - K^T*Qux + K^T*Quu*K
```
We can easily check that the two formulas are equivalent in the case of zero regularization by substituting `K = Quu^-1 * Qux`. This was explained in [1], in the paragraph above equation (11).

[1] Tassa, Y., Erez, T., & Todorov, E. (2012). Synthesis and stabilization of complex behaviors through online trajectory optimization. In Intelligent Robots and Systems (IROS), IEEE/RSJ International Conference on (pp. 4906–4913)

wxmerkt:
Related previous merge request on gepgitlab following an email conversation based on the same observation: https://gepgitlab.laas.fr/loco-3d/crocoddyl/-/merge_requests/241

cmastalli:
In Crocoddyl, we are regularizing all the Hessian terms (not only the diagonals one, i.e., `Qxx` and `Quu`). If we do so, the longer expression can be condensed with the shorter one. Additionally, Tassa equation for updating the Jacobian and Hessian of the Value function seems to be wrong, as it uses the nonregularization terms for `Qxx`, `Quu` and `Qux`. It is like comparing  apples with pears! Thus, from that perspective, it seems to be everything alright.

On the other hand, our regularization strategy is not strictly the one employed in nonlinear programs (e.g., KNITRO's interior-point algorithms). We are doing two dirty things that I understood a few months ago, but I did not have time to look on it.

First, we combine `Vxx` and `Quu` regularizations. Again, comparing apples with pears as described below
   -  `Vxx` regularization considers the curvature, e.g.,
      - `Qxu_r = lxu + fx^T*(Vxx + mu*I)*fu = Qxu  + mu*fx^T*fu`
   - `Quu` regularization doesn't consider the curvature (classical damped equation), e.g.,
     -  `Quu_r = Quu + mu*I`

Thus, the current regularization in Crocoddyl is
[`Qxx + mu*fx^T*fx`;     `Qxu + mu*fx^T*fu`]
[...]

PR #974, opened 2021-07-12
Used the shorter expression to compute Vx #925

cmastalli:
This PR solves the issue discussed in #925, i.e., it uses the shorter expression for computing `Vx`.
This reduces unnecessary computations but it also shows a bit better numerical results as it does avoid to propagate (more than unavoidable) numerical noise.

@andreadelprete -- I still keep the `Quu` regularisation as the results across many examples are better with it.
Despite that this is not a definitive answer, I would prefer to do not change something that it is working well.


cmastalli:
I'll merge this PR, but if somebody has a concern, I am happy to work on them immediately.

PR #975, opened 2021-07-13
Improved stopping criteria

cmastalli:
This PR proposes to change the stopping criteria from `||Qu||` to `|dV[0]|`. Below there is a justification of this choice

`dV[0]` represents the gradient of the cost function for the entire trajectory. Instead `||Qu||` assumes that the algorithm is a pure
single-shooting approach, i.e., it depends in the control only. However, this is not truth as DDP is a multiple-shooting approach with a nonlinear search that always guarantee dynamic feasibility. This argument can also be extended for other cases.

Additonally, in the log files (i.e., double pendulum, quadrotor, and arm manipulation), we can also observe that `||Qu||` criteria is stopping the algorithm much earlier than desired. It doesn't consider the further improvement needed for changes in `xs`.

I have similar conclusions for a prototype algorithm that handles equality constraints. Unfortunately, I cannot disclosure yet this work.


src/core/solvers/ddp.cpp
```diff
@@ -139,16 +139,7 @@ double SolverDDP::tryStep(const double steplength) {
...
-  const std::size_t T = this->problem_->get_T();
-  const std::vector<boost::shared_ptr<ActionModelAbstract> >& models = problem_->get_runningModels();
-
-  for (std::size_t t = 0; t < T; ++t) {
-    const std::size_t nu = models[t]->get_nu();
-    if (nu != 0) {
-      stop_ += Qu_[t].head(nu).squaredNorm();
-    }
-  }
+  stop_ = abs(d_[0]);
```
proyan, line 142: why do you need an abs here?

proyan:
I don't think I totally agree with your reasoning, but ||QuT Qu|| and ||QuQuu-1Qu|| would provide similar results for PosDef Quu, as we can see in the logs.
However, before merging or rejecting this PR, perhaps it is better if we discuss on Thursday about it.

cmastalli:
Yes, we should discuss this in private. Note that this ||Qu|| doesn't lead to similar results when you consider other algorithms as this metric is not correct, instead dV[0] is a general metric.

This claim is supported by results obtained in an algorithm (which I developed) that projects k, K along the nullspace of the equality constraint. Indeed, ||Qu|| fails to evaluate the algorithm convergence.

|dV[0]| also resembles the norm of the KKT matrix for a multiple-shooting formulation. And this holds for formulations with arbitrary, yet Markovian, constraints as well.

src/core/solvers/ddp.cpp
proyan: We don't need the abs here, d_[0] would remain positive by definition.

proyan: Could you please remove it?

cmastalli: This might be indeed a possible choice, despite that the `abs` operation doesn't hurt, but I am not fully sure now.
Please give me some time to come with a definite conclusion.

Side comment:
The `abs` operation is important in equality-constrained cases whether d[0] can be negative. I found often in these cases that very tiny negative values (appr. -1e-12) at the local optimum.

cmastalli: Another comment (sorry for multiple responses):
In equality-constrained cases, d[0] might compute an ascent direction (i.e., a substantially big negative value), which produces an undesired (silly) early stopping procedure as we check that `d[0] < th_stop`.

proyan: In that case, this is something that belongs to the specific solver that you are working with. 
I see no reason to change the stopping criteria in base ddp class.

cmastalli: > In that case, this is something that belongs to the specific solver that you are working with.
> I see no reason to change the stopping criteria in base ddp class.

Yes, I agree with that. To be honest, my main concern is with the FDDP, where we know we have (and can accept) ascent directions. So, if this is case we might need to re-implement the `stoppingCriteria` function.

proyan: For FDDP, we might handle separately this issue then, once we understand it better. 
I believe for your solver, implementing in the solver itself would be the best way forward. 
I'll close this PR following the discussion

proyan:
discussion carried out here: 
https://github.com/loco-3d/crocoddyl/pull/975#discussion_r669148973

src/core/solvers/ddp.cpp
cmastalli: Hey @proyan -- I understand well the need of using dV[0] as stopping criteria, please do not think the opposite. But you asked me to remove `abs` operation, and for this last bit, I need to think first.

The reason why I didn't check it first is due to the fact that the `abs` operation won't hurt (in any form) the measurement of the stopping criteria. And to be honest with you, I think we could go ahead as this doesn't really matters. What matters is to discuss `dV[0]` vs `||Qu||`, despite that I am convinced on my arguments and empirical evidence.

Therefore, I'll reopen this issue based on above explanations :)

cmastalli:
I will close this PR.
As discussed internally, we would introduce this stopping criteria in Crocoddyl v2.

PR #978, opened 2021-07-22
Support for pinocchio 2.9.0 constraint dynamics formulation

proyan:
This PR introduces constraint dynamics crocoddyl wrapper for the constraint dynamics algorithm introduced in pinocchio 2.9.0.

This PR does not yet deprecate the existing contact model framework used within crocoddyl. The existing ContactModel framework would have to be deprecated with full release of pinocchio 3.

The PR does not yet add the unittest for constraint-dynamics.


include/crocoddyl/multibody/actions/constraint-fwddyn.hxx
```diff
@@ -0,0 +1,196 @@
...
+  DifferentialActionDataConstraintFwdDynamicsTpl<Scalar>* d =
+      static_cast<DifferentialActionDataConstraintFwdDynamicsTpl<Scalar>*>(data.get());
+  const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> q = x.head(state_->get_nq());
+  const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> v = x.tail(state_->get_nv());
+
+  // Computing the forward dynamics with the holonomic constraints defined by the contact model
+
+  actuation_->calc(d->multibody.actuation, x, u);
+  // TODO: Add mu
+  d->xout = pinocchio::constraintDynamics(pinocchio_, d->pinocchio, q, v, d->multibody.actuation->tau, contacts_,
```
cmastalli, line 58: Should we call `pinocchio::computeCentroidalMomentum` for cost function that depends on centroidal quantities?

include/crocoddyl/multibody/actions/constraint-fwddyn.hpp
```diff
@@ -0,0 +1,150 @@
...
+  typedef DifferentialActionDataAbstractTpl<Scalar> DifferentialActionDataAbstract;
+  typedef DifferentialActionDataConstraintFwdDynamicsTpl<Scalar> DifferentialActionDataConstraintFwdDynamics;
+  typedef typename MathBase::VectorXs VectorXs;
+  typedef typename MathBase::MatrixXs MatrixXs;
+  typedef pinocchio::RigidConstraintModelTpl<Scalar, 0> RigidConstraintModel;
+  typedef pinocchio::ProximalSettingsTpl<Scalar> ProximalSettings;
+
+  DifferentialActionModelConstraintFwdDynamicsTpl(
+      boost::shared_ptr<StateMultibody> state, boost::shared_ptr<ActuationModelAbstract> actuation,
+      const PINOCCHIO_STD_VECTOR_WITH_EIGEN_ALLOCATOR(RigidConstraintModel) & contacts,
```
cmastalli, line 58: It is not clear to me how we could disable/enable constraints without going through the creating of a different model and its data. This feature is really important for MPC applications.

include/crocoddyl/multibody/actions/constraint-fwddyn.hxx
```diff
@@ -0,0 +1,196 @@
...
+
+  d->Fu.noalias() = d->pinocchio.ddq_dtau * d->multibody.actuation->dtau_du;
+  costs_->calcDiff(d->costs, x, u);
+}
+
+template <typename Scalar>
+boost::shared_ptr<DifferentialActionDataAbstractTpl<Scalar> >
+DifferentialActionModelConstraintFwdDynamicsTpl<Scalar>::createData() {
+  boost::shared_ptr<DifferentialActionDataAbstractTpl<Scalar> > data =
+      boost::make_shared<DifferentialActionDataConstraintFwdDynamics>(this);
```
cmastalli, line 100: This won't compile with vectorization support. For that, we need to use `Eigen::aligned_allocator` as suggested below
```suggestions
  return boost::allocate_shared<Data>(Eigen::aligned_allocator<Data>(), this);
```

This code assumes that you have defined the `Data` type in the `hpp` file, which I highly suggest to do it as it is follows the Crocoddyl code style.

include/crocoddyl/multibody/actions/constraint-fwddyn.hxx
```diff
@@ -0,0 +1,196 @@
...
+      static_cast<DifferentialActionDataConstraintFwdDynamics*>(data.get());
+  pinocchio::initConstraintDynamics(pinocchio_, d->pinocchio, contacts_);
+
+  return data;
+}
+
+template <typename Scalar>
+bool DifferentialActionModelConstraintFwdDynamicsTpl<Scalar>::checkData(
+    const boost::shared_ptr<DifferentialActionDataAbstract>& data) {
+  boost::shared_ptr<DifferentialActionDataConstraintFwdDynamics> d =
```
cmastalli, line 111: Use `const &` to avoid create unnecessarily an owner, i.e.,
```suggestion
  const boost::shared_ptr<Data>& d =
```

(I am aware that there are a few places with this issues, it is in my TODO task to clean them.)

include/crocoddyl/multibody/actions/constraint-fwddyn.hxx
```diff
@@ -0,0 +1,196 @@
...
+}
+
+template <typename Scalar>
+void DifferentialActionModelConstraintFwdDynamicsTpl<Scalar>::set_armature(const VectorXs& armature) {
+  if (static_cast<std::size_t>(armature.size()) != state_->get_nv()) {
+    throw_pretty("Invalid argument: "
+                 << "The armature dimension is wrong (it should be " + std::to_string(state_->get_nv()) + ")");
+  }
+  pinocchio_.armature = armature;
+}
```
cmastalli, line 181: Let's remove this function  (also `get_armature`) as this function can directly obtain from `get_pinocchio`.
This avoid code redundancy.

bindings/python/crocoddyl/multibody/multibody.cpp
```diff
@@ -30,6 +30,9 @@ void exposeMultibody() {
   exposeDataCollectorImpulses();
   exposeDifferentialActionFreeFwdDynamics();
   exposeDifferentialActionContactFwdDynamics();
+#if PINOCCHIO_VERSION_AT_LEAST(2, 9, 0)
```
cmastalli, line 33: Remove this macro as it is not needed (we have implemented in the expose function).

cmastalli (changes requested):
Hey Rohan,

The PR is quite clean and nice to review! I have left a few comments, and below, I enlist the important points:
  1. How to handle active/inactive constraints?
  This is a very vital point, and a missing feature when compared with the contact-dynamics solution. I have provided some insights of how I would approach it in the comments.
  2. `Fx` derivatives doesn't consider actuation state derivatives.
  3. Use `Eigen::aligned_allocator` in data creation.
  4. Define the `quasiStatic` function.
  5. Use shared pointer for constraints in model constructor.
  6. Adapt the friction, wrench, cop residuals to be used in this diff-action as well (see their data constructor).

include/crocoddyl/multibody/actions/constraint-fwddyn.hxx
```diff
@@ -0,0 +1,196 @@
...
+
+template <typename Scalar>
+const boost::shared_ptr<ActuationModelAbstractTpl<Scalar> >&
+DifferentialActionModelConstraintFwdDynamicsTpl<Scalar>::get_actuation() const {
+  return actuation_;
+}
+
+template <typename Scalar>
+const std::vector<pinocchio::RigidConstraintModelTpl<Scalar, 0>,
```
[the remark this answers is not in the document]
jcarpent: It will imply memory allocation. You can change the active contacts by providing the vector of active contacts.
We can think to change, but this not the priority for now.

proyan:
This PR is meant only for the few pinocchio 2.9 users who needed support in crocoddyl. Since the API of pinocchio is bound to be changed, and I don't think I could spend more time on this PR, I would close this here, and keep the support in my fork.

Thanks Carlos for the useful feedback, I'll take care of your comments and suggestions

PR #980, opened 2021-07-29
Minor improvements: clean-up examples, throw_pretty in Python actuaction, and remove memory allocation in quasiStatic

cmastalli:
The changes on the examples are targeted to extend them along a set of cases that used inverse-dynamics (instead of the current forward dynamics).

Note that the inverse-dynamics based optimal control is still an ongoing work which will be released in Crocoddyl v2.


include/crocoddyl/multibody/actions/free-fwddyn.hxx
```diff
@@ -154,12 +154,14 @@ void DifferentialActionModelFreeFwdDynamicsTpl<Scalar>::quasiStatic(
...
-  actuation_->calcDiff(d->multibody.actuation, d->tmp_xstatic, VectorXs::Zero(nu_));
+  d->tmp_xstatic.tail(state_->get_nv()).setZero();
+  u.setZero();
+
+  pinocchio::rnea(pinocchio_, d->pinocchio, q, d->tmp_xstatic.tail(state_->get_nv()),
+                  d->tmp_xstatic.tail(state_->get_nv()));
+  actuation_->calc(d->multibody.actuation, d->tmp_xstatic, u);
+  actuation_->calcDiff(d->multibody.actuation, d->tmp_xstatic, u);
 
   u.noalias() = pseudoInverse(d->multibody.actuation->dtau_du) * d->pinocchio.tau;
```
jcarpent, line 166: There is a huge memory allocation here!

cmastalli: Hey @jcarpent! thank you so much for the feedback.

Do you refer to the `pseudoInverse` function?
In any case, could you let me know how should I remove the memory allocation that you pointed out?

Below, I have attached the code of the `pseudoInverse` function:
https://github.com/loco-3d/crocoddyl/blob/master/include/crocoddyl/core/utils/math.hpp#L20-L54

jcarpent: The memory allocation occurs within `pseudoInverse` (the SVD struct from Eigen) but also when returning the value par copy. The best action would be to store within the data struct the SVD struct together with the Matrix storage to contain the temporaries and the pseudo-inverse matrix.

jcarpent: If I were you, I would keep the current signature, while also providing a new one which does not perform any allocation.

cmastalli: > The memory allocation occurs within pseudoInverse (the SVD struct from Eigen)

I thought that the `static` keyword in the SVD structure is avoiding to allocate memory, am I right? 

> but also when returning the value par copy.

I see this point! If I could create another signature (for `pseudoInverse`)  were we pass the output inversion. And of course, we can store this output in the action-data structure.

> The best action would be to store within the data struct the SVD struct together with the Matrix storage to contain the temporaries and the pseudo-inverse matrix.

I agree that this could be a workaround. But if I am right above, then I would prefer to avoid storing the SVD structure in the action-data structure.

jcarpent: > I thought that the `static` keyword in the SVD structure is avoiding to allocate memory, am I right?

Not at all. You are mixing the different possible use of the `static` keyword.

cmastalli: > Not at all. You are mixing the different possible use cases of the static keyword.

Thanks again for the feedback.

> If I were you, I would keep the current signature, while also providing a new one which does not perform any allocation.

This might be possible by defining two different data structures, and then parsing internally it. It seems a complex solution, but it could be implemented if everyone like it. Note that we cannot create two signatures for `quasiStatic` as it is a virtual function (which, for instance, is used in `ShootingProblem`).

Would be fine for you if I handle the issue with the `pseudoInverse` mem-allocation in another PR? (request due to time constraints).

If you agree, then I will create an issue to do not forget this task :)

proyan: Quasi-static is normally not called during runtime, the use is mostly for warm-starting the solver, so the fix would not be that urgent.

proyan (approved):

include/crocoddyl/multibody/actions/free-fwddyn.hxx
cmastalli: I am closing this conversation after I have created the issue: https://github.com/loco-3d/crocoddyl/issues/981.

Thanks guys for the feedback. Let me know if there is something left in this PR.

cmastalli:
@wxmerkt -- do you know why the ROS-CIs are not passing?

wxmerkt:
The momentum cost test failed. I restarted it in case it was flaky. However, it's likely that we are already using Pinocchio 2.6.3 on the ROS-CI which has the fixed centroidal momentum derivatives in case that has an impact here or requires adjustments to the unit test.
`  	 30 - pybinds-momemtum_cost (Failed)`

cmastalli:
> The momentum cost test failed. I restarted it in case it was flaky. However, it's likely that we are already using Pinocchio 2.6.3 on the ROS-CI which has the fixed centroidal momentum derivatives in case that has an impact here or requires adjustments to the unit test.
> ` 30 - pybinds-momemtum_cost (Failed)`

I don't think this might be related with Pinocchio stuffs because the unittests were passing with any version of Pinocchio (including 2.6.3 as we can see in the details of the gitlab-ci).

What could be other sources of problems?

cmastalli:
> The momentum cost test failed. I restarted it in case it was flaky. However, it's likely that we are already using Pinocchio 2.6.3 on the ROS-CI which has the fixed centroidal momentum derivatives in case that has an impact here or requires adjustments to the unit test.
> ` 30 - pybinds-momemtum_cost (Failed)`

I tested locally, the error appears with Pinocchio 2.3.6 (i.e., after the changes in https://github.com/stack-of-tasks/pinocchio/pull/1474).

It seems that we don't need anymore these lines: https://github.com/loco-3d/crocoddyl/blob/devel/include/crocoddyl/multibody/residuals/centroidal-momentum.hxx#L50-L52. At least, the test is passing if I remove them.

@jcarpent -- I need your help here. Should I remove those lines?

jcarpent:
> @jcarpent -- I need your help here. Should I remove those lines?

Yes, you have to.

proyan:
Yes, I put these lines as a local solution for the bug. Since we have now fixed it in pinocchio as well, we should also update the minimum version to 2.6.3

proyan:
Also, it looks like LAAS-CNRS is missing from the copyright in all residual models. Since the residual-models were refactored from the original cost-models, the copyright should remain.

cmastalli:
> Also, it looks like LAAS-CNRS is missing from the copyright in all residual models. Since the residual-models were refactored from the original cost-models, the copyright should remain.

Done!

The code is ready to be merged; however, gitlab-ci is failing since we need to deploy 2.6.3 in robotpkg. @nim65s -- do you know when it is expected the new release.

cmastalli:
I will merge this PR as ROS-CI is checking all the unittests (and I don't want to wait for `robotpkg` release)

nim65s:
We could have used a version check instead of requiring the latest version.

cmastalli:
> We could have used a version check instead of requiring the latest version.

I indeed though about this possibility, but it seems more convenient to constrain everyone moving to another version as this is a bug.
We could stikk add the version check (as it was just merged), if you believe it is a good idea.

proyan:
@cmastalli Accepting PRs when CI is failing is never a good idea. It might not be the cleanest way forward, but it is better to leave a PR open than to accept it without the CI.

cmastalli:
> @cmastalli Accepting PRs when CI is failing is never a good idea. It might not be the cleanest way forward, but it is better to leave a PR open than to accept it without the CI.

There is redundancy, the ROS-CI is testing the same bits but with Pinocchio 2.6.3.

PR #984, opened 2021-08-03
Release/1.8.1

nim65s:
Maintenance release for pinocchio 2.6.3:
- fix Vx computation
- fix memory allocations
- fix linkage of the python library
- deprecate FramXX in constructors and frames.hpp
- cleanups

NB: The CI is expected to fail until pinocchio 2.6.3 gets a binary release on robotpkg, but this release would break master, so we have to release crocoddyl 1.8.1


cmastalli:
Thanks @nim65s :)

PR #987, opened 2021-08-16
Minor improvements in the actuation and control doxygen documentation

cmastalli:


cmastalli:
I believe all these are minor changes which do not require any review.
So, I would proceed with the merging.

PR #992, opened 2021-08-31
Added other force-based costs for contact fwd-dyn

cmastalli:


cmastalli:
@nim65s -- why gitlab-ci status is pending? By inspecting the jobs, I can see that all the mandatory jobs are run and passed

nim65s:
A request was lost. A send it again, it's OK now.

Issue #962, opened 2021-06-10
terminal cost computation


andreadelprete:
Hi, 

I just realized that crocoddyl handles the terminal cost in an unusual way, which is making my life difficult, and is also leading to useless computations. Typically the terminal cost in an optimal control problem depends only on the terminal state. In Crocoddyl instead, the terminal cost is computed [calling the `calc` method of an action model with zero control inputs](https://github.com/loco-3d/crocoddyl/blob/master/src/core/solvers/ddp.cpp#L335). This means that:
* the dynamics is evaluated even at the terminal state, which is not needed
* the terminal cost depends on the time step
* the terminal cost depends on what happens when you apply zero inputs at the terminal state

This thing gets even worse when using high-order integration schemes (which I am currently playing with) because the extra useless computation associated with the terminal state are even larger, and more importantly it's basically impossible to formulate two "equivalent" problems using different integration schemes.

If my understanding of the problem is correct, I suggest we fix this by introducing a new method of ActionModel to compute the terminal cost without integrating the dynamics.

proyan:
Hi Andrea,
Yes, you are right. We seem to be doing additional computations, which we don't have to do at the terminal state. Thanks for noticing this.
I'll add a simple `update(data, x)` function, that updates the data based on the current state value. I think it should solve this issue. Do you see anything else that needs to be done?

nmansard:
Discussing with Andrea: there is the variable enable_integration in the integration-euler action model. Set it to false, and you will solve the issue.
Yet it is maybe not clear for the user that the flag should be set to false for the terminal node. We should maybe check it and warn the user in the solver?

nmansard:
Could we keep the default behaviour in the nominal action model, to avoid to explicitly implement the terminal behaviour in new action classes (in particular when prototyping)?

cmastalli:
> For terminal model, we could simply replace https://github.com/loco-3d/crocoddyl/blob/master/src/core/solvers/ddp.cpp#L335 by an update function, and remove the `enable_integration` from IntegratedActionModel. This would let us have `dt=0` in runningModels as well.

I don't think we need to include an extra `update` function. We could use `calc(data, x)` for that, we just need to define it as a virtual function and to implement the extra bits in each diff/action model.

cmastalli:
This won't require to create an extra "terminal action" (which I also don't like the idea) or to adapt user code as the default `calc(data,x)` implementation will still make the job.

andreadelprete:
[...]

Issue #991, opened 2021-08-27
Crocoddyl Actuation Model with External Forces


EpicDuckPotato:
I posted an issue in pinocchio recently where my robot contains thrusters and revolute joints, where the relative placement of the thrusters with respect to the base frame varies based on the joint angles.

stack-of-tasks/pinocchio#1513

The conclusion was that I needed to treat the thrusters as external forces in the individual joint frames, then compute dtau_dx in my actuation model using pinocchio's computeABADerivatives. It would be helpful if there were an actuation model predefined where some actuators map to external forces on the robot.

PR #995, opened 2021-09-09
Improve terminal cost computation

cmastalli:
This is draft PR that aims to handle the issue discussed in #962. 
The main purpose of creating this early PR is to demonstrate that there are issues in the terminal-node computations as mentioned in https://github.com/loco-3d/crocoddyl/issues/962#issuecomment-915154943.

Currently, the PR contains
 - Virtualization of `calc(data, x)` and `calcDiff(data, x)` for (differential) action models.
 - Implementation of the these functions for the diff-lqr, lqr, and unicycle models.
 - Implementation of the these functions for the Euler and RK4 integrators.
 - Removed the hard-coded `enable_integration` boolean.
 - Unittests of `calcDiff(data, x)` for (differential) action models.
 - Removed memory allocation in diff-lqr and lqr models.

There is still significantly work to be done. Some of the tasks are:
- [x] Introduce the notion of terminal computation in cost, residual and actuation models
- [x] Implement terminal computations for cost models
- [x] Implement terminal computations for residual models
- [x] Implement terminal computations for actuation models
- [x] Implement proper terminal computation for free fwddyn
- [x] Implement proper terminal computation for contact fwddyn

We need to agree whether the `calc(data, x)` and `calcDiff(data, x)` should be virtual or pure virtual functions. I am inclined to make them pure virtual functions as justified in https://github.com/loco-3d/crocoddyl/issues/962#issuecomment-915154943.

Last but not least, I am keen to receive any further help on the abovementioned tasks. @andreadelprete and @proyan -- feel free to contribute to this PR, and if so, then please contact me first.


cmastalli:
This PR is ready to be reviewed!

I implemented the terminal computation of the contact and free forward dynamics based on the discussion in https://github.com/loco-3d/crocoddyl/issues/962.

I have also included doxygen documentation of the free/contact/impulse fwddyn models. Now, RK4 integrator with RK4 parametrization works well (fyi @andreadelprete)

include/crocoddyl/core/action-base.hpp
```diff
@@ -87,6 +87,17 @@ class ActionModelAbstractTpl {
   virtual void calc(const boost::shared_ptr<ActionDataAbstract>& data, const Eigen::Ref<const VectorXs>& x,
                     const Eigen::Ref<const VectorXs>& u) = 0;
 
+  /**
+   * @brief Compute the total cost value for nodes that depends only on the state
+   *
+   * It updates the total cost when the next state remains the same to current one. This function is commonly used in
+   * the terminal nodes of an optimal control problem.
```
andreadelprete, line 94: Are there any other use cases of this method besides that of terminal nodes?

include/crocoddyl/core/actions/diff-lqr.hxx
```diff
@@ -59,6 +59,22 @@ void DifferentialActionModelLQRTpl<Scalar>::calc(const boost::shared_ptr<Differe
...
+                                                 const Eigen::Ref<const VectorXs>& x) {
+  if (static_cast<std::size_t>(x.size()) != state_->get_nx()) {
+    throw_pretty("Invalid argument: "
+                 << "x has wrong dimension (it should be " + std::to_string(state_->get_nx()) + ")");
+  }
+
+  if (drift_free_) {
+    data->xout.setZero();
+  } else {
+    data->xout = f0_;
```
andreadelprete, line 73: Is it really necessary to update `xout`? Is it going to be used?

include/crocoddyl/core/actions/lqr.hxx
```diff
@@ -51,6 +51,22 @@ void ActionModelLQRTpl<Scalar>::calc(const boost::shared_ptr<ActionDataAbstract>
...
+                                     const Eigen::Ref<const VectorXs>& x) {
+  if (static_cast<std::size_t>(x.size()) != state_->get_nx()) {
+    throw_pretty("Invalid argument: "
+                 << "x has wrong dimension (it should be " + std::to_string(state_->get_nx()) + ")");
+  }
+
+  if (drift_free_) {
+    data->xnext = x;
+  } else {
+    data->xnext = x + f0_;
```
andreadelprete, line 65: Same comment as above: is it really necessary to update `xnext`?

include/crocoddyl/core/actions/unicycle.hxx
```diff
@@ -39,6 +39,21 @@ void ActionModelUnicycleTpl<Scalar>::calc(const boost::shared_ptr<ActionDataAbst
...
+template <typename Scalar>
+void ActionModelUnicycleTpl<Scalar>::calc(const boost::shared_ptr<ActionDataAbstractTpl<Scalar> >& data,
+                                          const Eigen::Ref<const VectorXs>& x) {
+  if (static_cast<std::size_t>(x.size()) != state_->get_nx()) {
+    throw_pretty("Invalid argument: "
+                 << "x has wrong dimension (it should be " + std::to_string(state_->get_nx()) + ")");
+  }
+  Data* d = static_cast<Data*>(data.get());
+
+  d->xnext = x;
```
andreadelprete, line 51: same comment here

include/crocoddyl/core/actuation-base.hpp
```diff
@@ -71,6 +68,17 @@ class ActuationModelAbstractTpl {
   virtual void calc(const boost::shared_ptr<ActuationDataAbstract>& data, const Eigen::Ref<const VectorXs>& x,
                     const Eigen::Ref<const VectorXs>& u) = 0;
 
+  /**
+   * @brief Compute the actuation signal from the state point only
+   *
+   * It updates the actuation signal based on the state only. This function is commonly used in
+   * the terminal nodes of an optimal control problem.
```
andreadelprete, line 75: I'm probably missing something here, but I don't get why we should need to compute the actuation signal in the terminal node, since in the terminal node there's no actuation defined in the OCP.

include/crocoddyl/core/diff-action-base.hpp
```diff
@@ -83,6 +83,18 @@ class DifferentialActionModelAbstractTpl {
   virtual void calc(const boost::shared_ptr<DifferentialActionDataAbstract>& data, const Eigen::Ref<const VectorXs>& x,
                     const Eigen::Ref<const VectorXs>& u) = 0;
 
+  /**
+   * @brief Compute the total cost value for nodes that depends only on the state
+   *
+   * It updates the total cost when system acceleration as zero. This function is commonly used in the
```
andreadelprete, line 89: Why do you say that the acceleration is zero?!

include/crocoddyl/core/integrator/euler.hpp
```diff
@@ -80,6 +79,17 @@ class IntegratedActionModelEulerTpl : public IntegratedActionModelAbstractTpl<_S
   virtual void calc(const boost::shared_ptr<ActionDataAbstract>& data, const Eigen::Ref<const VectorXs>& x,
                     const Eigen::Ref<const VectorXs>& u);
 
+  /**
+   * @brief Integrate the total cost value for nodes that depends only on the state using symplectic Euler scheme
+   *
+   * It integrates the total cost and defines the next state as the current one. This function is commonly used in the
```
andreadelprete, line 85: It shouldn't integrate the cost because it's a terminal cost.

include/crocoddyl/core/integrator/euler.hpp
```diff
@@ -90,6 +100,17 @@ class IntegratedActionModelEulerTpl : public IntegratedActionModelAbstractTpl<_S
   virtual void calcDiff(const boost::shared_ptr<ActionDataAbstract>& data, const Eigen::Ref<const VectorXs>& x,
                         const Eigen::Ref<const VectorXs>& u);
 
+  /**
+   * @brief Compute the partial derivatives of the symplectic Euler cost integrator
```
andreadelprete, line 104: Same comment as above.

include/crocoddyl/core/integrator/euler.hxx
```diff
@@ -40,32 +40,35 @@ void IntegratedActionModelEulerTpl<Scalar>::calc(const boost::shared_ptr<ActionD
...
+  if (static_cast<std::size_t>(x.size()) != state_->get_nx()) {
+    throw_pretty("Invalid argument: "
+                 << "x has wrong dimension (it should be " + std::to_string(state_->get_nx()) + ")");
   }
+  Data* d = static_cast<Data*>(data.get());
 
-  // Updating the cost value
+  differential_->calc(d->differential, x);
+  d->dx.setZero();
+  d->xnext = x;
```
andreadelprete, line 70: Same comment

include/crocoddyl/core/integrator/rk4.hpp
```diff
@@ -77,6 +77,17 @@ class IntegratedActionModelRK4Tpl : public IntegratedActionModelAbstractTpl<_Sca
   virtual void calc(const boost::shared_ptr<ActionDataAbstract>& data, const Eigen::Ref<const VectorXs>& x,
                     const Eigen::Ref<const VectorXs>& u);
 
+  /**
+   * @brief Integrate the total cost value for nodes that depends only on the state using RK4 scheme
+   *
+   * It integrates the total cost and defines the next state as the current one. This function is commonly used in the
```
andreadelprete, line 83: The cost is not integrated actually because it's a terminal cost. Indeed the implementation of this method should be the same for all integration schemes and could be implemented just once in the base class, couldn't it?

include/crocoddyl/core/integrator/rk4.hpp
```diff
@@ -87,6 +98,17 @@ class IntegratedActionModelRK4Tpl : public IntegratedActionModelAbstractTpl<_Sca
   virtual void calcDiff(const boost::shared_ptr<ActionDataAbstract>& data, const Eigen::Ref<const VectorXs>& x,
                         const Eigen::Ref<const VectorXs>& u);
 
+  /**
+   * @brief Compute the partial derivatives of the RK4 cost integrator
```
andreadelprete, line 102: Same comment as above

include/crocoddyl/multibody/actions/contact-fwddyn.hxx
```diff
@@ -103,6 +103,28 @@ void DifferentialActionModelContactFwdDynamicsTpl<Scalar>::calc(
...
+  if (static_cast<std::size_t>(x.size()) != state_->get_nx()) {
+    throw_pretty("Invalid argument: "
+                 << "x has wrong dimension (it should be " + std::to_string(state_->get_nx()) + ")");
+  }
+
+  Data* d = static_cast<Data*>(data.get());
+  const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> q = x.head(state_->get_nq());
+  const Eigen::VectorBlock<const Eigen::Ref<const VectorXs>, Eigen::Dynamic> v = x.tail(state_->get_nv());
+
+  pinocchio::computeAllTerms(pinocchio_, d->pinocchio, q, v);
```
andreadelprete, line 118: This probably computes more than needed because it also computes some dynamics quantities such as M and h.

andreadelprete (commented):
Hi @cmastalli , thanks for the great work, and sorry I wasn't able to contribute but I really didn't have time. I've left a few comments, nothing major, just pointing out some potential improvements to the code and the doc.

PS: Really glad to hear that now RK4 is working fine!

include/crocoddyl/core/action-base.hpp
cmastalli: I don't think so.

include/crocoddyl/core/actuation-base.hpp
cmastalli: We defined the actuation signal as a function of `x` and `u`.
Currently, we have implemented actuation models that depends on `u` only.

The x dependency is a sensible approach as it could help issues such as https://github.com/loco-3d/crocoddyl/issues/991 (see Pinocchio issues as well).

This is not conventional choice, but yet useful.

Is this answer your question? Can I mark this conversation as resolved?

include/crocoddyl/core/integrator/rk4.hpp
cmastalli: At the moment it is not possible as, for instance, RK4 uses a std vector of differential data objects.
We could restructure the RK4 as:
  - data.differential (differential for 0 node)
  - data.differentials

But, this API was not convinced you last time.

I could handle this upon agreement (but in another PR). Would you change the RK4 data API for this reason?

include/crocoddyl/core/action-base.hpp
cmastalli: I will mark this conversation as resolved. Please reopen it if this is the case

include/crocoddyl/core/actions/diff-lqr.hxx
cmastalli: No, we don't have to. Apologies, I forgot to modify this bit.
Now it is done!

include/crocoddyl/core/actions/lqr.hxx
cmastalli: Same answer here.
Done!

include/crocoddyl/core/actions/unicycle.hxx
cmastalli: Also here
Done!

include/crocoddyl/core/integrator/euler.hxx
cmastalli: done

include/crocoddyl/core/integrator/rk4.hpp
cmastalli: Done the modification of the documentation. Thanks for catching this inaccurate documentation.

I will mark this conversation as resolved. Note that the RK4 API won't be changed in this PR, but I am still looking forward for your feedback

include/crocoddyl/core/integrator/rk4.hpp
cmastalli: Done

cmastalli: Also for the Euler integrator

include/crocoddyl/multibody/actions/contact-fwddyn.hxx
cmastalli: Hey @andreadelprete 

Your observation is right, but it is still not convenient for predictive control application. Let me explain this claim briefly.

In MPC applications, you might receded the running model. So, the terminal node will become the last running one and so on. Therefore, we will need to compute all the terms as we reuse the `calc` results (performed in the forward pass) in the next solver iteration: https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solvers/ddp.cpp#L127

(Note that we don't run `calc` before `calcDiff`)

Due to this justification, I won't follow this suggestion. We could open a discussion on this topic, if you think we need to dig more on it.

include/crocoddyl/core/action-base.hpp
andreadelprete: What I meant with my question is that, if there are no other use cases, instead of writing "this function is commonly used in the terminal nodes" we could write "this function is used in the terminal nodes", because that's the only use case.

include/crocoddyl/core/actuation-base.hpp
andreadelprete: No, you didn't answer my question. :)
I get it that the actuation signal can depend on both x and u, that's a good idea. What I don't get is why we would need an actuation signal at all for the terminal nodes, since the dynamics doesn't have to be computed and you only need to evaluate the terminal cost.

include/crocoddyl/core/integrator/rk4.hpp
andreadelprete: Yes, I agree this would mess up the RK4 data API. At the same time, not doing it could lead to a replication of code every time we add a new integration scheme. I'm not satisfied by any of these options. The clean solution (to address in another PR) would be to split the ActionModel class into two classes: RunningActionModel and TerminalActionModel. The integration schemes would then be children of RunningActionModel and would only implement `calc(data, x, u)`. The classes children of TerminalActionModel instead would only implement `calc(data, x)`, and would not perform any integration. This would also allow us to modify the OCP class API so that the last action model specified by the user be forced to be a TerminalActionModel.

include/crocoddyl/multibody/actions/contact-fwddyn.hxx
andreadelprete: It's fine by me. In any case it's a very small computation time.

include/crocoddyl/core/diff-action-base.hpp
```diff
@@ -86,8 +86,8 @@ class DifferentialActionModelAbstractTpl {
   /**
    * @brief Compute the total cost value for nodes that depends only on the state
    *
-   * It updates the total cost when system acceleration as zero. This function is commonly used in the
-   * terminal nodes of an optimal control problem.
+   * It updates the total cost and the system acceleration is not updated as it is expected to be zero. This function
```
andreadelprete, line 89: I still don't get why the system acceleration should be "expected to be zero". Since the control is not defined in the terminal node then also the acceleration is undefined.

include/crocoddyl/core/integrator/rk4.hpp
cmastalli: I don't think this is a good idea. We exchanged terminal nodes (both model and data) with running nodes in MPC applications.

Furthermore, this will replicate more code than in the integrators. For instance, we would need to create for each action class two action classes, i.e., one for running, another one for terminal.

Finally, this will completely break the API.

andreadelprete: I knew this suggestion would be widely unpopular! :)
I agree it would break the API, but that doesn't mean it's not a good design choice.

I think the exchange of terminal nodes with running nodes in MPC applications could be addressed somehow. Not sure how, but there could be a way.

I'm not sure this would replicate much code. Probably we would just need a single TerminalActionModel class implementation because the TerminalActionModel is extremely lightweight, it's basically just a cost function that depends on state only. It has no dynamics and no integration scheme. Or am I missing something?

include/crocoddyl/core/diff-action-base.hpp
andreadelprete: Just this one: https://github.com/loco-3d/crocoddyl/pull/995#discussion_r708269942

include/crocoddyl/core/actuation-base.hpp
cmastalli: You're right.
I have fixed this, and soon I will push it.

include/crocoddyl/core/action-base.hpp
cmastalli: I see your point. I will improve the documentation as suggested

cmastalli:
Thanks @andreadelprete for the nice review!

I have pushed the latest suggested changes in the actuation model and documentation.
I believe this PR could be merged (after passing all the jobs)

PR #998, opened 2021-09-14
Enable profiler at runtime

cmastalli:
This PR proposes a method for enabling at runtime the profiler. This will be something like this:
```c++
  getProfiler().enable_profiler();
 \\ run algorithm
  getProfiler().report_all();
```
or
```python
crocoddyl.enable_profiler()
# run algorithm
crocoddyl.stop_watch_report(5)
```

This simplifies at lot the performance analysis.

Additionally, this PR includes other macros for profiling DDP, FDDP and BoxDDP solvers.

### DIFF
--- a/src/core/utils/stop-watch.cpp
+++ b/src/core/utils/stop-watch.cpp
@@ -42,11 +42,17 @@ Stopwatch& getProfiler() {
   return s;
 }
 
-Stopwatch::Stopwatch(StopwatchMode _mode) : active(true), mode(_mode) {
+Stopwatch::Stopwatch(StopwatchMode _mode) : active(true), mode(_mode), profiler_active(false) {
   records_of = new map<string, PerformanceData>();
 }
 
 Stopwatch::~Stopwatch() { delete records_of; }
+
+void Stopwatch::enable_profiler() { profiler_active = true; }
+
+void Stopwatch::disable_profiler() { profiler_active = false; }
+
+bool Stopwatch::profiler_status() { return profiler_active; }
 
 void Stopwatch::set_mode(StopwatchMode new_mode) { mode = new_mode; }
 
--- a/bindings/python/crocoddyl/core/utils/stop-watch.cpp
+++ b/bindings/python/crocoddyl/core/utils/stop-watch.cpp
@@ -15,6 +15,10 @@ using namespace boost::python;
 namespace crocoddyl {
 namespace python {
 
+void enable_report() { getProfiler().enable_profiler(); }
+
+void disable_report() { getProfiler().disable_profiler(); }
+
 void stop_watch_report(int precision) { getProfiler().report_all(precision); }
 
 long double stop_watch_get_average_time(const std::string& perf_name) {
@@ -32,6 +36,10 @@ long double stop_watch_get_total_time(const std::string& perf_name) { return get
 void stop_watch_reset_all() { getProfiler().reset_all(); }
 
 void exposeStopWatch() {
+  bp::def("enable_profiler", enable_report, "Enable the profiler report.");
+
+  bp::def("disable_profiler", enable_report, "Disable the profiler report.");
+
   bp::def("stop_watch_report", stop_watch_report, "Report all the times measured by the shared stop-watch.");
 
   bp::def("stop_watch_get_average_time", stop_watch_get_average_time,
--- a/include/crocoddyl/core/optctrl/shooting.hxx
+++ b/include/crocoddyl/core/optctrl/shooting.hxx
@@ -10,6 +10,7 @@
 #ifdef CROCODDYL_WITH_MULTITHREADING
 #include <omp.h>
 #endif  // CROCODDYL_WITH_MULTITHREADING
+#include "crocoddyl/core/utils/stop-watch.hpp"
 
 namespace crocoddyl {
 
@@ -149,6 +150,7 @@ Scalar ShootingProblemTpl<Scalar>::calc(const std::vector<VectorXs>& xs, const s
     throw_pretty("Invalid argument: "
                  << "us has wrong dimension (it should be " + std::to_string(T_) + ")");
   }
+  START_PROFILER("ShootingProblem::calc");
 
 #ifdef CROCODDYL_WITH_MULTITHREADING
 #pragma omp parallel for num_threads(nthreads_)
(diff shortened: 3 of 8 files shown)


src/core/utils/stop-watch.cpp
```diff
@@ -42,12 +42,18 @@ Stopwatch& getProfiler() {
...
 
+void Stopwatch::enable_profiler() { profiler_active = true; }
+
+void Stopwatch::disable_profiler() { profiler_active = false; }
+
+bool Stopwatch::profiler_status() { return profiler_active; }
+
 void Stopwatch::set_mode(StopwatchMode new_mode) { mode = new_mode; }
 
 bool Stopwatch::performance_exists(string perf_name) { return (records_of->find(perf_name) != records_of->end()); }
```
jcarpent, line 59: Weird.

jcarpent (commented):
A small comment is that profiling is itself changing the execution timings .

src/core/utils/stop-watch.cpp
cmastalli: That piece of the code was included by @andreadelprete. So, he will the best to answer/handle your concern.

Btw, could you tell us what exactly you mean with "weird"?

cmastalli:
> A small comment is that profiling is itself changing the execution timings .

Yes, that the reason why the profiling is disable by default. Furthermore, with this approach, it shouldn't affect the code performance if the profiler is disable.

src/core/utils/stop-watch.cpp
jcarpent: This code is performing a copy, which is weird for a profiler ;)

jcarpent:
> Yes, that the reason why the profiling is disable by default. Furthermore, with this approach, it shouldn't affect the code performance if the profiler is disable.

What I wanted to mean here is that for small operations, profiling them will mean nothing, as the timers involved by the profiler might take more time to be enabled than the process itself.

cmastalli:
> What I wanted to mean here is that for small operations, profiling them will mean nothing, as the timers involved by the profiler might take more time to be enabled than the process itself.

I see, would you provide an example of what you consider "small operations"?

For instance, it seems useful to quantify operations such as "compute Qxx". Do you agree or not with it?

jcarpent:
> I see, would you provide an example of what you consider "small operations"?

This is in fact the reverse. You need first to evaluate the impact of your profiler on your timings.

src/core/utils/stop-watch.cpp
cmastalli: As far I understand, this function is used when we want to "report" *(i.e., print) the results of the profilers. It shouldn't affect the estimation of the computation time.

I will mark this conversation as resolved. Please reopen it if you believe so.

In any case, thanks for pointing out this bit.

proyan:
Thanks for the feature Carlos.
I have two main comments:
1) Regarding the API, start_profiler and stop_profiler seem to be the natural extension of the inbuilt macros. I would suggest that we replace enable_profiler and disable_report etc by those.
2) Even though the feature is useful, we are adding a bunch of conditional operators. Intuitively, this means that the solver timings would suffer. Also, more conditional operators put onus on the optimization done by the compiler as well, since some might understand this is meant to be harmless, but others might not. Of course, all depends on how much they suffer.

cmastalli:
Hey @proyan -- thank you for reviewing the code, I need a little of clarification before working on your suggestions. 

>     1. Regarding the API, start_profiler and stop_profiler seem to be the natural extension of the inbuilt macros. I would suggest that we replace enable_profiler and disable_report etc by those.

I am not sure what you suggest to replace. `START_PROFILER` and `STOP_PROFILER` are macros used to record the computation taken between those. You also associated a tag name that will be used when the ask for the report. Both macros were introduced by @andreadelprete.

Instead, `enable_profiler()` and `disable_profiler()` are functions that I introduced to be able to enable the profiler at runtime.

So, I understand correctly, you suggested to rename `enable_profiler()` and `disable_profiler()` functions by `start_profiler()` and `stop_profiler()`, respectively. Is this correct? If not, could you give more details?




>     2\. Even though the feature is useful, we are adding a bunch of conditional operators. Intuitively, this means that the solver timings would suffer. Also, more conditional operators put onus on the optimization done by the compiler as well, since some might understand this is meant to be harmless, but others might not. Of course, all depends on how much they suffer.

This is sensible and important topic, thanks for bringing in to the table (I also have the same concerns). This PR proposes to use *ternary operator* in the `START_PROFILER` and `STOP_PROFILER` macros: https://github.com/cmastalli/crocoddyl/blob/topic/profiler/include/crocoddyl/core/utils/stop-watch.hpp#L37-L38.

Even if the profiler is disable (i.e., `getProfiler().profiler_status()` is equals false), we are running a *ternary operator* and returning `void()` thing. I am not sure if this might affect solver timings, what do you think?

If this affects the solver timings, do you have another idea how to enable the profiler at runtime?

proyan:
There is an `enable_profiler`, and a `start_profiler`, where what the enable_profiler does is call start_profiler. I think we can simplify the API a bit.

Regarding the second comment, conditional operators add branchings which may or may not affect the performance here. It would be good to check.

cmastalli:
> There is an `enable_profiler`, and a `start_profiler`, where what the enable_profiler does is call start_profiler. I think we can simplify the API a bit.

Apologies, I am still confused and need further clarification.

Note that  `ENABLE_PROFILER` calls `profiler_status()`, which can be enable (True) or disable (False). Then, `enable_profiler()` is design to be manually call by the user (if he/she wishes to collect the timing data). Finally, the user also needs to call `report_all()` (or other functions to report specific things) when he/she wishes to get the profiler results.



> Regarding the second comment, conditional operators add branchings which may or may not affect the performance here. It would be good to check.

I will do it asap. Any particular suggestion for making a fair analysis?

proyan:
Ah sorry you are right. I misread profiler_status and start_profiler.

> I will do it asap. Any particular suggestion for making a fair analysis?

I would recommend trying it with both gcc and clang, since compiler optimization would matter here.

cmastalli:
> I would recommend trying it with both gcc and clang, since compiler optimization would matter here.

I use the walking ANYmal problem (with 107 nodes) to analysis the computational load of
 1. conditional operators
 2. profiler load
in both compilers: clang and gcc.
The computation time running these 100 problems (without profiler code) is about 11 and 13 secs, respectively.

Results on both compilers suggest that the conditional operator defined in the `START_PROFILER` and `STOP_PROFILER` doesn't affect the computation time.

Furthermore, enabling the profiler adds (as expected) around 250 and 470 msec of extra computation for the clang and gcc compilers, respectively.

Both results support the proposed choice in this PR, i.e., conditional operators defined inbuild macros which allows us to enable/disable the profiler at runtime.

cmastalli:
This PR is ready to be merged IMHO, is there any left to be improved?

proyan (approved):

Issue #986, opened 2021-08-14
Prismatic root_joint - controlling the Robot leg on the teststand


VasilyRakche:
### Setup
I am trying to create a good  control for the robotic leg with the teststand: 
<img src="https://user-images.githubusercontent.com/30003101/129442923-f32df9b0-5f11-4276-b568-3efe29c4d5d5.png" width="150" height="400">
In particular I want to create some interesting jumping motions so I can test them on the stand.

### Problem 
I have set up the crocoddyl and been able to run simulation with my urdf model with some modifications I did to the _**"Jumping"**_ of the [quadrupedal_gaits.py](https://github.com/loco-3d/crocoddyl/blob/9e79f80c6d52b2c6c000106570811d449341cad3/examples/quadrupedal_gaits.py) example : 

https://user-images.githubusercontent.com/30003101/129442784-baf327a0-ae41-42e8-8d76-36e3f2a26ae4.mp4

The problem here is that I want to constraint movement of the body to only be along z (vertical) axis, so I could focus on other parts of the problem and maybe make the algorithm work faster. Also here the algorithm Fails since it can not track referenced COM, which shouldn't be moving along x and y axis. 

### I have tried
Without changing urdf description, I have tried to use `pinocchio.JointModelPZ()` when using  `BuildFromUrdf`. I also  decided to keep the `ActuationModelFloatingBase` - I think it should work. If I understood [definition right,](https://github.com/loco-3d/crocoddyl/blob/9e79f80c6d52b2c6c000106570811d449341cad3/bindings/python/crocoddyl/multibody/actuations/floating-base.cpp#L21) it just doesnt try to control the `root_joint` as can be seen in [this line of the code](
https://github.com/loco-3d/crocoddyl/blob/9e79f80c6d52b2c6c000106570811d449341cad3/include/crocoddyl/multibody/actuations/floating-base.hpp#L30). `BuildFromUrdf` did its job correctly regarding defining the number of DOF but when running the robot is not moving at all.  What I am getting can be seen here:

https://user-images.githubusercontent.com/30003101/129443498-b109feb3-fd87-43a7-b7dc-d31cd55bfe9b.mp4

### Check out implementation
You can check it out by running the `fore_jumping.py` in the [z_constraint branch](https://github.com/Legged-robot/control_and_simulation/tree/z_constraint) (dont forget to check out to the **_develop_** branch in the submodule for the right urdf and srdf) and also to the **_z_constraint_** branch in original module. 
[...]

PR #1000, opened 2021-09-17
Manage 2d contact friction in xz plane

VasilyRakche:
In reference of #986 problem I have added some changes in order to have 2d contact taken into account for residual value calculation (before it was only possible for at least 3d contact). Drawbacks:

1. This solution works for xz plane for now (now I am able to generate motion with one legged robot stand):
image 
I am not sure how to make it more general. I would assume that for other cases I just need to take different columns of the A matrix, but not sure what would be the best way to get the axis of the plane of interest.

2. Maybe formatting is wrong since docker container image from [Developer-Guide](https://github.com/loco-3d/crocoddyl/wiki/Developer-Guide) would change almost all files in the repo.

### DIFF
--- a/include/crocoddyl/multibody/residuals/contact-friction-cone.hxx
+++ b/include/crocoddyl/multibody/residuals/contact-friction-cone.hxx
@@ -50,12 +50,25 @@ void ResidualModelContactFrictionConeTpl<Scalar>::calcDiff(const boost::shared_p
   const MatrixXs& df_dx = d->contact->df_dx;
   const MatrixXs& df_du = d->contact->df_du;
   const MatrixX3s& A = fref_.get_A();
-  if (d->more_than_3_constraints) {
-    data->Rx.noalias() = A * df_dx.template topRows<3>();
-    data->Ru.noalias() = A * df_du.template topRows<3>();
-  } else {
+  
+  switch (d->contact_type) {
+  case Contact2D: {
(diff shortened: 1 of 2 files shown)


include/crocoddyl/multibody/residuals/contact-friction-cone.hxx
```diff
@@ -50,12 +50,25 @@ void ResidualModelContactFrictionConeTpl<Scalar>::calcDiff(const boost::shared_p
...
   const MatrixX3s& A = fref_.get_A();
-  if (d->more_than_3_constraints) {
-    data->Rx.noalias() = A * df_dx.template topRows<3>();
-    data->Ru.noalias() = A * df_du.template topRows<3>();
-  } else {
+  
+  switch (d->contact_type) {
+  case Contact2D: {
+    MatrixXs A_2d(A.rows(), 2);
+    A_2d << A.col(0), A.col(2);
```
cmastalli, line 57: This allocates memory. Please remove the intermediate matrix, and makes the corresponding multiplications for computing `Rx`, and `Ru`.

cmastalli (changes requested):
We also need to include an unittest.
I suggest you to replace the 3d contact by 2d one here: https://github.com/loco-3d/crocoddyl/blob/master/unittest/factory/diff_action.cpp#L171-L184

PR #1002, opened 2021-09-23
Fixed computation of coefficient in  poly-two-rk4 parametrization

cmastalli:
Hey @andreadelprete!

I noticed we have an error in the computation of `c[1]`, which it is fixed in this PR.
Additionally, this fixes the numerical-convergence issues on the MPC algorithm which I reported to you.

Thanks for sharing your notes.
I will later (no in this PR) work on RK3 integrator and its parametrization.

### DIFF
--- a/include/crocoddyl/core/controls/poly-two-rk4.hxx
+++ b/include/crocoddyl/core/controls/poly-two-rk4.hxx
@@ -29,8 +29,8 @@ void ControlParametrizationModelPolyTwoRK4Tpl<Scalar>::calc(
   const Eigen::VectorBlock<const Eigen::Ref<const VectorXs> >& p2 = u.tail(nw_);
   d->tmp_t2 = t * t;
   d->c[2] = 2 * d->tmp_t2 - t;
-  d->c[1] = -2 * d->c[2];
-  d->c[0] = 1 - 2 * t + d->c[2];
+  d->c[1] = -2 * d->c[2] + 2 * t;
+  d->c[0] = d->c[2] - 2 * t + 1;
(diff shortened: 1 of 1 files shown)


hrp2-14:
Hi ! This project doesn't usually accept pull requests on master. If this wasn't intentionnal, you
can change the base branch of this pull request to devel (No need to close it for that). Best, a bot.

andreadelprete:
Hi @cmastalli , well done! I confirm there was an error. We should probably extend the unit tests so that these errors are catched.

cmastalli:
> Hi @cmastalli , well done! I confirm there was an error. We should probably extend the unit tests so that these errors are catched.

Yes, we should indeed test this. But, to be honest, I don't know how to do it as the error was in the `calc` function.
Note that we are testing `calcDiff` and the Jacobian-multiplication operators:
 - https://github.com/loco-3d/crocoddyl/blob/devel/unittest/test_controls.cpp#L22
 - https://github.com/loco-3d/crocoddyl/blob/devel/unittest/test_controls.cpp#L44
 - https://github.com/loco-3d/crocoddyl/blob/devel/unittest/test_controls.cpp#L71

If you have an idea how to unit-test the `calc` functions, then please let me know.

Btw, I would like to discuss with you again about the control-parametrization API.
At the moment, I don't see the justification of having two operators: one for the Jacobian, another one for its transpose.
Is there something that I am missing?

(Please note that this development has not been deployed in the master branch, so, we could easily adjust the API)

andreadelprete:
Maybe we can test that the parametrization PolyTwoRK4 gives p0 for t=0, p1 for t=0.5 and p2 for t=1.

PR #1004, opened 2021-09-29
RK2 and RK3

andreadelprete:
This PR is a starting point for implementing the integration schemes RK2 and RK3. The implementation I am providing was working with the previous API of the integratedActionModel, which has then been changed in a previous PR after discussion with @cmastalli and @proyan. Now @cmastalli has agreed to adapt this code to the new API. This PR also contains a new ControlParametrizationModel that is a second-order polynomial customized to work with RK3.


cmastalli:
@andreadelprete -- it seems we can replicate `IntegratedActionRK2,3,4` with a single `IntegratedActionRK`?
Same comment for the `ControlParametrizationRK2,3,4`
Do you agree with me?

This will reduce the amount of code and compilation time.

include/crocoddyl/core/integrator/euler.hpp
```diff
@@ -85,7 +85,7 @@ class IntegratedActionModelEulerTpl : public IntegratedActionModelAbstractTpl<_S
    * It computes the total cost and defines the next state as the current one. This function is used in the
    * terminal nodes of an optimal control problem.
    *
-   * @param[in] data  Action data
+   * @param[in] data  Symplectic Euler data
```
jcarpent, line 88: Why symplectic?

include/crocoddyl/core/integrator/rk4.hpp
```diff
@@ -83,7 +83,7 @@ class IntegratedActionModelRK4Tpl : public IntegratedActionModelAbstractTpl<_Sca
    * It computes the total cost and defines the next state as the current one. This function is used in the
    * terminal nodes of an optimal control problem.
    *
-   * @param[in] data  Action data
+   * @param[in] data  Semi-implicit RK4 data
```
jcarpent, line 86: Why semi-implicit?

cmastalli: Sorry, I was copying-pasting code.
I would remove "semi-implicit" asap.

cmastalli:
> @andreadelprete -- it seems we can replicate `IntegratedActionRK2,3,4` with a single `IntegratedActionRK`? Same comment for the `ControlParametrizationRK2,3,4` Do you agree with me?
> 
> This will reduce the amount of code and compilation time.

Another option could be to create sort of base class, and avoid code duplication.
But this won't reduce the compilation time (not at least significantly).

include/crocoddyl/core/integrator/euler.hpp
cmastalli: The Euler integrator is indeed symplectic.
We have implemented this exactly integrator: https://en.wikipedia.org/wiki/Semi-implicit_Euler_method

include/crocoddyl/core/integrator/rk4.hpp
cmastalli: Fixed

andreadelprete:
> @andreadelprete -- it seems we can replicate `IntegratedActionRK2,3,4` with a single `IntegratedActionRK`? Same comment for the `ControlParametrizationRK2,3,4` Do you agree with me?
> 
> This will reduce the amount of code and compilation time.

Yes, it's probably possible. The code of RK2/3/4 is very similar.

andreadelprete:
@cmastalli let me know if and when you'd like me to review the PR

cmastalli:
@andreadelprete -- This PR is ready to review. However, I assumed that we need to implement a RK2 parametrization here: https://github.com/andreadelprete/crocoddyl/blob/devel/include/crocoddyl/core/controls/poly-two-rk.hxx#L34
Could you let me know the coefficients for this parametrization?

Note that I have merged the integrators and parametrizations as discussed. The RK4 integrator is also deprecated.

proyan:
Hi Carlos, 
Thanks for the effort. Before I start reviewing, just a minor comment: enum rktype could be used as a template parameter for IAMRK. The API would then remain the same, and ```typedef typename IntegratedActionModelRKTpl<4> IntegratedActionModelRK4Tpl``` could be declated in rk4.hpp

andreadelprete:
> @andreadelprete -- This PR is ready to review. However, I assumed that we need to implement a RK2 parametrization here: https://github.com/andreadelprete/crocoddyl/blob/devel/include/crocoddyl/core/controls/poly-two-rk.hxx#L34 Could you let me know the coefficients for this parametrization?

Ok thanks, I'll review this tomorrow. For RK2, since it only computes the dynamics at t=0 and t=0.5 it's useless to use it with a control parametrization with more than 2 parameters, so a linear parametrization is sufficient (which is already implemented in PolyOne)

andreadelprete (commented):
Hi @cmastalli , I've just finished reviewing. Overall everything seems fine to me. I've fixed a few mistakes in the documentation, which I'm gonna push in a second. I approve the PR.

unittest/factory/integrator.cpp
```diff
@@ -50,17 +50,10 @@ boost::shared_ptr<crocoddyl::IntegratedActionModelAbstract> IntegratorFactory::c
       action = boost::make_shared<crocoddyl::IntegratedActionModelEuler>(model);
       break;
     case IntegratorTypes::IntegratorRK2:
-<<<<<<< HEAD
```
cmastalli: It seems you have rebase or something.
Please fix this issue.

andreadelprete: I have already fixed it, haven't I?

cmastalli: Yes, you're right.
Sorry for the noise

proyan:
@andreadelprete @cmastalli Thanks for this nice feature. I'm okay with this PR. The only modification to the API that I suggest is to templatize IAMRK with enum rktype. As a result:
1) The deprecation of IAMRK4 can be removed, and the header `rk4.hpp` would contain a simple typedef.
2) The python API would need to be updated as well. Since the rktype enum is only of size 3, I would suggest exposing IAMRK2/3/4 instead of taking the enum type as an argument. This would keep the API consistent with other IAM models.

cmastalli:
> The only modification to the API that I suggest is to templatize IAMRK with enum rktype

Thank you for the suggestion ;)
I have indeed though about it, but didn't comment in this thread.
After rethinking your previous suggestion I am still incline to have a *runtime resolution* as:
  - the *compile resolution* (i.e., your suggestion) seems not to improve importantly the computation speed and also increases compilation time, and
  - the runtime resolution can be extended to dynamically change the integrator without duplicating objects.

In short, I would vote to do not change it. Perhaps, others might express their thoughts. Otherwise, I would merge it.

PR #1010, opened 2021-11-09
Multiply operators to codegen and to exploit Fx, Fu structure in Riccati sweep

cmastalli:
This PR aims to develop a set of functions in which the user might describe the dynamics structure.
It also follows the spirit designed for the control-parametrization operators.

As described in the PR title, this brings to benefits:
  1. Exploiting the structure of Fx, Fu matrices
  2. Code-generation of the Riccati sweep

Note that this is a first PR on the topic as there is still (some) remaining work:
  - Define Fx operators in Euler integrator
  - Define Fx,Fu operators in RK integrator

which I will tackle in future PRs.


include/crocoddyl/core/diff-action-base.hxx
```diff
@@ -89,11 +89,189 @@ DifferentialActionModelAbstractTpl<Scalar>::createData() {
...
+      break;
+    case addto:
+      out.noalias() += FuTranspose * A;
+      break;
+    case rmfrom:
+      out.noalias() -= FuTranspose * A;
+      break;
+    default:
+      throw_pretty("Invalid argument: allowed operators: setto, addto, rmfrom");
+  }
```
jcarpent, line 231: Why this set of lines of code is duplicated everywhere?

cmastalli: There are two kind of operators:
  1. A * Jacobian
  2. Jacobian transpose * A

where the Jacobians could be `Fx` or `Fu`.
The latter uses a row-major storage, instead the former uses a column-major one.
As far I understand, we can try to condense both functions into a single one, but then, we'll affect the performance.
What are your thoughts/feedback about it?
(Note that the control-parametrization class follows the same pattern)

Additionally, these operators are defined in two classes:
  1. action model
  2. diff-action model

This is needed as we want to exploit the structure of both: integrators and differential action models.

Finally, this set of lines (as the one in the action-model class) describe a default implementation, which it is a `dense-multiplication` operator.
In short, we also need them.

cmastalli: @jcarpent  -- is there a follow-up point to discuss?

cmastalli:
I'll let this work for Crocoddyl v2.
@jcarpent if you have further comments, then feel free to add them.

PR #1011, opened 2021-11-10
[Update] Manage 2d contact friction in xz plane

VasilyRakche:
In reference of #986 problem, and as an update to #1000 I have found some time to add changes in order to accommodate previous mistakes and requests. It was decided to use new branch for cleaner history (I couldn't merge newest changes from develop without an additional merge commit).  I am using the newest Eigen 3.4 release for matrix multiplication. 

Notes to have in mind: 

1. This solution still only works for xz plane for now, I have tested it with my implementation of a test-stand, it works ok
2. In `unittest/factory/diff_action.cpp` I have not added force regularization when testing with friction (force regularization is not implemented yet for 2D contact) 
3. In `include/crocoddyl/multibody/residuals/contact-friction-cone.hpp` I dont have a cast for ImpulseData for 2D contact since that would require additional implementation

### DIFF
--- a/include/crocoddyl/multibody/residuals/contact-friction-cone.hpp
+++ b/include/crocoddyl/multibody/residuals/contact-friction-cone.hpp
@@ -234,7 +234,7 @@ struct ResidualDataContactFrictionConeTpl : public ResidualDataAbstractTpl<_Scal
             contact = it->second;
             break;
           }
-          throw_pretty("Domain error: there isn't defined at least a 3d contact for " + frame_name);
+          throw_pretty("Domain error: there isn't defined at least a 2d contact for " + frame_name);
           break;
         }
       }
--- a/include/crocoddyl/multibody/residuals/contact-friction-cone.hpp
+++ b/include/crocoddyl/multibody/residuals/contact-friction-cone.hpp
@@ -15,6 +15,7 @@
 #include "crocoddyl/multibody/contact-base.hpp"
 #include "crocoddyl/multibody/impulse-base.hpp"
 #include "crocoddyl/multibody/contacts/multiple-contacts.hpp"
+#include "crocoddyl/multibody/contacts/contact-2d.hpp"
 #include "crocoddyl/multibody/contacts/contact-3d.hpp"
 #include "crocoddyl/multibody/contacts/contact-6d.hpp"
 #include "crocoddyl/multibody/impulses/multiple-impulses.hpp"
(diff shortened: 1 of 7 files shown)


VasilyRakche:
Hey, have no idea why the doc test is failing, how can I get some help?

nim65s:
Probably because your code is not compatible with older Eigen versions, eg.:
```
…/crocoddyl-1.8.1/include/crocoddyl/multibody/residuals/contact-friction-cone.hxx:57:37: error: 'all' is not a member of 'Eigen'
57 |       data->Rx.noalias() = A(Eigen::all, {0, 2}) * df_dx;
|                                     ^~~
```

VasilyRakche:
Ohh, yes, @nim65s  I see now. The thing is that it provides quite elegant solution for removing intermediate matrix as requested previously:

> 
> ```
>    MatrixXs A_2d(A.rows(), 2);
>     A_2d << A.col(0), A.col(2)
> ```
> This allocates memory. Please remove the intermediate matrix, and make the corresponding multiplications for computing `Rx`, and `Ru`.
> _Originally posted by @cmastalli in https://github.com/loco-3d/crocoddyl/pull/1000#discussion_r712185450_

I couldn't think of any better solution - other one would be going with brute force matrix multiplication. 

When do you plan transfering to Eigen3.4 - Its released in mid august and introducing some nice features.

jcarpent:
> When do you plan transfering to Eigen3.4 - Its released in mid august and introducing some nice features.

As this software should operate on many platforms and many various OS, we cannot stick to a precise Eigen version. 
There is no change to move toward Eigen 3.4 now.

VasilyRakche:
@jcarpent  Got it, do you guys maybe have suggestion how to do this multiplication efficiently with older version of Eigen. 

So, in the implementation I got:
```
      data->Rx.noalias() = A(Eigen::all, {0, 2}) * df_dx;
```
A is 5x3 matrix initially and df_dx is 2x6 (right dimension for 2D Contact). I want to take only collumns with index 0 and 2 from matrix A which would correspond to x and z plane, and multiply it by df_dx (previously mentioned code should be good reference for what I want). But the question is how to achieve it with older Eigen.

jcarpent:
As simple as: `data->Rx.noalias() = A.col(0) * df_dx.row(0) + A.col(2) * df_dx.row(1)`

jcarpent:
I've just unrolled the operations.

VasilyRakche:
You are completely right! As simple as that. Thanks a lot!

cmastalli (approved):
Thank you for the patience and support on this topic @VasilyRakche 
Thank @nim65s and @jcarpent  for providing useful feedback in this PR.

Issue #934, opened 2021-04-23
Moving to a more recent C++ standard


wxmerkt:
With 16.04 reaching EOL next month and C++11 about ten years old, I wonder if there are requirements to continue to support C++98 in the community? In practice we _implicitly_ already require C++11 support (e.g. for URDF parsing in Pinocchio which translates to Crocoddyl) [1] and officially moving the minimum supported standard to a newer version could simplify development by benefitting from new features (e.g. using `chrono` in #931). I'd love to hear thoughts and plans on this topic for future developments :-).

[1] Our buildserver on Xenial also appears to require `CMAKE_CXX_STANDARD=11` for more than a year now (https://github.com/loco-3d/crocoddyl/commit/34bdda2e4d79c3295a1cba7dbc46d3d62bc4f9fa). I can't see the gepgitlab config so Guilhem would have to comment on that.

cmastalli:
I am in with moving to c++11

nim65s:
I think we could even move directly to C++14. 

I've seen many cases where developers think they are using a C++11 feature, while it is in fact a C++14 feature (eg. using `auto` as a function return type), and it works on their computer because the default standard is >= 14, but raises boring issues when they want to share their work.

I think the oldest system we support is 16.04, and activating C++11 or C++14 on 16.04 is the same thing.


To answer about the implicit requirement, I don't think this is the case. On gitlab, pinocchio is using urdfdom v0.4 provided by ubuntu, which doesn't require C++11.

proyan:
I've already stated my preference for updating our base support to c++11. If we can move to c++14, that is even better from my perspective. Crocoddyl is supposed to be on the working on the robot, and I don't think any current robots still operate on 12.04.
And Ubuntu 14.04 has c++14 support.  (https://www.gnu.org/software/gcc/projects/cxx-status.html#cxx14)
So I am okay with moving directly to c++14.

cmastalli:
I think we can close this issue (see #949 for more details)

PR #1021, opened 2022-01-05
Reduce temporary allocations

wxmerkt:
- CostModelResidual: Workaround for heap allocation in the product with the diagonal weighting matrix of the activation. This is a workaround for https://gitlab.com/libeigen/eigen/-/issues/2408
- LQR: Introduces a temporary variable for the Hessian-vector product to eliminate temporaries in the computation of the cost

### DIFF
--- a/include/crocoddyl/core/actions/lqr.hpp
+++ b/include/crocoddyl/core/actions/lqr.hpp
@@ -87,17 +87,19 @@ struct ActionDataLQRTpl : public ActionDataAbstractTpl<_Scalar> {
   typedef _Scalar Scalar;
   typedef MathBaseTpl<Scalar> MathBase;
   typedef ActionDataAbstractTpl<Scalar> Base;
+  typedef typename MathBase::VectorXs VectorXs;
 
   template <template <typename Scalar> class Model>
-  explicit ActionDataLQRTpl(Model<Scalar>* const model) : Base(model) {
+  explicit ActionDataLQRTpl(Model<Scalar>* const model)
+      : Base(model),
+        Luu_u_tmp(VectorXs::Zero(static_cast<Eigen::Index>(model->get_nu()))),
+        Lxx_x_tmp(VectorXs::Zero(static_cast<Eigen::Index>(model->get_state()->get_ndx()))) {
     // Setting the linear model and quadratic cost here because they are constant
     Fx = model->get_Fx();
     Fu = model->get_Fu();
     Lxx = model->get_Lxx();
     Luu = model->get_Luu();
     Lxu = model->get_Lxu();
-    Hu_tmp.resize(static_cast<Eigen::Index>(model->get_nu()));
-    Hx_tmp.resize(static_cast<Eigen::Index>(model->get_state()->get_ndx()));
   }
 
   using Base::cost;
@@ -110,8 +112,8 @@ struct ActionDataLQRTpl : public ActionDataAbstractTpl<_Scalar> {
   using Base::Lxx;
   using Base::r;
   using Base::xnext;
-  typename MathBase::VectorXs Hu_tmp;  // Temporary variable for storing Hessian-vector product (size: nu)
-  typename MathBase::VectorXs Hx_tmp;  // Temporary variable for storing Hessian-vector product (size: nx)
+  VectorXs Luu_u_tmp;  // Temporary variable for storing Hessian-vector product (size: nu)
+  VectorXs Lxx_x_tmp;  // Temporary variable for storing Hessian-vector product (size: nx)
 };
 
 }  // namespace crocoddyl
--- a/include/crocoddyl/core/actions/lqr.hpp
+++ b/include/crocoddyl/core/actions/lqr.hpp
@@ -96,6 +96,8 @@ struct ActionDataLQRTpl : public ActionDataAbstractTpl<_Scalar> {
     Lxx = model->get_Lxx();
     Luu = model->get_Luu();
     Lxu = model->get_Lxu();
+    Hu_tmp.resize(static_cast<Eigen::Index>(model->get_nu()));
+    Hx_tmp.resize(static_cast<Eigen::Index>(model->get_state()->get_ndx()));
   }
 
   using Base::cost;
@@ -108,6 +110,8 @@ struct ActionDataLQRTpl : public ActionDataAbstractTpl<_Scalar> {
   using Base::Lxx;
   using Base::r;
   using Base::xnext;
+  typename MathBase::VectorXs Hu_tmp;  // Temporary variable for storing Hessian-vector product (size: nu)
+  typename MathBase::VectorXs Hx_tmp;  // Temporary variable for storing Hessian-vector product (size: nx)
 };
 
 }  // namespace crocoddyl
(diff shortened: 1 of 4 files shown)


include/crocoddyl/core/actions/lqr.hxx
```diff
@@ -58,8 +78,12 @@ void ActionModelLQRTpl<Scalar>::calc(const boost::shared_ptr<ActionDataAbstract>
     throw_pretty("Invalid argument: "
                  << "x has wrong dimension (it should be " + std::to_string(state_->get_nx()) + ")");
   }
+  boost::shared_ptr<ActionDataLQRTpl<Scalar>> d = boost::static_pointer_cast<ActionDataLQRTpl<Scalar>>(data);
```
cmastalli, line 81: Same here as comment below

include/crocoddyl/core/actions/lqr.hpp
```diff
@@ -96,6 +96,8 @@ struct ActionDataLQRTpl : public ActionDataAbstractTpl<_Scalar> {
     Lxx = model->get_Lxx();
```
cmastalli, line 96: Please construct the size of the matrix and use `setZero`, e.g,
```c++
  explicit ActionDataLQRTpl(Model<Scalar>* const model) : Base(model), Luu_u(model->get_nu()) {
  Luu_u.setZero();
}
```

Note also that I have renamed the variable to `Luu_u`. Do the same for `Hx_tmp` (i.e., `Lxu_u`).

Further comment:
This notation is not allowed in Crocoddyl 2 (to be release asap). Those names are reserved to store constraint derivatives.

include/crocoddyl/core/actions/lqr.hxx
```diff
@@ -38,6 +38,7 @@ void ActionModelLQRTpl<Scalar>::calc(const boost::shared_ptr<ActionDataAbstract>
     throw_pretty("Invalid argument: "
                  << "u has wrong dimension (it should be " + std::to_string(nu_) + ")");
   }
+  boost::shared_ptr<ActionDataLQRTpl<Scalar>> d = boost::static_pointer_cast<ActionDataLQRTpl<Scalar>>(data);
```
cmastalli, line 41: We don't accept this code format. Similar to other comments.

include/crocoddyl/core/actions/lqr.hxx
```diff
@@ -47,8 +48,27 @@ void ActionModelLQRTpl<Scalar>::calc(const boost::shared_ptr<ActionDataAbstract>
...
+
+  // x^T*Lxu*u
+  d->Hx_tmp.noalias() = Lxu_ * u;
+  data->cost += x.transpose() * d->Hx_tmp;
+
+  // lx^T*x
+  data->cost += lx_.transpose() * x;
+
+  // lu^T*u
+  data->cost += lu_.transpose() * u;
```
cmastalli, line 71:
```suggestion
d->Lxx_x.noalias() = Lxx_ * x;
d->Luu_x.noalias() = Luu_ * u;
data->cost = Scalar(0.5) * x.dot(d->Lx_x);
data->cost += Scalar(0.5) * x.dot(d->Lx_u);
d->Lxx_x.noalias() = Lxu_ * u;
data->cost += Scalar(0.5) * x.dot(d->Lx_x);
data->cost += lx_.dot(x);
data->cost += lu_.dot(u);
```

Additionally, I think we could add some `dot` in the same line. I think this should not allocate memory as the results is a scalar. Am I right?

include/crocoddyl/core/costs/residual.hxx
```diff
@@ -60,28 +60,28 @@ void CostModelResidualTpl<Scalar>::calcDiff(const boost::shared_ptr<CostDataAbst
   const std::size_t nv = state_->get_nv();
   if (is_ru) {
     data->Lu.noalias() = data->residual->Ru.transpose() * data->activation->Ar;
-    d->Arr_Ru.noalias() = data->activation->Arr * data->residual->Ru;
```
cmastalli, line 63: Could you provide information about improve performance?
I ask this as the results reported in private where not better (in terms of computation time).

cmastalli (changes requested):

include/crocoddyl/core/actions/lqr.hxx
```diff
@@ -101,7 +125,7 @@ void ActionModelLQRTpl<Scalar>::calcDiff(const boost::shared_ptr<ActionDataAbstr
 }
 
 template <typename Scalar>
-boost::shared_ptr<ActionDataAbstractTpl<Scalar> > ActionModelLQRTpl<Scalar>::createData() {
```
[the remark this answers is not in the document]
wxmerkt: I did run it (and discussed with Guilhem) -- this is what the formatter caused :-(

include/crocoddyl/core/actions/lqr.hpp
wxmerkt: Okay, will do - in this case we will have to add an additional variable resulting in a slightly larger struct (Lxx_x, Lxu_u, Luu_u)

include/crocoddyl/core/actions/lqr.hxx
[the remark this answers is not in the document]
wxmerkt: I.e. the formatter does not consider this to be incorrect / does not fix it

include/crocoddyl/core/costs/residual.hxx
cmastalli: Ok, thanks for the info.

include/crocoddyl/core/actions/lqr.hpp
cmastalli: Lxx_x` and `Lxu_u` have the same dimension, you could reuse it.

Actually, it might be better to rename them as `Lxx_x_tmp` and `Luu_u_tmp`, as `Lxx_x` is not necessary `Lxx * x`.

include/crocoddyl/core/actions/lqr.hxx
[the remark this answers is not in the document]
cmastalli: @wxmerkt 
Did you use `clang-6`?
This code is not c++98 compatible; thus, it is weird that this is the case.

@nim65s 
Why is now different than before? 
Are we using other setup in the CI?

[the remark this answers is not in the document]
wxmerkt: From what I recall we moved to C++14 already since we implicitly had that requirement anyways (https://github.com/loco-3d/crocoddyl/issues/934), so it shouldn't matter if it's not C++98 compatible:

https://github.com/loco-3d/crocoddyl/blob/9e79f80c6d52b2c6c000106570811d449341cad3/CMakeLists.txt#L63

include/crocoddyl/core/actions/lqr.hxx
wxmerkt: Removed comments and amended with dot

include/crocoddyl/core/actions/lqr.hxx
cmastalli, line 41: I didn't realize this before.

We should cast a raw pointer, instead. This avoids to create an owner.
It should be something like this:
```c++
 ​  Data* d = ​static_cast​<Data*>(data.​get​());
```

cmastalli (commented):
Thanks @wxmerkt! I just left a tiny extra comment.

cmastalli (approved):

wxmerkt:
Thank you for handling the open comments. Let's merge this one after #1025 is merged

cmastalli:
I am afraid that the Python bindings in #1025 might take you considerably time.
Perhaps, you could cherry pick the commits needed to allow the CI passes.
Thus, we don't need to wait for it.

cmastalli:
@wxmerkt  -- thank you again for these fixes!

Issue #1022, opened 2022-01-05
(Semantic) re-use of final element of dx_


wxmerkt:
- `fs_` represents the difference between the roll-out and state trajectory, i.e., the dynamic defects/gaps and is of size T+1 (same as the trajectory - `xs_`, `xs_try_`)
- `dx_` represents the error during the forward-pass/roll-out. It is computed for each node where a control exists. Consequentially, it has size T

This is fine for DDP-based solvers. In FDDP-derived solvers, we also compute the difference at the terminal model from `dx_` to deflect `Vxx_`:
https://github.com/loco-3d/crocoddyl/blob/227e37886f521832f454e9263905ce1443822474/src/core/solvers/fddp.cpp#L122-L123

In this case, by using `dx_.back()` we are _semantically_ reusing the final element (index `T-1`) twice: First to compute the final node's dx_ (T+1-th element, but accessing T-th element `dx_.back()`), then for the trajectory (but the T-th component, `dx_[t=T-1]`).

This has no impact on the numerical performance since the T+1-th element is only used in the line above and not anywhere else in the code, but should we either (1) add a comment to explain that we are re-using this element in the line back `dx_.back()` [i.e. `back()` for `dx_` isn't the same as for `xs_` and `xs_try_`], or (2) increase the vector `dx_` to `T+1`. (2) is only relevant for FDDP-derived solvers [I've had an old branch where I had this dimension change fix but I don't think it's relevant to carry it forward], so I am inclined to go with (1). Thoughts?

cmastalli:
I also think it is better to go for 1.
You could add a comment in line 123!

PR #1023, opened 2022-01-05
Update doc for dx_ size

wxmerkt:
Possible solution to https://github.com/loco-3d/crocoddyl/issues/1022


cmastalli (approved):
Thanks @wxmerkt

PR #1015, opened 2021-12-07
Add ContactModel1D

skleff1994:
Implementation of a contact model 1D that constrains only the translation along z-axis . It allows to apply a normal force while moving tangentially, which is useful for e.g. scraping or sanding tasks, but not allowed by current contact models (2D, 3D, 6D). 

More specifically : 
- implementation of ContactModel1D 
- completion the contacts unittests factory 
- implemention of python binding
- completion of ResidualModelContactForce to allow normal force cost

I'm not sure if things are missing, please let me know !

PR #1025, opened 2022-01-05
ContactModel1D: Fix compilation with LLVM, fix unittests with clang

wxmerkt:
Fixes compilation issue introduced in #1015

Following #1024, we should also see CI status going green for clang on GitHub Actions


cmastalli (approved):
@wxmerkt thank you so much for the fix and CI integration.

Btw, I have a question regarding the CI that you developed. Are we testing multi-thread support?

wxmerkt:
This is ready for merging now. I managed to fix the unittest locally but it still fails on the buildserver. After spending a day on it, I deactivated the failing test for Clang to make the buildserver pass. I managed to track it down to a double corruption which appears to come from boost's unittest framework. Let's leave #964 open and resolve the issue at a later date.

unittest/test_diff_actions.cpp
```diff
@@ -20,29 +20,29 @@ using namespace crocoddyl::unittest;
 void test_check_data(DifferentialActionModelTypes::Type action_type) {
   // create the model
   DifferentialActionModelFactory factory;
-  const boost::shared_ptr<crocoddyl::DifferentialActionModelAbstract>& model = factory.create(action_type);
+  boost::shared_ptr<crocoddyl::DifferentialActionModelAbstract> model = factory.create(action_type);
```
cmastalli, line 23: Is this triggered issues in clang?

I wonder why do we need to this?

unittest/test_diff_actions.cpp
```diff
@@ -20,29 +20,29 @@ using namespace crocoddyl::unittest;
...
 
   // create the corresponding data object
-  const boost::shared_ptr<crocoddyl::DifferentialActionDataAbstract>& data = model->createData();
+  boost::shared_ptr<crocoddyl::DifferentialActionDataAbstract> data = model->createData();
 
   // Generating random state and control vectors
-  const Eigen::VectorXd& x = model->get_state()->rand();
-  const Eigen::VectorXd& u = Eigen::VectorXd::Random(model->get_nu());
+  const Eigen::VectorXd x = model->get_state()->rand();
```
cmastalli, line 44: same question here?

unittest/CMakeLists.txt
```diff
@@ -44,6 +44,13 @@ SET(${PROJECT_NAME}_CPP_TESTS
   test_solvers
 )
 
+# test_diff_actions fails with Clang, cf. https://github.com/loco-3d/crocoddyl/issues/964
+# It's difficult to identify the source - it appears to be a double free from inside boost's unittest framework.
+# Hence, we will remove it for now, it's still tested with g++:
+IF(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
cmastalli, line 50: If this works locally and not in the buildserver, wouldn't be better to have a flag for the buildserver (which is disable by default)?

This allows developers to run this test in their PCs.

cmastalli (changes requested):
Thanks for this effort @wxmerkt 

There are two main missing pieces in this PR:
 1. Python bindings of `get_active_set` and `get_inactive_set`.
 2. Deprecation of Python bindings of `get_active` and `set_inactive`.

It also seems to me that we need to disable the printing of warning messages (due to deprecation of functions in 2) during the Python bindings compilation.

Additionally, I have left two other comments. Please note the comment in CMakeList as it is especially important.

unittest/test_diff_actions.cpp
wxmerkt: I changed it during debugging. Note that the factory returns the shared_ptr and thus whoever calls `create` should own it. But we would only store a const-ref to the pointer. What we could do is return a unique_ptr from the factory thus signifying that whoever calls it needs to take ownership (and we can instantiate a shared from a unique_ptr)

unittest/CMakeLists.txt
wxmerkt: I can reproduce it when running from a Docker in the same way the buildserver is, but not on my local installation. Locally I am using a slightly different version of Pinocchio and other dependencies, so it might be that. If you confirm that commenting those lines it works on your setup, happy to make it conditional for the buildserver - otherwise this is the easiest and will be tested on the buildserver nonetheless with g++

wxmerkt:
> Python bindings of get_active_set and get_inactive_set.
> Deprecation of Python bindings of get_active and set_inactive.

`get_active` and `set_active` are not directly exposed, but as a property of the class. As thus my plan was to not directly expose `get_active_set` and `set_active_set` either but at the next version increase switch which methods are bound to the property. Otherwise we'd end up with an inconsistent API where some things are a property and others are getters/setters. What are your thoughts?

> It also seems to me that we need to disable the printing of warning messages (due to deprecation of functions in 2) during the Python bindings compilation.

Indeed, I already had this as part of another PR I am preparing (and which triggered the other prepatory PRs). I can cherry-pick that commit to this branch to get it in sooner.

unittest/test_diff_actions.cpp
cmastalli: It is not a good idea to use  an `unique_ptr` as we (something) need to use these pointers for constructing other Crocoddyl objects.

The proposed code is creating a second owner.
However, the current approach avoids it by passing a reference.
Note that multiple owners add extra (but tiny) runtime overhead.

My question is **does boost unit-test need this?**

bindings/python/crocoddyl/CMakeLists.txt
```diff
@@ -19,6 +19,8 @@ TARGET_LINK_BOOST_PYTHON(${PROJECT_NAME}_pywrap)
 # BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS spews conversion warnings from int to long unsigned int.
 # Unfortunately, using literals does not work in a macro. As such, this turns them off for the entire wrapper:
 TARGET_COMPILE_OPTIONS(${PROJECT_NAME}_pywrap PRIVATE "-Wno-conversion")
+# We are also exposing deprecated declarations until they are removed. Ignore the errors for the wrapper:
+TARGET_COMPILE_OPTIONS(${PROJECT_NAME}_pywrap PRIVATE "-Wno-deprecated-declarations")
```
cmastalli, line 23: I think it is better to add this flag in each file.
Otherwise, it is harder to detect if we haven't effectively update bits in the Python bindings
Do you agree with me?

cmastalli:
> `get_active` and `set_active` are not directly exposed, but as a property of the class. As thus my plan was to not directly expose `get_active_set` and `set_active_set` either but at the next version increase switch which methods are bound to the property. Otherwise we'd end up with an inconsistent API where some things are a property and others are getters/setters. What are your thoughts?

What do you mean with "methods are bound"? (I don't understand this comment).
Furthermore, we could call this attribute as `active_set` in Python as well.
Therefore we deprecate `active` one.

bindings/python/crocoddyl/CMakeLists.txt
wxmerkt: It's quite a lot of them since we already have a lot of warnings from the Residuals / Frame costs that also need to be silenced

unittest/test_diff_actions.cpp
wxmerkt: We can have methods that create objects return unique_ptr - if we need shared_ptr, the one instantiating can create a shared_ptr from a unique_ptr:
```cpp
std::unique_ptr<Something> createSomething() { return std::make_unique<Something>(); }

std::shared_ptr<Something> something = createSomething();
```

Cf. https://godbolt.org/z/xvf41ah6e

> The proposed code is creating a second owner.

I am not sure it is given return-value optimisation. The `create` method in this case does not keep a shared_ptr itself.

The change was part of the debugging to resolve the Clang unittest crashes.

@jcarpent Any thoughts?

unittest/test_diff_actions.cpp
wxmerkt: I don't think that rand() is returning a const-ref to something that is stored / a member variable. We may run the risk of reference to a temporary. What are your thoughts @jcarpent ?

bindings/python/crocoddyl/CMakeLists.txt
wxmerkt: I've looked into it and the deprecated Frame declarations are not silenced by the pragmas in the respective files. Due to this, I'll leave the general deactivation in for now until we remove the old Frame signatures. It's too noisy otherwise. For the newly deprecated code, I added it in the file properly as well though.

wxmerkt:
I've created a `set-converter` and exposed `StdSet_String` as well as the new signatures as `active_set`/`inactive_set`. The `StdSet` data structures are picklable. I did not get the visitor for operations on the exposed data type to work, so this will have to be future work.

The PR is ready for merge

bindings/python/crocoddyl/CMakeLists.txt
cmastalli: You're referring to clang compilation, right?

I am aware that current code is working on gcc only.

cmastalli (approved):

cmastalli:
Thanks @wxmerkt for this effort ;)

Issue #1027, opened 2022-01-06
Enhance Continuous Integration & Testing


wxmerkt:
- [ ] Add configuration for multithreading (e.g. clang multithreading fails unittests, cf. #1026)
- [ ] Build and test in Debug and Release mode
- [ ] Add configuration for vectorisation
- [ ] Add dashboard to track benchmark performance over PRs/commits to devel

cmastalli:
Testing vectorization will require to compile all the dependencies (pinocchio, hpp-fcl, eigenpy, etc) with the same flags, if we want to run Python-based tests and examples.
How are you planning to do this?

wxmerkt:
Good point. I don't have the bandwidth to implement these items but wanted to keep track of them. They are a good starting point if someone is keen to pick them up or if we'd have some engineer bandwidth

cmastalli:
> Good point. I don't have the bandwidth to implement these items but wanted to keep track of them. They are a good starting point if someone is keen to pick them up or if we'd have some engineer bandwidth

We could definitely start with testing compilation only (if this is not enough).

wxmerkt:
Pinocchio on ROS-CI now has HPP-FCL included, so the collision tests are run as well

cmastalli:
I added a configuration for codegen using CONDA; however, it would be nice to have another one in gitlab-ci.

Issue #1026, opened 2022-01-06
Unittests fail for clang + multi-threading


wxmerkt:
Related to #964 but distinct since it requires different fixes. More unittests fail with clang12 + multi-threading enabled

cmastalli:
Does https://github.com/loco-3d/crocoddyl/pull/1053 solve the reported issue with clang-12 and multi-threading?

Could you try it?

PR #1028, opened 2022-01-06
Increase CI coverage: multi-threading, format check

wxmerkt:
On top of #1025. Will rebase once that PR is merged.

Multi-threading is right now only configured for Clang (which works on GitHub Actions, but fails locally :man_shrugging:). We can add/change it to g++ (which works locally). Format checking (C++ only for now) is redundant with gepgitlab, but since it runs reasonably fast it can't hurt to have.

Relates to #1026, #1027


cmastalli (approved):
The CI coverage is OK.
Of course, there are other details to fix which we are discussing in #1025.

cmastalli:
@wxmerkt -- could you rebase this branch?

cmastalli (approved):

PR #1036, opened 2022-01-24
Fix multi-threading CMake setting + more detailed error message for SolverAbstract::setCandidate

wxmerkt:


src/core/solver-base.cpp
```diff
@@ -146,8 +146,8 @@ void SolverAbstract::setCandidate(const std::vector<Eigen::VectorXd>& xs_warm,
       const std::size_t nu = model->get_nu();
       if (static_cast<std::size_t>(us_warm[t].size()) != nu) {
         throw_pretty("Invalid argument: "
-                     << "us_init[" + std::to_string(t) + "] has wrong dimension (it should be equals to " +
-                            std::to_string(nu) + ")");
+                     << "us_init[" + std::to_string(t) + "] has wrong dimension (" << us_warm[t].size()
```
cmastalli, line 149: Could you apply the same message for xs_init (line 128)?
Also for the `set_xs` and `set_us` functions?

wxmerkt: Done :-)

cmastalli (approved):

src/core/solver-base.cpp
cmastalli: thx!

PR #1037, opened 2022-01-27
Make example-robot-data an optional dependency

wxmerkt:
Per Element chat - example-robot-data is required for unittests, benchmarking, and examples. If all three options are OFF, we don't need to add this required dependency


cmastalli (approved):
Thanks @wxmerkt ;)

PR #1043, opened 2022-02-03
Disable multithreading when we use models derived from Python code

cmastalli:


wxmerkt (approved):
Woohoo, thanks for the super swift resolution and fix!

PR #1048, opened 2022-02-22
ActionModelNumDiff: Forward control bounds, fixes #1047

### LINKED ISSUE #1047: Forwarding of control limits for ActionModelNumDiff
If we have an Action Model with control limits and wrap it in an `ActionModelNumDiff`, the control limits of the num-diff action model are not forwarded and left at default values / infinity (since they use the default base's values).

How do we want to handle this?

My proposal is to  set the action limits of the num-diff model from the base model in the constructor. Note, however, that this won't update in either direction if they are changed later on (although the latter case is the user's error).

wxmerkt:


cmastalli (approved):
Thanks @wxmerkt ;)

Issue #1050, opened 2022-03-01
Wrong comments about APIs


JaehyunShim:
Wrong comments about APIs in the following file (https://github.com/loco-3d/crocoddyl/blob/devel/include/crocoddyl/multibody/residuals/frame-velocity.hpp#L103)

The comments for get and set functions should be opposite.

cmastalli:
Could you fix this comment and create a PR?
Thx ;)

PR #1051, opened 2022-03-01
Fix wrong comments about APIs in frame-velocity.hpp

JaehyunShim:
@cmastalli  

Ref: (https://github.com/loco-3d/crocoddyl/issues/1050)


cmastalli (approved):
Thanks @JaehyunShim ;)

PR #1053, opened 2022-03-18
Tests: remove link to boost unit test framework

cmastalli:
This PR provides a similar solution proposed in https://github.com/humanoid-path-planner/hpp-fcl/pull/270.
Hopefully, this PR will fix the segfault that appears in Debian systems.


cmastalli:
@nim65s -- do you know why this job (https://gitlab.laas.fr/loco-3d/crocoddyl/-/jobs/141664) is broken?

cmastalli:
Also the buster job is now passing (as expected).
I suggest to do not allow failures anymore.

PR #1054, opened 2022-03-22
Allowing WORLD and LOCAL_WORLD_ALIGNED in ContactModel1D

skleff1994:
Minor enhancements for the 1D contact model . Related to #812 

- Add a mask for constraint dimension selection . The user selects optionally `type` in `{0,1,2}` for `{x,y,z}` the z-axis being the default one
- Add optional argument of pinocchio reference frame in {`LOCAL`, `WORLD` , `LOCAL_WORLD_ALIGNED`}, `LOCAL` being the default one

The mask could be extended to 2D contact , and the pinocchio reference frame option could be propagated to other contact models , if deemed useful . But I think all of this will be handled in Pinocchio in the future...

### DIFF
--- a/bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
+++ b/bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
@@ -15,6 +15,12 @@ namespace python {
 void exposeContact1D() {
   bp::register_ptr_to_python<boost::shared_ptr<ContactModel1D> >();
 
+  bp::enum_<pinocchio::ReferenceFrame>("ReferenceFrame")
+      .value("LOCAL", pinocchio::LOCAL)
+      .value("WORLD", pinocchio::WORLD)
+      .value("LOCAL_WORLD_ALIGNED", pinocchio::LOCAL_WORLD_ALIGNED)
+      .export_values();
+
   bp::class_<ContactModel1D, bp::bases<ContactModelAbstract> >(
       "ContactModel1D",
       "Rigid 1D contact model.\n\n"
@@ -24,21 +30,23 @@ void exposeContact1D() {
       "The calc and calcDiff functions compute the contact Jacobian and drift (holonomic constraint) or\n"
       "the derivatives of the holonomic constraint, respectively.",
       bp::init<boost::shared_ptr<StateMultibody>, pinocchio::FrameIndex, double, std::size_t,
-               bp::optional<Eigen::Vector2d> >(
-          bp::args("self", "state", "id", "xref", "nu", "gains"),
+               bp::optional<Eigen::Vector2d, std::size_t, pinocchio::ReferenceFrame> >(
+          bp::args("self", "state", "id", "xref", "nu", "gains", "type", "pinRefFrame"),
           "Initialize the contact model.\n\n"
           ":param state: state of the multibody system\n"
           ":param id: reference frame id of the contact\n"
           ":param xref: contact position used for the Baumgarte stabilization\n"
           ":param nu: dimension of control vector\n"
-          ":param gains: gains of the contact model (default np.matrix([0.,0.]))"))
-      .def(bp::init<boost::shared_ptr<StateMultibody>, pinocchio::FrameIndex, double, bp::optional<Eigen::Vector2d> >(
+          ":param gains: gains of the contact model (default np.matrix([0.,0.]))\n"
+          ":param type: axis of the contact constraint (0=x, 1=y or 2=z)\n"
+          ":param pinRefFrame: pin.ReferenceFrame in {LOCAL, WORLD, LOCAL_WORLD_ALIGNED}"))
+      .def(bp::init<boost::shared_ptr<StateMultibody>, pinocchio::FrameIndex, double, bp::optional<Eigen::Vector2d, pinocchio::ReferenceFrame> >(
           bp::args("self", "state", "id", "xref", "gains"),
           "Initialize the contact model.\n\n"
           ":param state: state of the multibody system\n"
           ":param id: reference frame id of the contact\n"
           ":param xref: contact position used for the Baumgarte stabilization\n"
-          ":param gains: gains of the contact model (default np.matrix([0.,0.]))"))
+          ":param gains: gains of the contact model (default np.matrix([0.,0.]))\n"))
       .def("calc", &ContactModel1D::calc, bp::args("self", "data", "x"),
            "Compute the 1D contact Jacobian and drift.\n\n"
            "The rigid contact model throught acceleration-base holonomic constraint\n"
@@ -68,7 +76,10 @@ void exposeContact1D() {
                     &ContactModel1D::set_reference, "reference contact translation")
       .add_property("gains",
                     bp::make_function(&ContactModel1D::get_gains, bp::return_value_policy<bp::return_by_value>()),
-                    "contact gains");
+                    "contact gains")
+      .add_property("pinReferenceFrame",
+                    bp::make_function(&ContactModel1D::get_pinReferenceFrame, bp::return_value_policy<bp::return_by_value>()),
+                    &ContactModel1D::set_pinReferenceFrame, "pin.ReferenceFrame");
 
   bp::register_ptr_to_python<boost::shared_ptr<ContactData1D> >();
 
--- a/bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
+++ b/bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
@@ -15,10 +15,10 @@ namespace python {
 void exposeContact1D() {
   bp::register_ptr_to_python<boost::shared_ptr<ContactModel1D> >();
 
-  bp::enum_<Contact1DMaskType>("Contact1DMaskType")
-      .value("X_MASK", X_MASK)
-      .value("Y_MASK", Y_MASK)
-      .value("Z_MASK", Z_MASK)
+  bp::enum_<Vector3MaskType>("Vector3MaskType")
+      .value("x", x)
+      .value("y", y)
+      .value("z", z)
       .export_values();
 
   bp::class_<ContactModel1D, bp::bases<ContactModelAbstract> >(
@@ -29,7 +29,7 @@ void exposeContact1D() {
       "The calc and calcDiff functions compute the contact Jacobian and drift (holonomic constraint) or\n"
       "the derivatives of the holonomic constraint, respectively.",
       bp::init<boost::shared_ptr<StateMultibody>, pinocchio::FrameIndex, double, std::size_t,
-               bp::optional<Eigen::Vector2d, Contact1DMaskType, pinocchio::ReferenceFrame> >(
+               bp::optional<Eigen::Vector2d, Vector3MaskType, pinocchio::ReferenceFrame> >(
           bp::args("self", "state", "id", "xref", "nu", "gains", "mask", "type"),
           "Initialize the contact model.\n\n"
           ":param state: state of the multibody system\n"
@@ -37,7 +37,7 @@ void exposeContact1D() {
           ":param xref: contact position used for the Baumgarte stabilization\n"
           ":param nu: dimension of control vector\n"
           ":param gains: gains of the contact model (default np.matrix([0.,0.]))\n"
-          ":param mask: axis of the contact constraint (0=x, 1=y or 2=z)\n"
+          ":param mask: axis of the contact constraint (default z)\n"
           ":param type: reference type of contact"))
       .def(bp::init<boost::shared_ptr<StateMultibody>, pinocchio::FrameIndex, double,
                     bp::optional<Eigen::Vector2d, pinocchio::ReferenceFrame> >(
--- a/bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
+++ b/bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
@@ -15,11 +15,7 @@ namespace python {
 void exposeContact1D() {
   bp::register_ptr_to_python<boost::shared_ptr<ContactModel1D> >();
 
-  bp::enum_<Vector3MaskType>("Vector3MaskType")
-      .value("x", x)
-      .value("y", y)
-      .value("z", z)
-      .export_values();
+  bp::enum_<Vector3MaskType>("Vector3MaskType").value("x", x).value("y", y).value("z", z).export_values();
 
   bp::class_<ContactModel1D, bp::bases<ContactModelAbstract> >(
       "ContactModel1D",
--- a/bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
+++ b/bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
@@ -15,38 +15,38 @@ namespace python {
 void exposeContact1D() {
   bp::register_ptr_to_python<boost::shared_ptr<ContactModel1D> >();
 
-  bp::enum_<pinocchio::ReferenceFrame>("ReferenceFrame")
-      .value("LOCAL", pinocchio::LOCAL)
-      .value("WORLD", pinocchio::WORLD)
-      .value("LOCAL_WORLD_ALIGNED", pinocchio::LOCAL_WORLD_ALIGNED)
+  bp::enum_<Contact1DMaskType>("Contact1DMaskType")
+      .value("X_MASK", X_MASK)
+      .value("Y_MASK", Y_MASK)
+      .value("Z_MASK", Z_MASK)
       .export_values();
 
   bp::class_<ContactModel1D, bp::bases<ContactModelAbstract> >(
       "ContactModel1D",
       "Rigid 1D contact model.\n\n"
       "It defines a rigid 1D contact model (point contact) based on acceleration-based holonomic constraints, in the "
-      "z "
-      "direction.\n"
+      "x, y or z direction.\n"
       "The calc and calcDiff functions compute the contact Jacobian and drift (holonomic constraint) or\n"
       "the derivatives of the holonomic constraint, respectively.",
       bp::init<boost::shared_ptr<StateMultibody>, pinocchio::FrameIndex, double, std::size_t,
-               bp::optional<Eigen::Vector2d, std::size_t, pinocchio::ReferenceFrame> >(
-          bp::args("self", "state", "id", "xref", "nu", "gains", "type", "pinRefFrame"),
+               bp::optional<Eigen::Vector2d, Contact1DMaskType, pinocchio::ReferenceFrame> >(
+          bp::args("self", "state", "id", "xref", "nu", "gains", "mask", "type"),
           "Initialize the contact model.\n\n"
           ":param state: state of the multibody system\n"
           ":param id: reference frame id of the contact\n"
           ":param xref: contact position used for the Baumgarte stabilization\n"
           ":param nu: dimension of control vector\n"
           ":param gains: gains of the contact model (default np.matrix([0.,0.]))\n"
-          ":param type: axis of the contact constraint (0=x, 1=y or 2=z)\n"
-          ":param pinRefFrame: pin.ReferenceFrame in {LOCAL, WORLD, LOCAL_WORLD_ALIGNED}"))
-      .def(bp::init<boost::shared_ptr<StateMultibody>, pinocchio::FrameIndex, double, bp::optional<Eigen::Vector2d, pinocchio::ReferenceFrame> >(
+          ":param mask: axis of the contact constraint (0=x, 1=y or 2=z)\n"
+          ":param type: reference type of contact"))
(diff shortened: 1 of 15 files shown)


bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
```diff
@@ -15,6 +15,12 @@ namespace python {
 void exposeContact1D() {
   bp::register_ptr_to_python<boost::shared_ptr<ContactModel1D> >();
 
+  bp::enum_<pinocchio::ReferenceFrame>("ReferenceFrame")
```
jcarpent, line 18: You should not duplicate this. This is a Pinocchio feature.

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
```diff
@@ -24,21 +30,23 @@ void exposeContact1D() {
       "The calc and calcDiff functions compute the contact Jacobian and drift (holonomic constraint) or\n"
       "the derivatives of the holonomic constraint, respectively.",
       bp::init<boost::shared_ptr<StateMultibody>, pinocchio::FrameIndex, double, std::size_t,
-               bp::optional<Eigen::Vector2d> >(
-          bp::args("self", "state", "id", "xref", "nu", "gains"),
+               bp::optional<Eigen::Vector2d, std::size_t, pinocchio::ReferenceFrame> >(
+          bp::args("self", "state", "id", "xref", "nu", "gains", "type", "pinRefFrame"),
```
jcarpent, line 34: Set default argument here.

include/crocoddyl/multibody/contacts/contact-1d.hxx
```diff
@@ -9,17 +9,36 @@
...
 template <typename Scalar>
-ContactModel1DTpl<Scalar>::ContactModel1DTpl(boost::shared_ptr<StateMultibody> state, const pinocchio::FrameIndex id,
-                                             const Scalar xref, const std::size_t nu, const Vector2s& gains)
-    : Base(state, 1, nu), xref_(xref), gains_(gains) {
+ContactModel1DTpl<Scalar>::ContactModel1DTpl(boost::shared_ptr<StateMultibody> state, 
+                                             const pinocchio::FrameIndex id,
+                                             const Scalar xref, 
+                                             const std::size_t nu, 
+                                             const Vector2s& gains, 
+                                             const std::size_t& type,
```
jcarpent, line 17: The reference is not needed as it is a basic type.

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
jcarpent: If you want to have this thing exposed in Crocoddyl, there is a specific helper function to import ReferenceFrame within the Croco namespace, assuming Pinocchio is exposed first.

jcarpent (changes requested):

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
cmastalli: Let's do not expose this in Crocoddyl.
There is no reason to duplicate this code.

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
cmastalli: Also rename the arguments as follows:
 - `type` to `mask`
 - `pinRefFrame` to `type`

This latter follows the standard used in other classes.

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
```diff
@@ -24,21 +30,23 @@ void exposeContact1D() {
...
           "Initialize the contact model.\n\n"
           ":param state: state of the multibody system\n"
           ":param id: reference frame id of the contact\n"
           ":param xref: contact position used for the Baumgarte stabilization\n"
           ":param nu: dimension of control vector\n"
-          ":param gains: gains of the contact model (default np.matrix([0.,0.]))"))
-      .def(bp::init<boost::shared_ptr<StateMultibody>, pinocchio::FrameIndex, double, bp::optional<Eigen::Vector2d> >(
+          ":param gains: gains of the contact model (default np.matrix([0.,0.]))\n"
+          ":param type: axis of the contact constraint (0=x, 1=y or 2=z)\n"
+          ":param pinRefFrame: pin.ReferenceFrame in {LOCAL, WORLD, LOCAL_WORLD_ALIGNED}"))
```
cmastalli, line 42: Please adapt the documentation as suggested above

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
```diff
@@ -24,21 +30,23 @@ void exposeContact1D() {
...
+      .def(bp::init<boost::shared_ptr<StateMultibody>, pinocchio::FrameIndex, double, bp::optional<Eigen::Vector2d, pinocchio::ReferenceFrame> >(
           bp::args("self", "state", "id", "xref", "gains"),
           "Initialize the contact model.\n\n"
           ":param state: state of the multibody system\n"
           ":param id: reference frame id of the contact\n"
           ":param xref: contact position used for the Baumgarte stabilization\n"
-          ":param gains: gains of the contact model (default np.matrix([0.,0.]))"))
+          ":param gains: gains of the contact model (default np.matrix([0.,0.]))\n"))
```
cmastalli, line 49: Revert this back

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
```diff
@@ -68,7 +76,10 @@ void exposeContact1D() {
                     &ContactModel1D::set_reference, "reference contact translation")
       .add_property("gains",
                     bp::make_function(&ContactModel1D::get_gains, bp::return_value_policy<bp::return_by_value>()),
-                    "contact gains");
+                    "contact gains")
+      .add_property("pinReferenceFrame",
```
cmastalli, line 80: 1. Renamed this as suggested above
2. Add the getter and setter for the `mask`

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
```diff
@@ -68,7 +76,10 @@ void exposeContact1D() {
...
+                    bp::make_function(&ContactModel1D::get_pinReferenceFrame, bp::return_value_policy<bp::return_by_value>()),
+                    &ContactModel1D::set_pinReferenceFrame, "pin.ReferenceFrame");
```
cmastalli, line 82: Change documentation to `reference type of contact`

include/crocoddyl/multibody/contacts/contact-1d.hpp
```diff
@@ -37,6 +37,7 @@ class ContactModel1DTpl : public ContactModelAbstractTpl<_Scalar> {
   typedef typename MathBase::Vector3s Vector3s;
   typedef typename MathBase::VectorXs VectorXs;
   typedef typename MathBase::Matrix3s Matrix3s;
+  typedef typename MathBase::Vector6s Vector6s;
```
cmastalli, line 40: Remove this line. It doesn't seem to be needed

include/crocoddyl/multibody/contacts/contact-1d.hxx
```diff
@@ -9,17 +9,36 @@
...
 
 template <typename Scalar>
-ContactModel1DTpl<Scalar>::ContactModel1DTpl(boost::shared_ptr<StateMultibody> state, const pinocchio::FrameIndex id,
-                                             const Scalar xref, const Vector2s& gains)
-    : Base(state, 1), xref_(xref), gains_(gains) {
+ContactModel1DTpl<Scalar>::ContactModel1DTpl(boost::shared_ptr<StateMultibody> state, 
+                                             const pinocchio::FrameIndex id,
+                                             const Scalar xref, 
+                                             const Vector2s& gains,
+                                             const pinocchio::ReferenceFrame pinRefFrame)
```
cmastalli, line 34: Change the name of the arguments as suggested above.

include/crocoddyl/multibody/contacts/contact-1d.hxx
```diff
@@ -9,17 +9,36 @@
...
-                                             const Scalar xref, const std::size_t nu, const Vector2s& gains)
-    : Base(state, 1, nu), xref_(xref), gains_(gains) {
+ContactModel1DTpl<Scalar>::ContactModel1DTpl(boost::shared_ptr<StateMultibody> state, 
+                                             const pinocchio::FrameIndex id,
+                                             const Scalar xref, 
+                                             const std::size_t nu, 
+                                             const Vector2s& gains, 
+                                             const std::size_t& type, 
+                                             const pinocchio::ReferenceFrame pinRefFrame)
+    : Base(state, 1, nu), xref_(xref), gains_(gains), type_(type), pinReferenceFrame_(pinRefFrame) {
```
cmastalli, line 19: Change the name of the arguments as suggested above

include/crocoddyl/multibody/contacts/contact-1d.hxx
```diff
@@ -9,17 +9,36 @@
...
   id_ = id;
+  if(type_ < 0 || type_ > 2){
```
cmastalli, line 21: This doesn't follow the code format. Please run the code formatter as described in the wiki

include/crocoddyl/multibody/contacts/contact-1d.hxx
cmastalli: Better, please define an enum for `type`.

cmastalli (changes requested):
There are still many changes to be tackled.
However, I don't understand why you need to define a reference frame that is not LOCAL.
These classes are not aimed to track a reference acceleration for instance.
So, why?

skleff1994:
Thanks for your review , I will now implement some of the requested changes.
 
I also noticed that the [contact force residual computation](https://github.com/skleff1994/crocoddyl/blob/31acf4adb4e556e15e34eaabe2196a447f6c082e/include/crocoddyl/multibody/residuals/contact-force.hxx#L44) needs to be updated as well  , and therefore be aware of the mask of the contact model .

@cmastalli  the purpose of this enhancement is to allow force computations in the WORLD frame, which is necessary for e.g. sanding tasks where the desired force is defined in WORLD (the LOCAL z-axis is not necessarily aligned with the vertical direction).  In that case , I believe we need to perform all contact-related computations in LOCAL_WORLD_ALIGNED , which is currently not allowed by Crocoddyl.

@jcarpent I believe that Baumgarte stabilizing term is already expressed in WORLD for all contact models - which I also don't really understand as the acceleration and force computations are currently in LOCAL .  Wouldn't it make sense to compute the Baumgarte terms according in the pinocchio reference frame ?

skleff1994:
Concerning the force residual, I was thinking I could add the mask to ContactData1D so that it can be accessed in the force residual computation ( we need to select x, y, or z component depending on the mask of the contact model ). Also , we should pass the pinocchio reference as an optional argument in force residual constructor to be able to compute the 1D force residual in any frame . From there, this could be straightforwardly extended to other contact models I believe . What do you think ?

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
skleff1994: @jcarpent  concerning the default argument I had

`( bp::args("self", "state", "id", "xref", "nu", "gains", "type"), bp::arg("pinRefFrame")=pinocchio::LOCAL ),`

but it seemed to make python unittest fail and caused an import error in python, I couldn't figure out why . I need to investigate this . Any clue?

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
skleff1994: [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
skleff1994: @cmastalli  addressed in [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

include/crocoddyl/multibody/contacts/contact-1d.hxx
skleff1994: addressed in [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
skleff1994: addressed in [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
skleff1994: addressed in [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
skleff1994: addressed in [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

include/crocoddyl/multibody/contacts/contact-1d.hpp
skleff1994: addressed in [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

include/crocoddyl/multibody/contacts/contact-1d.hxx
skleff1994: addressed in [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

include/crocoddyl/multibody/contacts/contact-1d.hxx
skleff1994: addressed in [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

include/crocoddyl/multibody/contacts/contact-1d.hxx
skleff1994: addressed in [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

bindings/python/crocoddyl/multibody/contacts/contact-1d.cpp
skleff1994: addressed in [fd58261](https://github.com/loco-3d/crocoddyl/pull/1054/commits/fd582614a34b626d8cb0751df8132d31f643845b)

include/crocoddyl/multibody/contacts/contact-1d.hxx
```diff
@@ -30,22 +49,25 @@ void ContactModel1DTpl<Scalar>::calc(const boost::shared_ptr<ContactDataAbstract
...
+  d->Jc.row(0) = d->fJf.row(type_);
   d->vw = d->v.angular();
   d->vv = d->v.linear();
-
-  d->a0[0] = d->a.linear()[2] + d->vw[0] * d->vv[1] - d->vw[1] * d->vv[0];
+  d->a0[0] = mask_.dot( d->a.linear() + d->vw.cross(d->vv) );
 
   if (gains_[0] != 0.) {
-    d->a0[0] += gains_[0] * (d->pinocchio->oMf[id_].translation()[2] - xref_);
+    if(pinReferenceFrame_ == pinocchio::WORLD){
```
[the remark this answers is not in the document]
skleff1994: I believe that Baumgarte stabilizing term is already expressed in WORLD for all contact models - which I also don't really understand as the acceleration and force computations are currently in LOCAL .  Wouldn't it make sense to compute the Baumgarte terms in the pinocchio reference frame ?

include/crocoddyl/multibody/contacts/contact-1d.hxx
```diff
@@ -90,9 +100,23 @@ void ContactModel1DTpl<Scalar>::updateForce(const boost::shared_ptr<ContactDataA
...
+    case pinocchio::LOCAL: {
+      jMc = d->jMf;
+      break;
+    }
+    case pinocchio::WORLD: {
+      jMc = d->jMf.act(d->pinocchio->oMf[id_].inverse());
+      break;
+    }
+    case pinocchio::LOCAL_WORLD_ALIGNED: {
+      jMc = pinocchio::SE3(Matrix3s::Identity(), d->jMf.translation());
```
cmastalli, line 114: This is allocating data dynamically.
Instead, you should modify the translation only, i.e.
```suggestion
 jMc_.translation = d->jMf.translation();
```

Then, you have to set the identity matrix only when you have define `LOCAL_WORLD_ALIGNED`

cmastalli, line 114: This is allocating data dynamically.
Instead, you should modify the translation only, i.e.
```suggestion
 jMc_.translation = d->jMf.translation();
```

Then, you have to set the identity matrix only when you have define `LOCAL_WORLD_ALIGNED`

include/crocoddyl/multibody/contacts/contact-1d.hxx
```diff
@@ -90,9 +100,23 @@ void ContactModel1DTpl<Scalar>::updateForce(const boost::shared_ptr<ContactDataA
                  << "lambda has wrong dimension (it should be 1)");
   }
   Data* d = static_cast<Data*>(data.get());
-  const Eigen::Ref<const Matrix3s> R = d->jMf.rotation();
-  data->f.linear() = R.col(2) * force[0];
-  data->f.angular() = d->jMf.translation().cross(data->f.linear());
+  pinocchio::SE3 jMc;
```
cmastalli, line 103: Let's define a global variable for this, and called it `jMc_`.
This avoids to allocate data dynamically.

cmastalli, line 103: Let's define a global variable for this, and called it `jMc_`.
This avoids to allocate data dynamically.

include/crocoddyl/multibody/contacts/contact-1d.hxx
```diff
@@ -121,4 +145,24 @@ void ContactModel1DTpl<Scalar>::set_reference(const Scalar reference) {
...
+}
+
+template <typename Scalar>
+const pinocchio::ReferenceFrame ContactModel1DTpl<Scalar>::get_type() const {
+  return type_;
+}
+
+template <typename Scalar>
+void ContactModel1DTpl<Scalar>::set_mask(const Contact1DMaskType mask) {
+  mask_ = mask;
```
cmastalli, line 160: If you change to `LOCAL_WORLD_ALIGNED`, then define the identity matrix as pointed out above.

cmastalli, line 160: If you change to `LOCAL_WORLD_ALIGNED`, then define the identity matrix as pointed out above.

cmastalli (changes requested):
Please see all the comments below.
A part of them, we also need to unittest the different mask cases.
To do so, you need to extend the contact factory.

skleff1994:
Thanks for the review , I addressed the requested changes in [1fecfb4b](https://github.com/loco-3d/crocoddyl/pull/1054/commits/1fecfb4bc96b4ebf966e2a443ddaf87b1bc86cf1) . 

In addition to those , I implemented 
- force computation using mask and reference type for contact 1D
- specific unittest for 1D contact ( testing over masks and reference frames ) 

Currently, the unittest pass only in LOCAL and WORLD. There seem to be a mistake in calcDiff for LOCAL_WORLD_ALIGNED but I can't spot it..

skleff1994:
So we wrote down the maths with @nmansard  and derived correct LOCAL_WORLD_ALIGNED derivatives for acceleration, I will fix the current implementation asap. 

Also @jcarpent  since the WORLD case doesn't make sense as you say I can remove it from the API. But then I think we should also clarify the Baumgarte terms: is there a reason why they are expressed in WORLD currently?

cmastalli:
I will close this issue due to a lack of activity. @skleff1994 -- feel free to re-open it anytime you wish.

PR #1058, opened 2022-03-28
Simple urdf example

justinberi:
**Description**

There was no simplified example using an urdf and activation model and I thought this might be useful to others who are also starting out.

All the existing examples (from example-robots-data) require meshes which are annoying for prototyping simple configurations. 

If you wish I can also do up a cartpole example.

### DIFF
--- a/examples/robots/double_pendulum_simple.urdf
+++ b/examples/robots/double_pendulum_simple.urdf
@@ -1,0 +1,174 @@
+<?xml version="1.0" encoding="utf-8"?>
+<robot
+  name="2dof_planar">
+  <link
+    name="base_link">
+    <inertial>
+      <origin
+        xyz="0 0 0"
+        rpy="0 0 0" />
+      <mass
+        value="0.1" />
+      <inertia
+        ixx="1"
+        ixy="0"
+        ixz="0"
+        iyy="1"
+        iyz="0"
+        izz="1" />
+    </inertial>
+    <visual>
+      <origin
+        xyz="0 0 0"
+        rpy="0 0 0" />
+      <geometry>
+        <box size="0.05 0.05 0.05"/>
+      </geometry>
+      <material
+        name="">
+        <color
+          rgba="1 0 0 1" />
+      </material>
+    </visual>
+    <collision>
+      <origin
+        xyz="0 0 0"
+        rpy="0 0 0" />
+      <geometry>
+        <box size="0.05 0.05 0.05"/>
+      </geometry>
+    </collision>
+  </link>
+  <link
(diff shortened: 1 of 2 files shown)


hrp2-14:
Hi ! This project doesn't usually accept pull requests on master. If this wasn't intentionnal, you
can change the base branch of this pull request to devel (No need to close it for that). Best, a bot.

cmastalli (changes requested):
Hi @justinberi 

Thank you for being a proactive member of our community.
We are keen to have many contributors like you.

However, for this particular case, I need to unfortunately reject the PR in its form.
We would like to avoid having URDF (or robot related) code in our repository.
We developed `example-robot-data` for such purpose.

Said that, I believe your PR goal is fair.
If you wish to pursue it, then I suggest to take the following action points:
 1. Substitute the URDF for a Pinocchio code that creates this model.
 - This shouldn't be difficult if you know enough Pinocchio.
 2. Create a Jupyter notebook, instead of a Python script
 - This will be nicer as you go step-by-step in the creation of the robot and actuation models

Let me know if you wish to do it. Otherwise, I will close this PR as I don't have bandwidth to resolve these action points.

Again, thanks for being proactive!

nim65s:
Hi @justinberi,
I agree that it would be nice to have a simple robot model with basic geometries instead of meshes (in example-robot-data, only `simple_humanoid_description` doesn't have meshes, but it doesn't have visual at all…). 

But as @cmastalli noted, we usually share our robot models among our different projects in that example-robot-data package. Therefore, I guess you could open a PR with this urdf there. Would this work for you ?

wxmerkt:
We can always use `pinocchio.buildSampleModelManipulator` and `pinocchio.buildSampleGeometryModelHumanoid` or manual creation as pointed out by Carlos to have examples that won't require example-robot-data

justinberi:
Hi all,

Thank you for the feedback. All seems logical and I'd like to pursue the idea. The goal of the PR was to create an urdf and example script that others could use to easily different robot configurations. So I will do the following:

1. Close this PR.
2. Create a PR (or issue?) in example-robot-data and move `double_pendulum_simple.urdf` to that repo.
3. Create a PR (or issue?) in crocoddyl for and convert `simple_urdf.py` to a notebook.

@cmastalli I may create a separate issue to demonstrate pinocchio for creating the model in the future but I will limit the scope of this work for moment

Does that seem fair?

cmastalli:
> Does that seem fair?

Yes, it does.
Note that we need to wait for a new release (in example-robot-data) before merging the notebook in Crocoddyl.

PR #1060, opened 2022-03-29
FDDP: Do not allow ascent steps once feasibility has been achieved

wxmerkt:
This is to fix a case where the solver, close to convergence, accepted a significantly negative dV-Exp after already having achieved feasibility - resulting in divergence.

Cf. the following output - this fix avoids the disimprovement in iteration 99 from 2.3e-1 to 1.5e+7
```
98  2.28141e-01  2.66446e-03  3.87063e+03  1.00000e-03  1.00000e-03 0.0010     1  6.24123e+02  -9.65938e+02
99  1.55556e+07  6.67014e+10  3.26809e+08  1.00000e-02  1.00000e-02 1.0000     1  -1.62977e+08  -1.55556e+07
iter     cost         stop          grad          xreg        ureg step    feas   dV-exp        dV
100  7.12945e+06  1.52536e+10  1.79564e+03  1.00000e-03  1.00000e-03 0.5012     1  -1.96396e+06  8.42617e+06
```

I've not seen a negative impact in the stored logs, but it might be possible -- is there a potential downside to this fix? Would there be a better fix?


cmastalli:
Could you rebase your code and then update the log files?
I would like to inspect the different before merging this PR ;)

PR #1061, opened 2022-03-30
Add option to adjust verbosity level/precision; check log files as part of tests

wxmerkt:
This PR adds extra functionality to modify the verbosity and precision of the `CallbackVerbose`. It subsequently updates all log files, and then adds a test that checks the logfiles as part of the test targets / CI.

Hopefully the numerical results on CI aren't too different from our developer machines, otherwise we'd have to deactivate the test.


examples/CMakeLists.txt
```diff
@@ -22,7 +22,9 @@ FOREACH(examples ${${PROJECT_NAME}_EXAMPLES_PYTHON})
     ${PYTHON_EXECUTABLE} -c "import ${examples}" \${INPUT})
 
   # examples are too slow in Debug mode to be used as tests
-  IF(CMAKE_BUILD_TYPE STREQUAL "Release")
+  IF(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
     ADD_PYTHON_UNIT_TEST("example-python-${examples}" "examples/${examples}.py" bindings/python)
-  ENDIF(CMAKE_BUILD_TYPE STREQUAL "Release")
+    # Compare logs (NB: this depends on /bin/bash so won't work on non-Unix systems)
```
cmastalli, line 27: I guess, we check if this is a unix system.
Not sure how to do it.

cmastalli (approved):
Thanks @wxmerkt!

I just have one comment.
See below

examples/CMakeLists.txt
wxmerkt: Good point - I've added the check and fixed an issue with the default python version in the update_logfiles script. It should now work

cmastalli:
Thank you so much @wxmerkt !!!

This was very much needed for a long time ;)

PR #1059, opened 2022-03-29
Adds acrobot example

justinberi:
Adds an example for acrobot. See https://github.com/loco-3d/crocoddyl/pull/1058

PR #1063, opened 2022-04-05
add dummy .pre-commit-config.yaml

nim65s:
ref. https://github.com/stack-of-tasks/pinocchio/pull/1629 (also for the loco-3d organization)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,8 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0
+    hooks:
+    -   id: check-json
+    -   id: check-symlinks
+    -   id: check-toml
+    -   id: check-yaml
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,8 @@
+ci:
+    autoupdate_branch: 'devel'
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.1.0
+    rev: v4.2.0
     hooks:
     -   id: check-json
     -   id: check-symlinks


cmastalli (approved):
Thanks @nim65s 

I am curious, what are the typical changes that this tool does automatically?

nim65s:
This tool mainly help to run other tools. It comes with some simple hooks : https://pre-commit.com/hooks.html but you can add any tool you want. And we intend to to that with clang-format and black (which is replacing yapf…)

nim65s:
This is boring… All versions above 3.7.0 are fine, including 4.0.0, but it looks like we can't tell that to CMake.

So we could either consider that 3.7.0 is old enough and drop the number in the requirements, or handle this manually. I tried the second option in a couple places, like : https://github.com/humanoid-path-planner/hpp-environments/pull/24/commits/ad545a1eb3802452dd1d710c2b897f21bf93b23b

nim65s:
But for the particular case of #1059, we might need to bump this requirement to 4.0.0 in crocoddyl anyway ^^"

cmastalli:
> But for the particular case of #1059, we might need to bump this requirement to 4.0.0 in crocoddyl anyway ^^"

Then, in our case, it seems better to go consider 4.0.0 as the required version (rather than manually handling it).
We just need to release this version first. Also to be able to merge #1059.

Perhaps, let's try to merge this PR first.

wxmerkt:
> Then, in our case, it seems better to go consider 4.0.0 as the required version (rather than manually handling it).
We just need to release this version first. Also to be able to merge https://github.com/loco-3d/crocoddyl/pull/1059.

jrl-cmakemodules has been fixed in the meantime and bumping it and fixing the CMakeLists to accept AnyNewerVersion will resolve the issue :-)

nim65s:
Exactly, and 4.0.1 has been released to include this fix

nim65s:
The CI looks happy now, and I installed pre-commit.ci on the loco-3d.

cmastalli:
> Exactly, and 4.0.1 has been released to include this fix

Are you talking about `example-robot-data`?
If so, why I don't see this version is released here: http://robotpkg.openrobots.org/robotpkg/robots/example-robot-data/index.html?

nim65s:
Because I'm really late on robotpkg

PR #1074, opened 2022-05-19
Sync submodule cmake

jcarpent:
After that, I will do a new release to fix packaging issues.
This is needed by conda for instance. @cmastalli Is it fine for you?


cmastalli (approved):
Yes, it is fine if we do a new release.
I would also double check that it is fine for @nim65s.
 
Let's call this version v1.9.1.

PR #1084, opened 2022-08-01
Fix CMake for eigenpy 2.7.10

nim65s:


jcarpent (approved):

cmastalli:
Does the submodule update requires an update in eigenpy?
Or is there any reason to do so?

nim65s:
the submodule update is not required here.
The main issue is that we now have `eigenpyConfig.cmake` correctly calling `FINDPYTHON()` /`FIND_NUMPY()` automatically for us, but therefore if we try to call it manually after or before eigenpy, it will raise an error.

cmastalli (approved):

nim65s:
@wxmerkt : can we turn off ROS CI Format job ? we decided not to reformat everything in crocoddyl until we stabilze  the API for v2

wxmerkt:
Sure, happy to turn it off for now

wxmerkt (approved):
CI is now green. Note that I had to turn the Bionic/18.04 builds with Python2 off as it could not find Python without explicitly specifying PYTHON_EXECUTABLE.

dependencies.rosinstall
```diff
@@ -1,4 +1,4 @@
 - git:
     uri: https://github.com/Gepetto/example-robot-data.git
-    version: master
+    version: devel
```
cmastalli, line 3: Is there a reason why we need to use this devel branch?

cmastalli:
> CI is now green. Note that I had to turn the Bionic/18.04 builds with Python2 off as it could not find Python without explicitly specifying PYTHON_EXECUTABLE.

@nim65s do you know how to tackle this? Or are we planning to do not support 18.04/Python 2 anymore?

nim65s:
Don't know. Let's try with just a submodule update

nim65s:
Looks like an issue in example-robot-data

dependencies.rosinstall
wxmerkt: Because we need a fix from the devel branch until there is another example-robot-data release

cmastalli:
Thanks guys @nim65s and @wxmerkt :)

PR #1091, opened 2022-09-26
Vanila constraint API

cmastalli:
Hi @nmansard -- here you can find a smaller PR that contains the constraint API.

There are a few extra things that I left out of this PR. Those things are needed for MPC, however, this PR is a good enough start for implementing constrained OC solvers.

Please be aware that I couldn't add those features as they are developed along with other aspects to be merged later.

@wxmerkt @nim65s -- This PR has the same code than the other one. There are changes that improves the documentation that I didn't include in this PR.


cmastalli:
@wxmerkt -- could you update about the issues with the ROS jobs?

wxmerkt:
Ubuntu Bionic / ROS Melodic has issues since Pinocchio currently does not build on Melodic. I'll check it out tomorrow

wxmerkt:
I've restarted the jobs since the issue on Melodic which blocked the other builds is now fixed.

Since this PR is a cherry-picked/rebased version of #1080, this may cause issues when having re-rebase branches. I haven't seen any comments from Nicolas on the other PR or here to hold up either -- but if/once this PR "looks good", would it make sense to instead merge #1080 which includes this PR + the new features to avoid re-rebasing all working branches? What is the current plan and feedback on moving ahead?

cmastalli:
> I've restarted the jobs since the issue on Melodic which blocked the other builds is now fixed.
> 
> Since this PR is a cherry-picked/rebased version of #1080, this may cause issues when having re-rebase branches. I haven't seen any comments from Nicolas on the other PR or here to hold up either -- but if/once this PR "looks good", would it make sense to instead merge #1080 which includes this PR + the new features to avoid re-rebasing all working branches? What is the current plan and feedback on moving ahead?

I have rebased #1080 in another branch. I am keeping the one in #1080 as a backup.

@nmansard asked me to create a smaller PR with the constraint API as he is not able to review the code. I might later slip again the remaining PR if there are further concerns.

@nmansard -- please let us know if you have any comment regarding this PR. There are many ongoing activities and it is important for us to iterate faster.

include/crocoddyl/core/constraints/residual.hxx
```diff
@@ -0,0 +1,301 @@
...
+    if (constraint_type_(i) == 0) {  // equality constraint
+      if (is_rq && is_rv) {
+        data->Hx.row(nh_i) = data->residual->Rx.row(i);
+      } else if (is_rq) {
+        data->Hx.row(nh_i).head(nv) = data->residual->Rx.row(i).head(nv);
+      } else if (is_rv) {
+        data->Hx.row(nh_i).tail(nv) = data->residual->Rx.row(i).tail(nv);
+      }
+      ++nh_i;
+    } else if (constraint_type_(i) == 1) {  // inequality constraint
```
nmansard, line 206: use an enum for this.

include/crocoddyl/core/constraints/residual.hxx
```diff
@@ -0,0 +1,301 @@
...
+
+  // Fill the residual values for its corresponding type of constraint
+  std::size_t nh_i = 0;
+  std::size_t ng_i = 0;
+  const std::size_t nv = state_->get_nv();
+  const std::size_t nr = residual_->get_nr();
+  const bool is_rq = residual_->get_q_dependent();
+  const bool is_rv = residual_->get_v_dependent();
+  for (std::size_t i = 0; i < nr; ++i) {
+    if (constraint_type_(i) == 0) {  // equality constraint
```
nmansard, line 197: this switch would make the copy of the jacobians very unefficient. I dont think we should be that permissive with the constraints type. See main review.

include/crocoddyl/core/constraints/residual.hxx
```diff
@@ -0,0 +1,301 @@
...
+  }
+  if ((lower.array() == std::numeric_limits<Scalar>::infinity()).any()) {
+    throw_pretty("Invalid argument: the lower bound cannot containt a positive infinity value");
+  }
+  if ((upper.array() == -std::numeric_limits<Scalar>::infinity()).any()) {
+    throw_pretty("Invalid argument: the lower bound cannot containt a negative infinity value");
+  }
+
+  // Update the information related to the bounds
+  ng_ = 2 * ((upper - lower).array() != 0.).count() -
```
nmansard, line 269: I am not sure where you update the data buffers (Gx, Gu ... ) when you do this. Can you check, and at least mention it in a comment.

include/crocoddyl/core/constraints/residual.hxx
```diff
@@ -0,0 +1,301 @@
...
+#include "crocoddyl/core/utils/exception.hpp"
+
+namespace crocoddyl {
+
+template <typename Scalar>
+ConstraintModelResidualTpl<Scalar>::ConstraintModelResidualTpl(boost::shared_ptr<typename Base::StateAbstract> state,
+                                                               boost::shared_ptr<ResidualModelAbstract> residual,
+                                                               const VectorXs& lower, const VectorXs& upper)
+    : Base(state, residual,
+           2 * ((upper - lower).array() != 0.).count() -
```
nmansard, line 18: this is very inconvenient to read, and error prone if further modifications are needed. Please formulate it differently (ideally, initialize the ng and nh counters later).

include/crocoddyl/core/constraints/residual.hxx
```diff
@@ -0,0 +1,301 @@
...
+  // Fill the residual values for its corresponding type of constraint
+  std::size_t nh_i = 0;
+  std::size_t ng_i = 0;
+  const std::size_t nr = residual_->get_nr();
+  for (std::size_t i = 0; i < nr; ++i) {
+    if (constraint_type_(i) == 0) {  // equality constraint
+      data->h(nh_i) = data->residual->r(i) - ub_(i);
+      ++nh_i;
+    } else if (constraint_type_(i) == 1) {  // inequality constraint
+      data->g(ng_i) = data->residual->r(i) - ub_(i);
```
nmansard, line 70: This is a bad strategy to loose the information on the double bounds. Better store both for later use by the solver.

include/crocoddyl/core/action-base.hpp
```diff
@@ -217,9 +248,15 @@ class ActionModelAbstractTpl {
...
   virtual void print(std::ostream& os) const;
 
+ private:
+  std::size_t ng_internal_;  //!< Internal object for storing the number of inequatility constraints
+  std::size_t nh_internal_;  //!< Internal object for storing the number of equatility constraints
+
  protected:
   std::size_t nu_;                          //!< Control dimension
   std::size_t nr_;                          //!< Dimension of the cost residual
+  std::size_t* ng_;                         //!< Number of inequality constraints
```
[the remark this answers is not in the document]
cmastalli: > Ok, I reviewed the code. Seems good in general. The "shareMemory" design is kept from cost-sum, which is good.
> 
> There is three major flaws to possibly discuss:
> 
> * the strange design pattern you used to store the ng and nh counters.

This design is needed for MPC applications when the dimension of constraints (e.g., contact constraints) changes at each MPC iteration. Again, I introduced this strategy needed in our inverse-dynamics MPC. Do you see a better alternative here? (e.g., smart pointers) 

> * the choice of allowing various constraint types (==, <=, <=<=) for a single residual. I dont think it is a good idea as it would be much less efficient to copy the memory from the residual model derivatives to the constraint. And I dont see much use of this in practice, do you?

This is indeed a good point. We could create different ConstraintResidual classes, one generic as proposed here. And other ones tailored to the cases ==, < and >. Let me know if you have alternative suggestions here as well. (I could implement these solutions in this PR or later; depending on how critical is to have it now)


> * you are evaluating double bounds as two rows in the inequality vector and derivatives. This is not efficient and many solvers would make use of the information.

It seems a reasonable choice to not tighten the bounds. I will handle it soon.

cmastalli:
Thank @nmansard for your feedback. See below my comments. Please provide a feedback about the point 1 and 2.

> Ok, I reviewed the code. Seems good in general. The "shareMemory" design is kept from cost-sum, which is good.
> 
> There is three major flaws to possibly discuss:
> 
> * the strange design pattern you used to store the ng and nh counters.

This design is needed for MPC applications when the dimension of constraints (e.g., contact constraints) changes at each MPC iteration. Again, I introduced this strategy needed in our inverse-dynamics MPC. Do you see a better alternative here? (e.g., smart pointers) 

> * the choice of allowing various constraint types (==, <=, <=<=) for a single residual. I dont think it is a good idea as it would be much less efficient to copy the memory from the residual model derivatives to the constraint. And I dont see much use of this in practice, do you?

This is indeed a good point. We could create different ConstraintResidual classes, one generic as proposed here. And other ones tailored to the cases ==, < and >. Let me know if you have alternative suggestions here as well. (I could implement these solutions in this PR or later; depending on how critical is to have it now)


> * you are evaluating double bounds as two rows in the inequality vector and derivatives. This is not efficient and many solvers would make use of the information.

It seems a reasonable choice to not tighten the bounds. I will handle it soon.

nmansard:
> > * the strange design pattern you used to store the ng and nh counters.
> 
> This design is needed for MPC applications when the dimension of constraints (e.g., contact constraints) changes at each MPC iteration. Again, I introduced this strategy needed in our inverse-dynamics MPC. Do you see a better alternative here? (e.g., smart pointers)

It is not a sufficient explanation to justify we go with a weird risky pattern in the very root of the API. You can think of several way of handling activation and deactivation of constraints that do not need something like that. 
We should maybe discuss more in depth what can of access point we allow in the low-level API for varying OCP, ie OCP whose very nature (number of constraints or number of terms in the cost sum) varies with time. If we believe that this is a key feature, then we could choose better ways of implementing it. In that case I would rather be in favor of a clean explicit set of methods in both model and data to handle it. 
In any case, going with a pointer to an integer is not a sane way, let's remove that.


> 
> > * the choice of allowing various constraint types (==, <=, <=<=) for a single residual. I dont think it is a good idea as it would be much less efficient to copy the memory from the residual model derivatives to the constraint. And I dont see much use of this in practice, do you?
> 
> This is indeed a good point. We could create different ConstraintResidual classes, one generic as proposed here. And other ones tailored to the cases ==, < and >. Let me know if you have alternative suggestions here as well. (I could implement these solutions in this PR or later; depending on how critical is to have it now)

I don't think we need one class per type: a single class with an enum would be likely enough, just the enum should be for all the rows of the residual. If you feel you need to keep a second class with versatile constraint types for each row, then why not, it does not hurt (it is indeed one of the nice part of decoupling residual from constraint and cost definition). We just have to make sure it is clearly documented.

> 
> > * you are evaluating double bounds as two rows in the inequality vector and derivatives. This is not efficient and many solvers would make use of the information.
> 
> It seems a reasonable choice to not tighten the bounds. I will handle it soon.

Thx

cmastalli:
> > This design is needed for MPC applications when the dimension of constraints (e.g., contact constraints) changes at each MPC iteration. Again, I introduced this strategy needed in our inverse-dynamics MPC. Do you see a better alternative here? (e.g., smart pointers)
> 
> It is not a sufficient explanation to justify we go with a weird risky pattern in the very root of the API. You can think of several way of handling activation and deactivation of constraints that do not need something like that. We should maybe discuss more in depth what can of access point we allow in the low-level API for varying OCP, ie OCP whose very nature (number of constraints or number of terms in the cost sum) varies with time. If we believe that this is a key feature, then we could choose better ways of implementing it. In that case I would rather be in favor of a clean explicit set of methods in both model and data to handle it. In any case, going with a pointer to an integer is not a sane way, let's remove that.

We already have an activation/deactivation strategy for constraint. However, here we need to change online the dimension of the data in the action model. This information is also used by the solver to redefine the dimension of its data. This feature is important for inverse-dynamics MPC, but useful in general.

We use a pointer as the dimension could be defined by the action model, diff action model or constraint manager. c++ and Python developers do not interact with pointers but rather with objects. In my humbler opinion, this is not error-prone as we do not permit developers to manipulate the pointers.

However, the only alternative that I see here is to define a `std::map` that contains the constraint data in the action model. This definitely introduces algorithmic complexity when parsing the data by the solver. If we do so, we also need to expose the activation information in the action model, so, the solver will parse the items that are active.

We should perhaps discuss it in private about it.  I am not convinced to go for a `std::map` and do not agree that this is error-prone (as you can see above).

> > This is indeed a good point. We could create different ConstraintResidual classes, one generic as proposed here. And other ones tailored to the cases ==, < and >. Let me know if you have alternative suggestions here as well. (I could implement these solutions in this PR or later; depending on how critical is to have it now)
> 
> I don't think we need one class per type: a single class with an enum would be likely enough, just the enum should be for all the rows of the residual. If you feel you need to keep a second class with versatile constraint types for each row, then why not, it does not hurt (it is indeed one of the nice part of decoupling residual from constraint and cost definition). We just have to make sure it is clearly documented.

Ok, I will work on it asap.

include/crocoddyl/core/constraints/residual.hxx
cmastalli: To pass the bound from the constraint residual to the solver, we need to perform a number of copies. These copies are from the constraint residual to the constraint manager, then from the constraint manager to the diff action model, and then from the diff action to the action model. This is very convolute and error-prone.

Additionally, we need to do this every time as we wish to enable users to update the bounds. There are many examples where we wish this.

One might share the memory between these objects, however, this is not trivial as the bounds vector for the diff action (or action) model bounds depends on a number of constraint residual bounds. Moreover, some constraint residuals might be equality or inequality, and we are not arranging them using this property.

Again, this is very convolute and painful to develop (simpler equals better). Could you motivate the need of passing this information to the solver?

include/crocoddyl/core/constraints/residual.hxx
cmastalli: In this PR, I am resizing the data inside the RK integrator. However, there are other cases that are missed here, which you can find them in the big PR. Note that I have unit-tests and working MPC (which changes ng online) using this latter PR.


Unfortunately, this PR represents a tiny part of the development and it is not possible to bring other features as they are are developed nonlineary. I will mark this conversation as resolved.

cmastalli: Just to add more details. These are the type of modifications that you can see in the other PR:
  - function to easy modify dimension in the constraint manager: https://github.com/cmastalli/crocoddyl/blob/topic/preview-v2/include/crocoddyl/core/constraints/constraint-manager.hpp#L330-L348
  - use of this function in each diff action model, e.g.,https://github.com/cmastalli/crocoddyl/blob/topic/preview-v2/include/crocoddyl/multibody/actions/contact-fwddyn.hxx#L131
  - unit-test: https://github.com/cmastalli/crocoddyl/blob/topic/preview-v2/unittest/test_constraint_manager.cpp#L377
  - use the action dimensions to resize solver data online: https://github.com/cmastalli/crocoddyl/blob/topic/preview-v2/src/core/solvers/intro.cpp#L191-L216
  - virtual function to resize data in each solver: https://github.com/cmastalli/crocoddyl/blob/topic/preview-v2/include/crocoddyl/core/solver-base.hpp#L132

This last point also resolved our issue with different nu dimensions.

include/crocoddyl/core/constraints/residual.hxx
nmansard: The solver would benefit from knowing that there are some double bounds, and would handle them more efficiently than considering two simple bounds. 

There is no complexity at all in handling that. You just need to consider double bounds as one of the possible option for the high-level action-model API. Again, there are several ways of doing it: either allow double bounds by default with infinite values, or with additional flags to activate one side of the double bounds, or store double bounds apart from single bounds. 

At the end, you are storing this information in the specialized (pinocchio) part of your front end. So the point here is simply to consider it at the low level.

include/crocoddyl/core/constraints/residual.hxx
cmastalli: I included these changes in the PR as they are an important part of the current API.

include/crocoddyl/core/constraints/residual.hxx
cmastalli: Solved

include/crocoddyl/core/constraints/residual.hxx
cmastalli: Done

include/crocoddyl/core/constraints/residual.hxx
cmastalli: Done

include/crocoddyl/core/constraints/residual.hxx
cmastalli: Defined the bounds in the front-end

cmastalli:
@nmansard and @wxmerkt -- I have pushed the changes discussed in today's meeting. Concretely, I modified the inequality front end. Now, we have the lower and upper bounds in the action model and we are not tightening into the constraint vector.

Similarly to the ongoing discussion regarding the `*ng_`, I have to define a virtual function for getting the bounds. 
 This is needed to get them from the constraint manager or action model. For a quick review, you can see these lines:
  - action: https://github.com/cmastalli/crocoddyl/blob/topic/constraint-API/include/crocoddyl/core/action-base.hpp#L216
  - diff action: https://github.com/cmastalli/crocoddyl/blob/topic/constraint-API/include/crocoddyl/core/diff-action-base.hpp#L240
  - integrated action implementation: https://github.com/cmastalli/crocoddyl/blob/topic/constraint-API/include/crocoddyl/core/integ-action-base.hxx#L79-L82
  - returning the bound from the constraint manager: https://github.com/cmastalli/crocoddyl/blob/topic/constraint-API/include/crocoddyl/multibody/actions/contact-fwddyn.hxx#L298-L305

After listening to different points of view, I believe one simple yet effective design is to remove `*ng` using the same pattern described above. Any quick feedback about it? I could really implement it on Tuesday and move forward with this PR.

cmastalli:
As described above, I removed the use of raw pointer for nh and ng.

I decided to move ahead with the PR as it  incorporates all the discussed aspects.

PR #1093, opened 2022-10-12
Finite difference based computation of Lxx and Luu inside NumdiffDifferentialActionModel

spykspeigel:
Hello,
I have added the computation of Lxx and Luu using finite differences. I see this is important for solving problems using DDP on numerical derivatives information than analytical derivatives.  

Below is a small comment on the formulas used.

First order derivatives
     1)   $L_x[i] = \frac{L(x+ \delta x) - L(x)}{ \delta x}$                   # this formula has higher order terms of $O(\delta x)$
     2)   $L_x[i] = \frac{L(x+ \delta x) - L(x- \delta x)} { \delta x}$      # this formula has higher order terms of $O(\delta x^2)$

Second Order Derivatives
     1)  $L_{xx}[i,i] = \frac{Lx(x_i+\delta x_i)-2*Lx(x_i)+Lx(x_i-\delta x_i))}{\delta x_i ^2}$           # Diagonal terms
     2)  $L_{xx} [i,j] = \frac{L_x(x_i+\delta x_i, x_j+\delta x_j) - L_x(x_i+\delta x_i, x_j) - L_x(x_i, x_j+\delta x_j) + L_x(x_i, x_j) }{\delta x_i *\delta x_j} $             
#this formula has a higher order term of order $O(\delta x)$

One can write a similar formula for the finite difference for second-order derivatives with $O(\delta x^2)$     
2)  $L_{xx} [i,j] = \frac{L_x(x_i+\delta x_i, x_j+\delta x_j) - L_x(x_i+\delta x_i, x_j - \delta x_j) - L_x(x_i - \delta x_i, x_j+\delta x_j) + L_x(x_i - \delta x_i, x_j - \delta x_j) }{\delta x_i *\delta x_j} $
 
But from a computation time perspective, the formula for the second-order derivative with $O(\delta x)$ is more efficient as we can reuse 2 terms.

I look forward to your comments.

Regards,
Saroj


include/crocoddyl/core/numdiff/diff-action.hxx
```diff
@@ -83,13 +83,55 @@ void DifferentialActionModelNumDiffTpl<Scalar>::calcDiff(const boost::shared_ptr
...
+    
+    // The finite difference of the partial derivatives 
+    //
+    //  First order derivatives
+    //  Lx[i] = (L(x+ \delta x) - L(x))/disturbance                   # this formula has a higher order terms of O(\delta x)
+    //  Lx[i] = (L(x+ \delta x) - L(x- \delta x)) / disturbance       # this formula has a higher order terms of O(\delta x^2)
+
+    //  Lxx[i,i] = (Lx(x_i+\delta x_i)-2*Lx(x_i)+Lx(x_i-\delta x_i))/disturbance**2
+    //  Lxx[i,j] = (Lx(x_i+\delta x_i, x_j+\delta x_j) - Lx(x_i+\delta x_i, x_j) - Lx(x_i, x_j+\delta x_j) + Lx(x_i, x_j)) /  (delta x_i *delta x_j)      #this formula has a higher order terms of O(\delta x)
+    //  One can write a similar formula for the finite difference with O(delta x**2) but from computation time perspective the above formula is more efficient as we can reuse 2 terms
```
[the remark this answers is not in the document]
cmastalli: Btw, why are you writing 'Lx' ...?

include/crocoddyl/core/numdiff/diff-action.hxx
```diff
@@ -83,13 +83,55 @@ void DifferentialActionModelNumDiffTpl<Scalar>::calcDiff(const boost::shared_ptr
...
+      std::cout<<"Here is the value of dx :"<<d->dx(jx)<<"\n";
+
+      d->dx(ix) = 0.0;
+      model_->get_state()->integrate(x, d->dx, d->xp); 
+      model_->calc(d->data_x[ix], d->xp, u);
+      const Scalar c_zp = d->data_x[ix]->cost;  //cost due to zero disturance in 'i' and positive disturbance in 'j' direction
+      std::cout<<"Here is the value of dx :"<<d->dx(jx)<<"\n";
+      std::cout<<"Here is the value of c,  c_pp, c_zp :"<<cp << c_pp << c_zp<<"\n";
+    
+      data->Lxx(ix,jx) = (c_pp - c_zp - cp + c0)/ (disturbance_*disturbance_);
```
cmastalli, line 126: To follow the comment, we can write a more readable code as
```c++
data->Lxx(ix,jx) = (c_pp - cp - c_zp + c0)/ (disturbance_*disturbance_);
```

Right?

include/crocoddyl/core/numdiff/diff-action.hxx
```diff
@@ -99,22 +141,55 @@ void DifferentialActionModelNumDiffTpl<Scalar>::calcDiff(const boost::shared_ptr
...
-    const Scalar c = d->data_u[iu]->cost;
+    const Scalar cp = d->data_u[iu]->cost;
     data->Fu.col(iu) = (xn - xn0) / disturbance_;
 
-    data->Lu(iu) = (c - c0) / disturbance_;
+    data->Lu(iu) = (cp - c0) / disturbance_;
     d->Ru.col(iu) = (d->data_u[iu]->r - d->data_0->r) / disturbance_;
+
+    // We can apply the same formulas for finite difference as above
+
```
cmastalli, line 151: Add the comment as in the action case. Same for the doxygen code inside the hpp file.

include/crocoddyl/core/numdiff/diff-action.hxx
```diff
@@ -99,22 +141,55 @@ void DifferentialActionModelNumDiffTpl<Scalar>::calcDiff(const boost::shared_ptr
...
+    // d->du(iu) = -disturbance_; 
+    model_->calc(d->data_u[iu], d->xp, u- d->du);
+    data->Luu(iu,iu) = (d->data_u[iu]->cost + cp - 2*c0)/ (disturbance_*disturbance_);    
+    std::cout<<"Here is the value of iu :"<<iu<<"\n";
+
+    for (std::size_t ju = iu+1; ju < model_->get_nu(); ++ju) {
+      
+      std::cout<<"Here is the value of ju :"<<ju<<"\n";
```
cmastalli, line 159: Same here. Remove these comments used for debugging.

include/crocoddyl/core/numdiff/diff-action.hxx
```diff
@@ -99,22 +141,55 @@ void DifferentialActionModelNumDiffTpl<Scalar>::calcDiff(const boost::shared_ptr
...
     data->Lxu = d->Rx.transpose() * d->Ru;
-    data->Luu = d->Ru.transpose() * d->Ru;
+    // data->Luu = d->Ru.transpose() * d->Ru;
   } else {
-    data->Lxx.setZero();
+    // data->Lxx.setZero();
     data->Lxu.setZero();
-    data->Luu.setZero();
+    // data->Luu.setZero();
   }
```
cmastalli, line 193: It seems you forget to remove these lines. 

As in the action case, I would remove the Rx, Ru computations and also deprecate the "with_gauss_approx" argument in the constructor.

cmastalli (changes requested):
@spykspeigel this is a very useful PR. Thanks a lot!

There are a few minor comments that I left. Here I am going to add other points:
  - Update the unit tests to assert the code. To so, you need to remove these lines: https://github.com/loco-3d/crocoddyl/blob/devel/unittest/test_actions.cpp#L121-L123, https://github.com/loco-3d/crocoddyl/blob/devel/unittest/test_actions.cpp#L103-L107, https://github.com/loco-3d/crocoddyl/blob/devel/unittest/test_diff_actions.cpp#L139-L143,  https://github.com/loco-3d/crocoddyl/blob/devel/unittest/test_diff_actions.cpp#L157-L159
  - Deprecate the "with_gauss_approx". Do you know how to do it? (I can meet you to explain to you carefully)
  - Update the copyright
  - Formatted the code as explained here: https://github.com/loco-3d/crocoddyl/wiki/Developer-Guide
  - Provide logs of examples using the numdiff as a comment (arm_manipulation, quadrotor, quadruped, biped)

This last point is for curiosity, I would like to understand the effect of numdiff and also to validate the code empirically.

include/crocoddyl/core/numdiff/diff-action.hxx
```diff
@@ -83,13 +83,55 @@ void DifferentialActionModelNumDiffTpl<Scalar>::calcDiff(const boost::shared_ptr
...
+
+    // d->dx(ix) = -disturbance_;
+    model_->get_state()->integrate(x, -d->dx, d->xp); 
+    model_->calc(d->data_x[ix], d->xp, u);
+    data->Lxx(ix,ix) = (d->data_x[ix]->cost + cp - 2*c0)/ (disturbance_*disturbance_);    
+    std::cout<<"Here is the value of ix :"<<ix<<"\n";
+
+    for (std::size_t jx = ix+1; jx < state_->get_ndx(); ++jx) {
+      
+      std::cout<<"Here is the value of jx :"<<jx<<"\n";
```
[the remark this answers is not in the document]
spykspeigel: done

include/crocoddyl/core/numdiff/diff-action.hxx
```diff
@@ -83,13 +83,55 @@ void DifferentialActionModelNumDiffTpl<Scalar>::calcDiff(const boost::shared_ptr
...
+    //  Lx[i] = (L(x+ \delta x) - L(x- \delta x)) / disturbance       # this formula has a higher order terms of O(\delta x^2)
+
+    //  Lxx[i,i] = (Lx(x_i+\delta x_i)-2*Lx(x_i)+Lx(x_i-\delta x_i))/disturbance**2
+    //  Lxx[i,j] = (Lx(x_i+\delta x_i, x_j+\delta x_j) - Lx(x_i+\delta x_i, x_j) - Lx(x_i, x_j+\delta x_j) + Lx(x_i, x_j)) /  (delta x_i *delta x_j)      #this formula has a higher order terms of O(\delta x)
+    //  One can write a similar formula for the finite difference with O(delta x**2) but from computation time perspective the above formula is more efficient as we can reuse 2 terms
+
+    // d->dx(ix) = -disturbance_;
+    model_->get_state()->integrate(x, -d->dx, d->xp); 
+    model_->calc(d->data_x[ix], d->xp, u);
+    data->Lxx(ix,ix) = (d->data_x[ix]->cost + cp - 2*c0)/ (disturbance_*disturbance_);
```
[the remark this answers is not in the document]
jcarpent: Modern compilers are smart enough to understand it.

spykspeigel:
@cmastalli 
these are the log files
[arm_manipulation_numdiff.log](https://github.com/loco-3d/crocoddyl/files/9792423/arm_manipulation_numdiff.log)
[quadrotor_numdiff.log](https://github.com/loco-3d/crocoddyl/files/9792424/quadrotor_numdiff.log)
[bipedal_walk.log](https://github.com/loco-3d/crocoddyl/files/9792439/bipedal_walk.log)

cmastalli:
@spykspeigel -- could you rebase this branch and fix the issues with the CI jobs?
What else is needed to merge this PR?

spykspeigel:
Hi Carlos, Could you confirm the changes to unittest? The changes you suggested are not entirely related to Hessian computation.
I will try to run the test after that and will push the changes asap.

cmastalli:
> Hi Carlos, Could you confirm the changes to unittest? The changes you suggested are not entirely related to Hessian computation.
> I will try to run the test after that and will push the changes asap.

They are related as it is the only way to validate that the numdiff code is OK.

cmastalli:
Hi @spykspeigel,

I have made different improvements in this PR. The list of changes is summarised as follows:
  - fixed bug in computing Luu for diff action
  - added computation of Lxu for diff action
  - adopted a formal strategy to define the disturbance (it depends on the norm of the point)
  - used a formal tolerance

Then, I propagated all these changes and strategies to each numerical differentiation class and unit test. In short, there is nothing remaining in this PR. However, it would be interesting if you can run different OC problems using numdiff classes (diff-action and also action numdiffs). Can you do this and report these results here?

cmastalli (approved):

CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ SET(PROJECT_NAME crocoddyl)
 SET(PROJECT_DESCRIPTION "Contact RObot COntrol by Differential DYnamic programming Library (Crocoddyl)")
 SET(PROJECT_URL https://github.com/${PROJECT_NAMESPACE}/${PROJECT_NAME})
 
+OPTION(INSTALL_DOCUMENTATION "Generate and install the documentation" OFF)
```
cmastalli, line 9: @wxmerkt -- Focal/clang/Debug job is failing, and this seems to be the cause. Should I enable this flag here https://github.com/loco-3d/crocoddyl/blob/master/.github/workflows/industrial_ci_action.yml#L29?

wxmerkt: I don't think that is the case. The issue is likely that unittests with Debug mode just take too long to run and never output to the console. The new finite difference might allocate (I didnt check the source code but that would be my first hunch given the symptoms) or just be slow in Debug mode

cmastalli: Yes, it is much slower now as we need to do much more computations. I don't think we are allocating extra data.

I need to think about what strategy to take. Options are: 
  - deactivating the computation of Hessians in Debug mode, or
  - reducing the number of tests in Debug mode.

In any case, we are not checking the cost Hessians in the unit tests as the numerical error associated with the Gauss-Newton approximation is not possible to account for. This is a kind of pity, as I would like to check these Hessians as well.

Do you have any opinion about the difference options?

wxmerkt: We may need the former option for offline development but could guard it with a `#ifndef` to deactivate during unittesting since it's not checked? So `#ifndef SKIP_FINITE_DIFFERENCE_HESSIANS` or similar around that computation and we then define that for CI Debug builds?

cmastalli: Good point! Thanks for your feedback :)

cmastalli:
@nim65s -- You might be aware that I created a mechanism to deactivate the computation of cost Hessians via finite difference. Also, this PR introduces the possibility of computing these Hessians.

However, both DEBUG mode jobs (doc-coverage and ROS-CI/.../Debug) are not passing as they take too long. I deactivated the latter, but I need you to do the same for the former. Can you do this as soon as possible? (Thanks).

To do this, you need to define the following flag: `-DBUILD_WITH_FINITE_DIFF_HESSIANS=OFF`.

PR #1096, opened 2022-11-02
This enables the checking of logfiles in the jobs

cmastalli:


cmastalli:
As expected, the results change across different machines. This was an attempt to understand the instability.

PR #1099, opened 2022-11-04
Changed solver stopping + added function to compute equality-constraint feasibility

cmastalli:
As discussed in #1080, we need to change the solver's stopping criteria, as it doesn't evaluate properly the norm of the KKT point. This affects the log files, the stopping value changes and some might require fewer or more iterations to reach the stopping criteria.

For the DDP solver, the previous stopping criteria was $\|\|\mathbf{Qu}\|\|_2$, where the new one is $\Delta V=\mathbf{Qu}\mathbf{Quu}^{-1}\mathbf{Qu}$.

Instead, we compute the equality-constraint feasibility using the $\ell_1$ or $\ell_\infty$ norms. The default approach is the first one.


src/core/utils/callbacks.cpp
```diff
@@ -71,7 +72,8 @@ void CallbackVerbose::operator()(SolverAbstract& solver) {
   std::cout << solver.get_stop() << "  " << -solver.get_d()[1] << "  ";
   std::cout << solver.get_xreg() << "  " << solver.get_ureg() << "  ";
   std::cout << std::fixed << std::setprecision(4) << solver.get_steplength() << "  ";
-  std::cout << std::scientific << std::setprecision(precision_) << solver.get_ffeas();
+  std::cout << std::scientific << std::setprecision(precision_) << solver.get_ffeas() << "  ";
+  std::cout << std::scientific << std::setprecision(precision_) << solver.get_hfeas();
```
wxmerkt, line 76: We may want to also output the inequality infeasibility

src/core/solver-base.cpp
```diff
@@ -100,6 +102,33 @@ double SolverAbstract::computeDynamicFeasibility() {
   return tmp_feas_;
 }
 
+double SolverAbstract::computeEqualityFeasibility() {
```
wxmerkt, line 105: Perhaps we should generalise this to `computeConstraintFeasibility` to do both equality and inequality in the same go?

bindings/python/crocoddyl/utils/__init__.py
```diff
@@ -1282,7 +1282,8 @@ def solve(self, init_xs=[], init_us=[], maxiter=100, isFeasible=False, regInit=N
                     self.dV = self.tryStep(a)
                 except ArithmeticError:
                     continue
-                d1, d2 = self.expectedImprovement()
+                self.d = self.expectedImprovement()
+                d1, d2 = np.asscalar(self.d[0]), np.asscalar(self.d[1])
```
wxmerkt, line 1286: This will break with newer versions of NumPy (was fixed in https://github.com/cmastalli/crocoddyl/pull/11)

src/core/solver-base.cpp
cmastalli: No, we cannot. Typically solvers need to differentia both. I will push a function for inequality constraints.

bindings/python/crocoddyl/utils/__init__.py
cmastalli: I am planning to merge later that commit. However, I have fixed this issue for this particular line.

src/core/utils/callbacks.cpp
cmastalli: Now, we are printing the feasibility of the inequality constraints. I have to modify the log files as well.

Note that I created a function for computing the inequality feasibility.

cmastalli:
@wxmerkt -- I have included solutions for all your comments. Please let me know if there is something left.

include/crocoddyl/core/solver-base.hpp
```diff
@@ -142,6 +142,22 @@ class SolverAbstract {
    */
   double computeDynamicFeasibility();
 
+  /**
+   * @brief Compute the feasibility of the inequality constraints for the current guess
+   *
+   * The feasibility can be computed using the computed using the l-1 and l-inf norms.
```
wxmerkt, line 148:
```suggestion
   * The feasibility can be computed using the l-1 and l-inf norms.
```

include/crocoddyl/core/solver-base.hpp
```diff
@@ -142,6 +142,22 @@ class SolverAbstract {
...
+   * By default we use the l-inf norm, however, we can use the l-1 norm by doing set_inffeas(false).
+   */
+  double computeInequalityFeasibility();
+
+  /**
+   * @brief Compute the feasibility of the equality constraints for the current guess
+   *
+   * The feasibility can be computed using the computed using the l-1 and l-inf norms.
```
wxmerkt, line 156:
```suggestion
   * The feasibility can be computed using the l-1 and l-inf norms.
```

include/crocoddyl/core/solver-base.hpp
```diff
@@ -264,7 +280,19 @@ class SolverAbstract {
   double get_ffeas() const;
 
   /**
-   * @brief Return the norm used for the computing the feasibility (true for \f$\ell_\infty\f$, false for \f$\ell_1\f$)
+   * @brief Return the norm used for the computing the inequality feasibility (true for \f$\ell_\infty\f$, false for
+   * \f$\ell_1\f$)
```
wxmerkt, line 284:
```suggestion
   * @brief Return the inequality feasibility of the current guess
```

include/crocoddyl/core/solver-base.hpp
```diff
@@ -264,7 +280,19 @@ class SolverAbstract {
...
+   */
+  double get_gfeas() const;
+
+  /**
+   * @brief Return the norm used for the computing the equality feasibility (true for \f$\ell_\infty\f$, false for
+   * \f$\ell_1\f$)
```
wxmerkt, line 290:
```suggestion
   * @brief Return the equality feasibility of the current guess
```

wxmerkt (approved):
Looks great - I left some suggestions for correcting the docstrings which you can accept/commit via the GitHub UI prior to merging

wxmerkt:
Quick note on CI - the test-solvers unit tests are failing with Debug - did you check this locally in Debug mode in case some Eigen dimension is throwing?

cmastalli:
> Quick note on CI - the test-solvers unit tests are failing with Debug - did you check this locally in Debug mode in case some Eigen dimension is throwing?

I have fixed the documentation, but more importantly, I have replaced the boolean with an enum for defining the feasibility norm. This will allows us to later include other norms if we need them.

I realised that we weren't checking the dimension of the terminal constraints when computing the feasibilities. I believe this will fix the issue in the debug job.

Thanks @wxmerkt for fast reviewing this PR. I will still need to create other PRs.

PR #1111, opened 2023-01-10
CMake : add Python stubs and fix non-system IPOPT on Linux

ManifoldFR:
This PR:

* adds stubs using pybind11-stubgen from the cmake submodule
* fixes compilation against a non-system IPOPT (e.g. a conda env) by using the full path (as in [the docs for `target_link_libraries`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#overview))
* makes building with IPOPT optional (enabled by default)

In the future, any bindings will need to be made carefully as not to break the stubs parser. anything like `bp::args("self, t, model")` *will* break the stubs.


cmastalli (commented):
Thanks for supporting Python stubs, @ManifoldFR!

I have a question related to this topic. Is there a way to trigger a compilation error if the stub parser is broken? If so, why we do not define "GENERATE_PYTHON_STUBS=TRUE" by default?

In short, it will ideal to unit test this bit. I wonder if this is possible and implementable in this PR.

Finally, note that the job that checks the c++ format is not passing: https://gitlab.laas.fr/loco-3d/crocoddyl/-/jobs/183518. You can follow the instructions on how to automatically adjust the code format here: https://github.com/loco-3d/crocoddyl/wiki/Developer-Guide.

jcarpent:
@ManifoldFR This is not exactly the pybind11 support, but something rather adapted from

CMakeLists.txt
```diff
@@ -179,7 +181,7 @@ IF(UNIX)
     IF(APPLE)
       TARGET_LINK_LIBRARIES(${PROJECT_NAME} ${IPOPT_LIBRARY_DIRS}/libipopt.dylib)
     ELSE()
-      TARGET_LINK_LIBRARIES(${PROJECT_NAME} ipopt)
+      TARGET_LINK_LIBRARIES(${PROJECT_NAME} ${IPOPT_LIBRARY_DIRS}/libipopt.so)
```
jcarpent, line 184: Normally, we should be platform independent for the linkage. This is not a good fix.

ManifoldFR: I thought it was a bit iffy. Maybe we should be using `find_library()` to get the full path to the lib directly?

jcarpent: No, you need to add the LIBRARY_DIRS to the target via target_link_directories and just link to `ipopt`.

ManifoldFR:
> Thanks for supporting Python stubs, @ManifoldFR!
> 
> I have a question related to this topic. Is there a way to trigger a compilation error if the stub parser is broken? If so, why we do not define "GENERATE_PYTHON_STUBS=TRUE" by default?
>
> In short, it will ideal to unit test this bit. I wonder if this is possible and implementable in this PR.
 
We could switch it on to ON by default indeed, and the make command will return 1 when stubgen errors out (with something like an IndexError) and fail the CI itself.

> Finally, note that the job that checks the c++ format is not passing: https://gitlab.laas.fr/loco-3d/crocoddyl/-/jobs/183518. You can follow the instructions on how to automatically adjust the code format here: https://github.com/loco-3d/crocoddyl/wiki/Developer-Guide.

Ah, I thought it would have been caught by the pre-commit, but it seems the current config doesn't run clang-format yet.

cmastalli:
> We could switch it on to ON by default indeed, and the make command will return 1 when stubgen errors out (with something like an IndexError) and fail the CI itself.

@nim65s and @wxmerkt what do you suggest here? Should we enable by default this?

ManifoldFR:
The CI pipeline failed for an odd reason:

```  
  [build] Found '1' packages in 0.0 seconds.
  Starting  >>> catkin_tools_prebuild               
  Finished  <<< catkin_tools_prebuild                [ 2.1 seconds ]
  Starting  >>> crocoddyl                           
  
  Errors     << crocoddyl:cmake /root/target_ws/logs/crocoddyl/build.cmake.000.log
  CMake Warning at /root/target_ws/src/crocoddyl/cmake/doxygen.cmake:494 (message):
    Failed to find Doxygen, documentation will not be generated.
  Call Stack (most recent call first):
    cmake/post-project.cmake:59 (_setup_project_documentation)
    CMakeLists.txt:60 (PROJECT)
```

cmastalli:
> > > We could switch it on to ON by default indeed, and the make command will return 1 when stubgen errors out (with something like an IndexError) and fail the CI itself.
> > 
> > 
> > @nim65s and @wxmerkt what do you suggest here? Should we enable by default this?
> 
> Following my experience with this feature I have added, it should be OFF by default like (eigenpy, hpp-fcl, Pinocchio, etc.)

Ok, let's do it like that @ManifoldFR. We can always revisit this later

ManifoldFR:
> The issue is that `target_link_directories` is not available with the CMake shipped by default on Ubuntu 20.04, cf.
> 
> ```
>   CMake Error at /root/target_ws/src/crocoddyl/CMakeLists.txt:181 (TARGET_LINK_DIRECTORIES):
>     Unknown CMake command "TARGET_LINK_DIRECTORIES".
> ```

We can do `LINK_DIRECTORIES` and it should be fine I think.

jcarpent (approved):

cmastalli:
Thanks, @ManifoldFR and everyone!!!

PR #1112, opened 2023-01-13
Fix unittest factory for future compatibility

ManifoldFR:
The main change is as follows:

*unittest : use GeometryObject's CollisionGeometryPtr typedef* (4f6f3e4aa42fd1c93a19ccd06100bac60cf62d12)

Which makes the factory code independent on whether installed release of Pinocchio (or hpp-fcl) uses std or boost shared_ptr.

### DIFF
--- a/bindings/python/crocoddyl/fwd.hpp
+++ b/bindings/python/crocoddyl/fwd.hpp
@@ -10,9 +10,6 @@
 #define BINDINGS_PYTHON_CROCODDYL_CORE_FWD_HPP_
 
 #include <eigenpy/eigenpy.hpp>
-#include <boost/python.hpp>
-#include <boost/python/enum.hpp>
-#include <boost/python/scope.hpp>
 
 namespace crocoddyl {
 namespace python {
--- a/unittest/factory/residual.cpp
+++ b/unittest/factory/residual.cpp
@@ -86,10 +86,10 @@ boost::shared_ptr<crocoddyl::ResidualModelAbstract> ResidualModelFactory::create
   pinocchio::GeomIndex ig_frame = geometry->addGeometryObject(
       pinocchio::GeometryObject("frame", frame_index, state->get_pinocchio()->frames[frame_index].parent,
                                 CollisionGeometryPtr(new hpp::fcl::Sphere(0)), frame_SE3));
-  pinocchio::GeomIndex ig_obs = geometry->addGeometryObject(pinocchio::GeometryObject(
-      "obs", state->get_pinocchio()->getFrameId("universe"),
-      state->get_pinocchio()->frames[state->get_pinocchio()->getFrameId("universe")].parent,
-      CollisionGeometryPtr(new hpp::fcl::Sphere(0)), frame_SE3_obstacle));
+  pinocchio::GeomIndex ig_obs = geometry->addGeometryObject(
+      pinocchio::GeometryObject("obs", state->get_pinocchio()->getFrameId("universe"),
+                                state->get_pinocchio()->frames[state->get_pinocchio()->getFrameId("universe")].parent,
+                                CollisionGeometryPtr(new hpp::fcl::Sphere(0)), frame_SE3_obstacle));
   geometry->addCollisionPair(pinocchio::CollisionPair(ig_frame, ig_obs));
 #endif  // PINOCCHIO_WITH_HPP_FCL
   if (nu == std::numeric_limits<std::size_t>::max()) {
--- a/unittest/factory/residual.cpp
+++ b/unittest/factory/residual.cpp
@@ -85,11 +85,11 @@ boost::shared_ptr<crocoddyl::ResidualModelAbstract> ResidualModelFactory::create
       boost::make_shared<pinocchio::GeometryModel>(pinocchio::GeometryModel());
   pinocchio::GeomIndex ig_frame = geometry->addGeometryObject(
       pinocchio::GeometryObject("frame", frame_index, state->get_pinocchio()->frames[frame_index].parent,
-                                boost::shared_ptr<hpp::fcl::CollisionGeometry>(new hpp::fcl::Sphere(0)), frame_SE3));
+                                CollisionGeometryPtr(new hpp::fcl::Sphere(0)), frame_SE3));
   pinocchio::GeomIndex ig_obs = geometry->addGeometryObject(pinocchio::GeometryObject(
       "obs", state->get_pinocchio()->getFrameId("universe"),
       state->get_pinocchio()->frames[state->get_pinocchio()->getFrameId("universe")].parent,
-      boost::shared_ptr<hpp::fcl::CollisionGeometry>(new hpp::fcl::Sphere(0)), frame_SE3_obstacle));
+      CollisionGeometryPtr(new hpp::fcl::Sphere(0)), frame_SE3_obstacle));
   geometry->addCollisionPair(pinocchio::CollisionPair(ig_frame, ig_obs));
 #endif  // PINOCCHIO_WITH_HPP_FCL
   if (nu == std::numeric_limits<std::size_t>::max()) {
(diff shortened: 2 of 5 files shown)


jcarpent (approved):

wxmerkt:
CI is failing. Does this need a bump in the Pinocchio version?

```
/root/target_ws/src/crocoddyl/unittest/factory/cost.cpp:215:63: error: no matching constructor for initialization of 'pinocchio::GeometryObject'
    pinocchio::GeomIndex ig_frame = geometry->addGeometryObject(pinocchio::GeometryObject(
                                                                ^
  /opt/ros/noetic/include/pinocchio/multibody/fcl.hpp:181:3: note: candidate constructor not viable: no known conversion from 'pinocchio::SE3' (aka 'SE3Tpl<double, 0>') to 'const pinocchio::GeometryObject::CollisionGeometryPtr' (aka 'const shared_ptr<fcl::CollisionGeometry>') for 4th argument
    GeometryObject(const std::string & name,
    ^
  /opt/ros/noetic/include/pinocchio/multibody/fcl.hpp:222:3: note: candidate constructor not viable: no known conversion from 'pinocchio::FrameTpl::JointIndex' (aka 'unsigned long') to 'const pinocchio::GeometryObject::CollisionGeometryPtr' (aka 'const shared_ptr<fcl::CollisionGeometry>') for 3rd argument
    GeometryObject(const std::string & name,
```

unittest/factory/residual.cpp
```diff
@@ -85,11 +85,11 @@ boost::shared_ptr<crocoddyl::ResidualModelAbstract> ResidualModelFactory::create
       boost::make_shared<pinocchio::GeometryModel>(pinocchio::GeometryModel());
   pinocchio::GeomIndex ig_frame = geometry->addGeometryObject(
       pinocchio::GeometryObject("frame", frame_index, state->get_pinocchio()->frames[frame_index].parent,
-                                boost::shared_ptr<hpp::fcl::CollisionGeometry>(new hpp::fcl::Sphere(0)), frame_SE3));
+                                CollisionGeometryPtr(new hpp::fcl::Sphere(0)), frame_SE3));
```
cmastalli, line 88: Perhaps, we should use this code with the correct versions of Pinocchio and HPP-FCL. The idea is to still support old versions of both packages.

bindings/python/crocoddyl/fwd.hpp
```diff
@@ -10,9 +10,6 @@
 #define BINDINGS_PYTHON_CROCODDYL_CORE_FWD_HPP_
 
 #include <eigenpy/eigenpy.hpp>
-#include <boost/python.hpp>
-#include <boost/python/enum.hpp>
-#include <boost/python/scope.hpp>
```
cmastalli, line 15: Is there a particular reason why we do this? It seems this might trigger compilation issues.

unittest/factory/residual.hpp
```diff
@@ -10,7 +10,6 @@
 #define CROCODDYL_RESIDUAL_FACTORY_HPP_
 
 #include "state.hpp"
-#include "residual.hpp"
```
cmastalli, line 13: This also seems to be removed by mistake. The header includes our factory for creating different residual objects.

ManifoldFR: Isn't that the same file it's being included in?

bindings/python/crocoddyl/fwd.hpp
ManifoldFR: From what I can tell, `boost/python.hpp` has included both `enum.hpp` and `scope.hpp` [as far back as 1.60](https://www.boost.org/doc/libs/1_60_0/boost/python.hpp), so there's not need to include them.

Meanwhile, eigenpy's `fwd.hpp` includes `boost/python.hpp` itself [as far back as tag 2.7.10](https://github.com/stack-of-tasks/eigenpy/blob/v2.7.10/include/eigenpy/fwd.hpp) which is the version required in the CMakeLists file, so there's no need to include it.
 
I think in the absolute eigenpy should handle including these headers itself.

ManifoldFR:
> CI is failing. Does this need a bump in the Pinocchio version?
> 
> ```
> /root/target_ws/src/crocoddyl/unittest/factory/cost.cpp:215:63: error: no matching constructor for initialization of 'pinocchio::GeometryObject'
>     pinocchio::GeomIndex ig_frame = geometry->addGeometryObject(pinocchio::GeometryObject(
>                                                                 ^
>   /opt/ros/noetic/include/pinocchio/multibody/fcl.hpp:181:3: note: candidate constructor not viable: no known conversion from 'pinocchio::SE3' (aka 'SE3Tpl<double, 0>') to 'const pinocchio::GeometryObject::CollisionGeometryPtr' (aka 'const shared_ptr<fcl::CollisionGeometry>') for 4th argument
>     GeometryObject(const std::string & name,
>     ^
>   /opt/ros/noetic/include/pinocchio/multibody/fcl.hpp:222:3: note: candidate constructor not viable: no known conversion from 'pinocchio::FrameTpl::JointIndex' (aka 'unsigned long') to 'const pinocchio::GeometryObject::CollisionGeometryPtr' (aka 'const shared_ptr<fcl::CollisionGeometry>') for 3rd argument
>     GeometryObject(const std::string & name,
> ```

Hm, it seems like the version of Pinocchio in the CI was older than I thought. This is due to a switch in the order of arguments of the ctor (the old-order ctor will be deprecated). I'll change it back to old order.

unittest/factory/residual.cpp
ManifoldFR: The [current master branch](https://github.com/stack-of-tasks/pinocchio/blob/483713fff9531c0628ce3de6a5578368865976ac/src/multibody/fcl.hpp#L141) of Pinocchio defines the typedef. The shared_ptr type (boost or std) there is decided [from a compile-time definition](https://github.com/stack-of-tasks/pinocchio/blob/483713fff9531c0628ce3de6a5578368865976ac/src/multibody/fcl.hpp#L67) now.

From what I can see it also exists [in tag 2.6.3](https://github.com/stack-of-tasks/pinocchio/blob/5daf5de1c158363e835d322435bab4aa2f9b9b0b/src/multibody/fcl.hpp#L122) which is the minimum required version in the CMakeLists.

unittest/factory/residual.hpp
cmastalli: Yes, you're right! Thanks for cleaning up!

bindings/python/crocoddyl/fwd.hpp
cmastalli: Ok, thanks for the explanation.

ManifoldFR:
Last thing is fixing the formatting. I remember that @cmastalli linked me to a `clang-format` command on the wiki which mentioned clang-format 6.0 or something. Applying that command (I'm on clang-15) changed the formatting of all the files on my end which is strange.

Which style format are we supposed to use here ? Google, llvm ?

cmastalli:
> Last thing is fixing the formatting. I remember that @cmastalli linked me to a `clang-format` command on the wiki which mentioned clang-format 6.0 or something. Applying that command (I'm on clang-15) changed the formatting of all the files on my end which is strange.
> 
> Which style format are we supposed to use here ? Google, llvm ? I'll update the wiki if needed

It shouldn't change all the documents. There is clearly something wrong.

You might need to install clang-6, but not entirely sure. Have you put the configuration file as described in the wiki? If so, have you done this before running the clang command?

Note that you put the config file after running for the first time, then clang doesn't consider anymore that config file for some weird reason. When this happens, I clone again the repo.

unittest/factory/residual.cpp
cmastalli: I have another question. Have you checked that the Python bindings work well?

I know that boost python was not properly binding 'std::shared_ptr'. I also know that boost python developers were planning to sort out this situation.

jcarpent: Boost Python works perfectly with both std and boost shared_ptr.

jcarpent: This is what we are using in Pinocchio 3x and more.

cmastalli: Thanks for the confirmation!

cmastalli:
Thanks, @ManifoldFR!!

PR #1115, opened 2023-01-30
Fix size bug in free-fwddyn 

skleff1994:
Fix for what seems to be inconsistent matrix sizes in free-fwddyn : `Minv` is `nv x nv` while `u_drift` and `dtau_dx` have `nu` rows.


cmastalli (approved):
Thanks, @skleff1994!

PR #1117, opened 2023-02-08
cmake : use cmakemodules' external FindOpenMP.cmake

ManifoldFR:
This PR:

* makes it use the jrl-cmakemodules' [FindOpenMP.cmake](https://github.com/jrl-umi3218/jrl-cmakemodules/tree/master/find-external/OpenMP) file which fixes the OpenMP_CXX target
* uses ADD_COMPILE_OPTIONS to add the required compiler flags
* makes the root cmake file use the OpenMP_CXX_FLAGS variable to add openmp flags
* use the OpenMP::OpenMP_CXX target in the unittest cmakelists

The rationale is that when compiling against a conda environment where an OpenMP is installed, the system OpenMP is somehow found instead

### DIFF
--- a/benchmark/bipedal-timings.cpp
+++ b/benchmark/bipedal-timings.cpp
@@ -1,7 +1,8 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (C) 2019-2021, LAAS-CNRS, CTU, INRIA, University of Edinburgh
+// Copyright (C) 2019-2023, LAAS-CNRS, CTU, INRIA, University of Edinburgh,
+//                          Heriot-Watt University
 // Copyright note valid unless otherwise stated in individual files.
 // All rights reserved.
 ///////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -126,11 +133,7 @@ IF(BUILD_WITH_MULTITHREADS)
   FIND_PACKAGE(OpenMP)
 ENDIF()
 IF(OPENMP_FOUND AND BUILD_WITH_MULTITHREADS)
```
jcarpent, line 135:
```suggestion
IF(BUILD_WITH_MULTITHREADS)
```

CMakeLists.txt
```diff
@@ -126,11 +133,7 @@ IF(BUILD_WITH_MULTITHREADS)
   FIND_PACKAGE(OpenMP)
```
jcarpent, line 133:
```suggestion
  FIND_PACKAGE(OpenMP REQUIRED)
```

unittest/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ IF(UNIX)
   TARGET_LINK_LIBRARIES(${PROJECT_NAME}_unittest ${PROJECT_NAME} example-robot-data::example-robot-data)
 
   IF(OPENMP_FOUND)
```
jcarpent, line 15:
```suggestion
  IF(BUILD_WITH_MULTITHREADS)
```

CMakeLists.txt
```diff
@@ -123,14 +130,10 @@ ENDIF()
...
+  FIND_PACKAGE(OpenMP REQUIRED)
 ENDIF()
-IF(OPENMP_FOUND AND BUILD_WITH_MULTITHREADS)
-  IF(APPLE)
-    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Xpreprocessor -fopenmp")
-  ELSE()
-    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp")
-  ENDIF(APPLE)
+IF(BUILD_WITH_MULTITHREADS)
+  ADD_COMPILE_OPTIONS(${PROJECT_NAME} ${OpenMP_CXX_FLAGS})
```
cmastalli, line 136: The cmake submodule is defining `-Xpreprocessor` for all the cases. I know this flag is needed for OSX but not for Linux.

Have been tested this compilation flag in Ubuntu and Windows?
If this works, what is the effect in Ubuntu and Windows?

CMakeLists.txt
```diff
@@ -18,6 +18,13 @@ INCLUDE(cmake/base.cmake)
 INCLUDE(cmake/boost.cmake)
 INCLUDE(cmake/apple.cmake)
 INCLUDE(cmake/stubs.cmake)
+IF(APPLE) # Use the handmade approach
+  SET(CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake/find-external/OpenMP ${CMAKE_MODULE_PATH})
+ELSEIF(UNIX)
+  IF(${CMAKE_VERSION} VERSION_GREATER "3.20.0" OR ${CMAKE_VERSION} VERSION_EQUAL "3.20.0")
+    set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake/find-external/OpenMP ${CMAKE_MODULE_PATH})
+  ENDIF()
```
cmastalli, line 26: Why are we handling this here instead of the cmake submodule?

Also, it seems we aren't including OpenMP when cmake is lower than 3.20. Why are we doing so?

Is there something that I am missing?

PR #1118, opened 2023-02-16
Adding missing header for std::set

Sergim96:
This PR adds a missing header for the usage of std::set in the constraint manager. The lack of the header was responsible for compilation errors in some machines.

### DIFF
--- a/include/crocoddyl/core/constraints/constraint-manager.hpp
+++ b/include/crocoddyl/core/constraints/constraint-manager.hpp
@@ -12,6 +12,7 @@
 #include <string>
 #include <map>
 #include <utility>
+#include <set>
 
 #include "crocoddyl/core/fwd.hpp"
 #include "crocoddyl/core/constraint-base.hpp"


cmastalli (approved):
Thanks, @Sergim96!

cmastalli:
@nim65s it seems one of the jobs is using std::shared pointers for FCL geometry.
Can you advise us on what to do?

nim65s:
for those types, we shouldn't explicit std or boost, but rely on exposed types by hpp-fcl or pinocchio

cmastalli:
> for those types, we shouldn't explicit std or boost, but rely on exposed types by hpp-fcl or pinocchio

`std::set` is not supported by either dependency. At least, I couldn't see any use of it in those repositories.

Additionally, we are already including `set` in other parts of Crocoddyl, e.g., https://github.com/loco-3d/crocoddyl/blob/devel/include/crocoddyl/core/costs/cost-sum.hpp#L15, which didn't trigger any issue in past. Would this triggers an issue now? If so, it should be a deprecation procedure developed in the above mentioned dependencies; this might affect many other packages.

@Sergim96 could you test this PR with a docker? @nim65s could you share these dockers with him?

Given all these details, @nim65s what is your suggestion for sorting out this PR?

wxmerkt:
I believe it's about not hard-coding the return type to be `boost::shared_ptr` but use the Pinocchio specified ptr-type as for `CollisionGeometryPtr`:
image

I.e. in `cost.cpp` try the `CollisionGeometryPtr` type as the return object. Or use `auto`

cmastalli:
Closing this issue due to inactivity.

PR #1121, opened 2023-03-04
Enabled copy and deepcopy in all objects

cmastalli:
This PR allows us to do the following:
```python
import crocoddyl
import copy

m = crocoddyl.ActionModelUnicycle()
d = m.createData()

m1 = copy.copy(m)
m2 = copy.deepcopy(m)

d1 = copy.copy(d)
d2 = copy.deepcopy(d)
```


unittest/bindings/test_costs.py
```diff
@@ -260,19 +260,19 @@ class FrameVelocityCostSumTest(CostModelSumTestCase):
...
-    loader = unittest.TestLoader()
-    suites_list = []
-    for test_class in test_classes_to_run:
-        suite = loader.loadTestsFromTestCase(test_class)
-        suites_list.append(suite)
-    big_suite = unittest.TestSuite(suites_list)
-    runner = unittest.TextTestRunner()
-    results = runner.run(big_suite)
-    sys.exit(not results.wasSuccessful())
+# test to be run
```
wxmerkt, line 263: What is the reason behind removing the `if __name__ == '__main__'` from these tests?

cmastalli: I created cmake target commands to be able to run specific python unit tests. But if I run them with `if __name__ == '__main__'`, then they don't trigger failures in the tests.

nim65s: this is because you are not running the python files, but importing them.

cmastalli: @nim65s:

Is bad to remove the `if __name__ ...`?
If so, could you help me with a working solution?

unittest/bindings/CMakeLists.txt
```diff
@@ -8,8 +8,13 @@ SET(${PROJECT_NAME}_PYTHON_BINDINGS_TESTS
...
+  copy
   )
 
 FOREACH(TEST ${${PROJECT_NAME}_PYTHON_BINDINGS_TESTS})
-  ADD_PYTHON_UNIT_TEST("pybinds-${TEST}" "unittest/bindings/test_${TEST}.py" "bindings/python")
+  PYTHON_BUILD(. "test_${TEST}.py")
+  ADD_PYTHON_UNIT_TEST("test_pybinds_${TEST}" "unittest/bindings/test_${TEST}.py" bindings/python)
+  ADD_CUSTOM_TARGET("test_pybinds_${TEST}"
+    ${CMAKE_COMMAND} -E env PYTHONPATH=${PROJECT_BINARY_DIR}/bindings/python:$ENV{PYTHONPATH}
+    ${PYTHON_EXECUTABLE} -c "import test_${TEST}")
```
nim65s, line 19:
```suggestion
    ${PYTHON_EXECUTABLE}  "${CMAKE_CURRENT_SOURCE_DIR}/test_${TEST}.py"))
```

cmastalli: Done, thanks!

unittest/bindings/test_costs.py
wxmerkt: It would be better to keep it if we can make the test-targets work

wxmerkt:
Thank you for the swift action. No further comments, this looks great and is a significant improvement for users deriving/prototyping new models

PR #1127, opened 2023-04-18
Using eigenpy's StdVectorPythonVisitor

ManifoldFR:
Following PR stack-of-tasks/eigenpy#325, we moved `StdVectorPythonVisitor` from Pinocchio upstream to Eigenpy, the reasoning being to centralize utilities upstream into Eigenpy.

Since the code in `bindings/python/utils/vector-converter` is derived from Pinocchio's own implementation, I'm thinking it could be worth it to move Crocoddyl's bindings codebase to use the new upstream implementation (the changes are rather minimal).


bindings/python/crocoddyl/utils/map-converter.hpp
```diff
@@ -27,7 +27,7 @@ namespace bp = boost::python;
  * \sa Pickle
  */
 template <typename Container>
-struct PickleMap : public PickleVector<Container> {
+struct PickleMap : public eigenpy::PickleVector<Container> {
```
cmastalli, line 30: I am not familiar with the latest `Eigenpy` development. However, it might be worth moving the `PickleMap` to it. Do you agree or not with it? If not, I would appreciate the reasoning.

cmastalli (approved):
Thanks, @ManifoldFR!

I agree with moving this code to `Eigenpy`.Perhaps, we could also consider the `PickleMap` in this or a future PR.

cmastalli:
@ManifoldFR  -- it seems we need to update the `eigenpy` version to 2.9.0 here: https://github.com/loco-3d/crocoddyl/blob/devel/CMakeLists.txt#L74

ManifoldFR:
> @ManifoldFR  -- it seems we need to update the `eigenpy` version to 2.9.0 here: https://github.com/loco-3d/crocoddyl/blob/devel/CMakeLists.txt#L74 

Okay, will do!

ManifoldFR:
@cmastalli it seems we would just need 2.8.0 for this feature, according to the [releases page](https://github.com/stack-of-tasks/eigenpy/releases/tag/v2.8.0)

jcarpent (approved):
Great uniformization @ManifoldFR. Thanks a lot for this PR.

PR #1134, opened 2023-05-19
Fix build on aarch64-linux

wegank:
In nixpkgs, we encountered [the following error](https://logs.ofborg.org/?key=nixos/nixpkgs.232585&attempt_id=a548c3a7-18fe-4568-a125-8c820509a67d) on aarch64-linux when bumping to 2.0.0:

```bash
/build/source/src/core/solvers/intro.cpp:196:19: error: no matching function for call to 'max(double&, int)'
  196 |   stop_ = std::max(hfeas_, abs(d_[0] + 0.5 * d_[1]));
```

The obvious fix is to replace `abs` with `std::abs`.

### DIFF
--- a/src/core/solvers/ddp.cpp
+++ b/src/core/solvers/ddp.cpp
@@ -95,7 +95,7 @@ bool SolverDDP::solve(const std::vector<Eigen::VectorXd>& init_xs,
       dVexp_ = steplength_ * (d_[0] + 0.5 * steplength_ * d_[1]);
 
       if (dVexp_ >= 0) {  // descend direction
-        if (abs(d_[0]) < th_grad_ || !is_feasible_ ||
+        if (std::abs(d_[0]) < th_grad_ || !is_feasible_ ||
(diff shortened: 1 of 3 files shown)


hrp2-14:
Hi ! This project doesn't usually accept pull requests on the main branch.
If this wasn't intentionnal, you can change the base branch of this PR to devel
(No need to close it for that). Best, a bot.

jcarpent (approved):

nim65s:
Hi @wegank,

Thanks for this PR and your packaging of crocoddyl (and its dependencies) in nix. I already gave it a try a year ago or so, but got distracted on the way, so I'm very glad you started it.

FYI, I'll make a crocoddyl v2.0.1 soon, with 
1. this PR
2. some fixes following example-robot-data v4.0.7
3. a couple reverts to get everything working on 18.04, as @cmastalli needs it for now, 

so maybe you can wait for this instead of providing a patched v2.0.0 on nix, and then pushing v2.0.1 a couple days later. But that's as you prefer :)

PR #1138, opened 2023-05-25
cmake: use find-external FindOpenMP.cmake

ManifoldFR:
This PR changes a couple of CMake configuration options (I needed the changes to compile against a conda environment. We should wait for the CI to tell us if it breaks on non-conda setups).

* use find-external FindOpenMP.cmake
* include the pinocchio/fwd.hpp header in bindings multibody/frames.cpp (this might be required under pin3x or something)
* ~use `add_project_dependency` for OpenMP instead of `find_package`, for other CMake projects consuming crocoddyl (e.g. sobec)~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,10 +70,13 @@ check_minimal_cxx_standard(14 ENFORCE)
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/find-external/CppAD/"
                       ${CMAKE_MODULE_PATH})
 if(APPLE) # Use the handmade approach
-  set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake/find-external/OpenMP ${CMAKE_MODULE_PATH})
+  set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake/find-external/OpenMP
+                        ${CMAKE_MODULE_PATH})
 elseif(UNIX)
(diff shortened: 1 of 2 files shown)


nim65s:
Could you document your current CMake version and the error message you have to require this change ?

cmastalli:
@nim65s -- Before proceeding with this PR, can we return the 18.04 support? 

(It would be important to check that there is no issue there).

wxmerkt:
Somehow this breaks support on regular Ubuntu 20.04 with standard tooling as with the Focal test. Is the FindOpenMP newer?

ManifoldFR:
Huh, weird. Maybe we should add a check for the CMake version like what's
inside of Pinocchio.

jcarpent (approved):

jmirabel:
Hi,
The `FindOpenMP.cmake` shipped inside the CMake submodule does not work for CMake 2.22. It uses argument `SOURCE_FROM_VAR` of `try_compile` which, if I understand the doc correctly was added in 3.25.

@ManifoldFR not sure what you tested but since [CMake 3.22 doc](https://cmake.org/cmake/help/v3.22/command/try_compile.html#try-compiling-source-files) does not mention `SOURCE_FROM_VAR`, I would assume it is not supported and hence would not use the internal `FindOpenMP.cmake`.

ManifoldFR:
Since this PR is from nearly two years ago, I really don't remember what I tried anymore.

Mind you the use of `SOURCE_FROM_VAR` was added in this commit (https://github.com/jrl-umi3218/jrl-cmakemodules/commit/9887cf83b33ea6af621930ba0a41fd0a3be1d302) on the cmake modules, which happened a month after this PR was merged.

jmirabel:
Thank you for your answer.

jmirabel:
In #1334, I propose to increase the version to 3.25.

PR #1140, opened 2023-05-31
Add Ubuntu 18.04 support back

nim65s:
HI,

As asked by @cmastalli for some not-yet-upgraded-robots, support is extended for
- CMake 3.10
- Boost 1.64
- Python 3.6

While here, fixing git-blame-ignore-revs & sorting pre-commit


cmastalli (approved):
Thanks, @nim65s!

Is possible to activate the 18.04 job? (I guess this is not critical but helpful if we can get it)

nim65s:
Do you mean on gitlab-ci ?
No: it was a python 2.7 job, so it won't work.

But adding one on github should be easy, I'll try to add this soon, if nobody handle this first.

jcarpent:
I think you cannot use ubuntu-18.04 on github actions (https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/)

nim65s:
True, it might not be as easy as I thought, but I guess we can probably use `ubuntu:18.04` containers inside a `ubuntu-latest` runner

PR #1151, opened 2023-07-30
Fixex cppadcg building + code maintenance

cmastalli:
This PR contains a few improvements. Here I list them:
 - Fixed compilation issues when building with codegen support
 - Removed deprecated frame classes (finally after 3 years)
 - Simplified cmake code for Crocoddyl library compilation
 - Moved Python factory code from utils folder to unittest/bindings
 - Updated log files
 - Simplified copyable visitor code
 - Changed verbose printed message to be compliance with pre-commit.ci


bindings/python/crocoddyl/utils/copyable.hpp
```diff
@@ -56,12 +24,25 @@ struct CopyableVisitor : public bp::def_visitor<CopyableVisitor<C> > {
...
     cl.def("copy", &copy, bp::arg("self"), "Returns a copy of *this.");
-    cl.def("__copy__", &generic__copy__<C>);
-    cl.def("__deepcopy__", &generic__deepcopy__<C>);
+    cl.def("__copy__", &copy, bp::arg("self"), "Returns a copy of *this.");
+    cl.def("__deepcopy__", &deepcopy, bp::args("self", "memo"),
+           "Returns a deep copy of *this.");
+    cl.def("copy", &copy_ptr, bp::arg("self"), "Returns a copy of *this.");
+    cl.def("__copy__", &copy_ptr, bp::arg("self"), "Returns a copy of *this.");
+    cl.def("__deepcopy__", &deepcopy_ptr, bp::args("self", "memo"),
+           "Returns a deep copy of *this.");
```
jcarpent, line 33:
```suggestion
    cl.def("__copy__", &copy, bp::arg("self"), "Returns a copy of *this.");
    cl.def("__deepcopy__", &deepcopy, bp::args("self", "memo"),
           "Returns a deep copy of *this.");
```

bindings/python/crocoddyl/utils/copyable.hpp
```diff
@@ -56,12 +24,25 @@ struct CopyableVisitor : public bp::def_visitor<CopyableVisitor<C> > {
...
 
  private:
   static C copy(const C& self) { return C(self); }
+  static C deepcopy(const C& self, bp::dict) { return C(self); }
+  static boost::shared_ptr<C> copy_ptr(const boost::shared_ptr<C>& self) {
+    return boost::shared_ptr<C>(self);
+  }
+  static boost::shared_ptr<C> deepcopy_ptr(const boost::shared_ptr<C>& self,
+                                           bp::dict) {
+    return boost::shared_ptr<C>(self);
```
jcarpent, line 44: I don't think these signatures are required.

cmastalli: I need to enable copy and deepcopy in shared pointers as well. Without this code, it was failing test_copy.py.

I would say that this is good thing to propagate in other repos. I can handle this if you agree with doing so.

bindings/python/crocoddyl/utils/copyable.hpp
cmastalli: Closing this thread as it belongs to discussion in https://github.com/loco-3d/crocoddyl/pull/1151#discussion_r1278619301.

bindings/python/crocoddyl/utils/copyable.hpp
cmastalli: After investigating deeply this point, I realised that @jcarpent is right. We don't need these signatures.

Crocoddyl unit tests were failing due to other reasons:
 1. Base classes are non-copyable by definition.
 2. Base data structures were defined as non-copyable, but they should be copyable.
 3. Unit tests were not checking ID for copy and deepcopy.

Using shared pointers was a fake remedy as the code could run but the ID between copied objects remain the same. This requires addressing 3 as well.

Thanks @jcarpent for pointing out this issue. I have removed the signatures as suggested while sorting out 1, 2, and 3 points.

PR #1156, opened 2023-08-09
Enabled conda jobs with CppADCodeGen

cmastalli:


cmastalli:
@nim65s -- I am enabling CppADCodeGen via Conda and GitHub Actions. However, the ubuntu-latest jobs are taking between 3.5 and 4.5 hrs. What is more convenient in your experience? Removing the ubuntu-latest jobs or being patient. Do you have other ideas for improving this?

Additionally, is there any work from you on deploying GitHub action runners? I imagine the answer is no. But it is worth confirming it.

nim65s:
2h of `make test` in Debug:
> Total Test time (real) = 7237.05 sec

I guess this is the first optimization path :)

Issue #1162, opened 2023-08-29
Problems trying to switch from default datatypes to floats


will-gerard:
Thanks for all your work creating and open sourcing such a great trajopt library. I have gone through the tutorials and have been experimenting with it, and have been finding it to work very smoothly. I am able to get a DDP solve working correctly for a Kuka arm, both in python via the bindings and in the c++ directly. However I also want to test with floats, rather than with doubles, and when I try to switch to the templated versions of the functions and pass float as the template parameter, I am running into issues. I saw from [this](https://github.com/stack-of-tasks/pinocchio/issues/2037) github issue that for parsing a URDF only doubles are supported, but that they can then be cast to floats, which is what I am doing for the robot model I create, and this part seems to work:

```
ModelTpl<double> model_double;
pinocchio::urdf::buildModel(urdf_filename, model_double);
pinocchio::ModelTpl<float> model = model_double.cast<float>();
```
However, I am unable to create a state multibody object using the float robot model, and I am not sure why. What I was doing before, and which works correctly, is:

```
boost::shared_ptr<crocoddyl::StateMultibody> state = 
    boost::make_shared<crocoddyl::StateMultibody>(
	    boost::make_shared<pinocchio::ModelTpl<double>>(robot_model));
```

And so what I want to do now is something like:

```
boost::shared_ptr<crocoddyl::StateMultibodyTpl<float>> state =
    boost::make_shared<crocoddyl::StateMultibodyTpl<float>>(
        boost::make_shared<pinocchio::ModelTpl<float>>(model));
```

But this results in the following errors:

```
/home/a2rlab/anaconda3/envs/crocoddyl/lib/pkgconfig/../../include/crocoddyl/multibody/states/multibody.hxx:64:6:   required from here
/home/a2rlab/anaconda3/envs/crocoddyl/lib/pkgconfig/../../include/pinocchio/multibody/liegroup/special-orthogonal.hpp:74:45: error: no matching function for call to ‘if_then_else(pinocchio::internal::ComparisonOperators, const Scalar&, double, float, pinocchio::internal::if_then_else_impl<float, float, float, float>::ReturnType)’
   74 |                                 if_then_else(internal::GT, tr, Scalar(2) - 1e-2,
      |                                 ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   75 |                                              asin((R(1,0) - R(0,1)) / Scalar(2)), // then
[...]

PR #1165, opened 2023-09-01
Missed Scalar in cost sum and activation data

cmastalli:


cmastalli:
@will-gerard -- I don't think this will fix your issue, but it is worth trying this PR. Could you try it please?

will-gerard:
Hi @cmastalli, this did in fact fix my issue! After building crocoddyl from the devel branch, including this commit, I am no longer hitting the error I reported [here](https://github.com/loco-3d/crocoddyl/issues/1162). I will note that previously I had compiled from the master branch, and so I have more updates now than just this change, but from the devel branch I am able to run 

```
			boost::shared_ptr<crocoddyl::ResidualModelStateTpl<float>> terminal_state_residual =
				boost::make_shared<crocoddyl::ResidualModelStateTpl<float>>(state, x_ref_t);
```

the line which was giving me issues before, successfully. I will test using the devel branch from now on. Thank you for the fix!

PR #1194, opened 2023-12-06
Fixed broken bench + updated ROS CI

cmastalli:
This PR fixes broken bench examples and update the ROS CI jobs.


cmastalli:
@wxmerkt -- do you know what is happening with ROS Noetic jobs?

.github/workflows/ros1-ci.yaml
```diff
@@ -29,20 +25,17 @@ jobs:
       # target after completion of the regular test target. The output of this step does affect the output of the CI process.
       # Note, this does not affect projects that do not have pure CMake projects in their upstream_ws.
       BUILDER: catkin_tools
-      AFTER_RUN_TARGET_TEST: 'source /root/target_ws/install/setup.bash && cd /root/target_ws/build/crocoddyl && env && make test'
```
wxmerkt, line 32: This together with the continue-on-failure will most likely deactivate running the tests at all due to the way the jrl-cmakemodules interact with catkin_tools. I didnt get to see the PR until now

cmastalli: I noticed we were running twice the tests. This is the reason why I removed this.

I also defined fail-fast as False. It doesn't look like "continuous-on-error" is stopping running other jobs' tests. I will check carefully this again.

In any case, the noetic Python test are failing. It looks like it cannot find the Python bindings.

wxmerkt: Thank you for the detail - I can take a look next week - I assume something in the tooling changed so it's no longer picking things up the way it should

cmastalli: Thanks @wxmerkt! As you might know this happening in Pinocchio

PR #1195, opened 2023-12-06
Fix stub generation

jorisv:
Stub generation is broken :

- The stub generator is launched in the wrong directory
- The stub generator don't depend of the binding and then can be launched before


jorisv:
Before merge, I must check if pyc generation dependency is well defined

jorisv:
> Before merge, I must check if pyc generation dependency is well defined

done

cmastalli (approved):
Thanks for fixing this, @jorisv!

I will proceed with merging this after sorting out the issue with our CI. Please bear with me.

PR #1200, opened 2023-12-12
Set Python library output

cmastalli:


nim65s (commented):
Could you develop what you are trying to do here ?

At first glance, you are adding 3 lines of CMake code just to get the default behaviour

cmastalli:
> Could you develop what you are trying to do here ?
> 
> At first glance, you are adding 3 lines of CMake code just to get the default behaviour

Hi @nim65s 

I should have marked this as a draft. I am clueless regarding the issue with Noetic jobs. I wanted to check this!

nim65s:
We also have issues with noetic on pinocchio, so this might be related

PR #1228, opened 2024-03-14
Generated Python stubs in Conda CI

cmastalli:


cmastalli:
Hi @nim65s

I'll proceed with merging this, but it seems the gitlab-ci format conflicts with the new changes proposed by pre-commit. One alternative is to remove this job as I believe we don't need it anymore. However, I'm happy with any action you decide.

nim65s:
I'll fix it, but yes, we can ignore it for now

PR #1249, opened 2024-05-14
CMake: update

nim65s:


cmastalli (commented):
Hi, @nim65s!

Let's ensure that macOS jobs pass before merging. Thanks!

cmastalli (approved):

PR #1253, opened 2024-05-28
Pinocchio v3 compatibility

nim65s:


unittest/bindings/factory.py
```diff
@@ -747,9 +747,16 @@ def calc(self, data, x, u=None):
...
+        # The old code was looking like this.
+        # But, the std::vector<Eigen::MatrixXd> returned by Jdiff is destroyed
+        # before the assignment.
+        # To avoid this issue, we store the std::vector in a variable.
+        # data.residual.Rx[:] = self.state.Jdiff(
+        #     self.xref, x, crocoddyl.Jcomponent.second
+        # )[0]
+
+        diff = self.state.Jdiff(self.xref, x, crocoddyl.Jcomponent.second)
+        data.residual.Rx[:] = diff[0]
```
cmastalli, line 759: Could you avoid breaking our Python API?

I am afraid this will create a lot of stress in many projects based on Crocoddyl.

This seems to be related to changes in eigenpy as you can see here: https://github.com/loco-3d/crocoddyl/blob/devel/bindings/python/crocoddyl/utils/vector-converter.hpp#L12

cmastalli (changes requested):
Thanks for adding compatibility with Pinocchio 3, @nim65s!

Let us sort out the breaking of Python API. The problem might arise from recent changes in eigenpy.

nim65s:
As you can see in the commit history, @jorisv is doing most of the work, and deserve the credits :)
I just opened this so that we can more easily talk about those changes for now.

unittest/bindings/factory.py
jorisv: Hello, I don't think the API is broken here (I just store the Jdiff return in a temporary variable).

But this fix hide something really weird. We should investigate if some change in pinocchio 3 std::vector binding can impact the crocoddyl std::vector management.

nim65s:
NB: there is another blocking points than what we can see here. They are being worked on, in private spaces. @LudovicDeMatteis will be able to provide a green light when everything will be ready enough for him.

unittest/bindings/factory.py
cmastalli: Thanks for investigating this. But if I remember correctly, Crocoddyl is using eigenpy std::vector bindings.

nim65s: I guess this is non blocking then.

nim65s: ref #1255

PR #1271, opened 2024-06-22
Replaced parent by parentJoint introduced in Pinocchio 3

cmastalli:


cmastalli:
@nim65s -- it seems gitlab-ci is not yet installing Pinocchio 3. Should I proceed to ignore and merge this PR?

PR #1296, opened 2024-09-04
Nix

nim65s:
Same as https://github.com/stack-of-tasks/eigenpy/pull/498

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,19 @@
+name: "CI - Nix"
+
+on:
+  push:
+
+jobs:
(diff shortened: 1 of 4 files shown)


cmastalli (commented):
Thanks for supporting Nix, @nim65s!

Could you include the Nix installation instructions and badget in the Readme file?

nim65s:
Done. While here, I'm also documenting the archlinux package.

cmastalli (approved):
Brilliant, thanks!

Please be aware that ROS jobs are not passing due to Pinocchio, i.e., no need to wait for them.

PR #1307, opened 2024-09-23
Add Bounds to the Inequality Feasibility Calculation

aristotpap:
This pull request introduces a refinement to the calculation of the inequality feasibility by adding the inequality bounds to the calculation. The update ensures that when a constraint is within its specified bounds, its feasibility is considered 0. Only if the constraint exceeds the bounds, its feasibility is greater than zero.

### DIFF
--- a/src/core/solver-base.cpp
+++ b/src/core/solver-base.cpp
@@ -115,33 +115,47 @@ double SolverAbstract::computeDynamicFeasibility() {
 
 double SolverAbstract::computeInequalityFeasibility() {
   tmp_feas_ = 0.;
+  Eigen::VectorXd g_adj;
+
   const std::size_t T = problem_->get_T();
   const std::vector<boost::shared_ptr<ActionModelAbstract> >& models =
       problem_->get_runningModels();
   const std::vector<boost::shared_ptr<ActionDataAbstract> >& datas =
       problem_->get_runningDatas();
+
   switch (feasnorm_) {
     case LInf:
       for (std::size_t t = 0; t < T; ++t) {
         if (models[t]->get_ng() > 0) {
-          tmp_feas_ =
-              std::max(tmp_feas_, datas[t]->g.lpNorm<Eigen::Infinity>());
+          g_adj = datas[t]
+                      ->g.cwiseMax(models[t]->get_g_lb())
+                      .cwiseMin(models[t]->get_g_ub());
+          tmp_feas_ = std::max(tmp_feas_,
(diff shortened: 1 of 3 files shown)


src/core/solver-base.cpp
```diff
@@ -115,33 +115,47 @@ double SolverAbstract::computeDynamicFeasibility() {
 
 double SolverAbstract::computeInequalityFeasibility() {
   tmp_feas_ = 0.;
+  Eigen::VectorXd g_adj;
```
cmastalli, line 118: 1. Declare this object as a private member of the SolverAbstract class.
2. Rename it to g_adj_

cmastalli (changes requested):
Thanks, @aristotpap!

I have left a minor detail to be addressed.

cmastalli:
@aristotpap -- Also, do not forget to update our changelog.

aristotpap:
Hi @cmastalli,

Thank you for the suggestions. I can definitely create a private member for `g_adj`, however we need to consider whether we want or not dynamic allocation during the calculations. A situation like this may arise if certain action models have different numbers of constraints. In these situations we have two alternatives as I see it:

1. Dynamically resize the vector like below:
```c
g_adj_.resize(models[t]->ng).setZero()
```

2. Find the maximum constraint dimension during the construction of the Shooting Problem and use part of the pre-allocated vector for the feasibility calculations, as below:
```c
g_adj_.resize(max_ng).setZero() # During the Construction of the Shooting Problem
...
...
g_adj_.head(models[t]->ng) # Use during the calculations with models[t] -> ng <= max_ng
```

Let me know what you prefer according to Crocoddyl's requirements.

cmastalli:
I don't like the idea of using a "max dimension". It seems to me that the best implementation is when we create a list of g_adj_ items and resize them when needed.

We have defined a `resizeData` function that is called every time the shooting problem is changed. Our MPC highly depends on it. For more details see: https://github.com/loco-3d/crocoddyl/blob/devel/src/core/solver-base.cpp#L68-L79.

Note that `resizeData` is a virtual function that can be specialized for any solver.

aristotpap:
Hello @cmastalli,

I have implemented the requested changes and updated the `CHANGELOG.md` in the [most recent commit](https://github.com/loco-3d/crocoddyl/pull/1307/commits/80654e1ed10c9f3662c1cba14661137308472bec). Let me know if there is anything else that requires my attention.

cmastalli (approved):
Perfect, thanks!

PR #1310, opened 2024-09-26
[actuation] fixed thrust direction for calculation of thrust jacobian

sugikazu75:
I fixed the direction of thrust which is required for calculation of thrust jacobian. 
When the orientation of thruster is not identical to that of parent link, this patch is required.

Since rotational moment generated by thrust is $\lambda p \times u$, where $\lambda$, $p$, and $u$ are thrust, position of thruster, and direction of thrust, respectively.


cmastalli (approved):

cmastalli:
Thanks for the fix, @sugikazu75 !

PR #1315, opened 2024-10-08
fix ipython notebooks

nim65s:


.pre-commit-config.yaml
```diff
@@ -23,6 +23,7 @@ repos:
   - id: clang-format
     args:
     - --style=Google
+    exclude: '^.*\.ipynb$'
```
cmastalli, line 26: Is there any reason why to not pre-commit notebooks?

nim65s: our notebooks are python code, not C++ code, so there is no reason for clang-format to try format them.
Also, try to open one from the current devel: they are broken.
This break is even detected by the `check-json¿` hook from pre-commit/pre-commit-hooks.

PR #1327, opened 2024-12-11
Fix bug in quadruped.py and biped.py

Sergim96:
Hi @cmastalli,

This PR fixes a bug in both scripts quadruped.py and biped.py

The current implementation adds an impulse for all the foot_ids that are already in contact, not to the ones that are gaining a contact.

I also updated the logs


cmastalli (approved):
Brilliant, thanks!

PR #1339, opened 2025-01-29
{boost -> std}::shared_ptr

nim65s:
Hi,

I think it is time to move away from boost::shared_ptr.

@cmastalli: do you agree ? This would be a breaking change, and the next release would be v3.0.0.

We'll also have left https://github.com/stack-of-tasks/pinocchio/issues/2563 to fix before that release.

### DIFF
--- a/benchmark/all_robots.cpp
+++ b/benchmark/all_robots.cpp
@@ -34,7 +34,7 @@ void print_benchmark(RobotEENames robot) {
   unsigned int T = 1e3;  // number of trials
 
   // Building the running and terminal models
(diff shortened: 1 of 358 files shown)


cmastalli:
Thanks for doing this, @nim65s!

I was never against it. However, some time ago, I had issues with binding `std::shared_ptr`. Are you sure they are sorted out with the latest Boost Python?

I would definitely try this myself before approving. But, it would be helpful if you confirm/check this as well.

cmastalli (approved):
I have checked this myself and am happy to merge it!

PR #1348, opened 2025-02-28
Force contacts to update force derivatives

Sergim96:
Hi @cmastalli this PR fixes a bug in the computation of the derivatives of the contact forces.
This bug can be observed by adding a disabled contact to the `DifferentialActionModelContactInvDynamics` class: 
```python
contactModel = crocoddyl.ContactModelMultiple(state, nu)
supportContactModel = crocoddyl.ContactModel6D(
    state,
    target_id,
    pinocchio.SE3.Identity(),
    pinocchio.LOCAL,
    nu,
    np.array([0, 0]),
)
contactModel.addContact("contact", supportContactModel, False)
runningModel = crocoddyl.IntegratedActionModelEuler(
    crocoddyl.DifferentialActionModelContactInvDynamics(
        state, actuation, contactModel, runningCosts), dt
)
```
when running `calc` and `calcDiff` it can be observed that the jacobian of the constraint `Hu` is empty as `df_du` is only updated when the contact is active.

The proposed fix, termporally sets all the contacts to active before calling `updateForceDiff` and then restores the original status

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+* Force DifferentialActionDataContactInvDynamics to update df_du for all contacts in  https://github.com/loco-3d/crocoddyl/pull/1348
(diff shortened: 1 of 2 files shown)


cmastalli (approved):

cmastalli:
@Sergim96  -- please update the changelog!

PR #1351, opened 2025-03-04
Fix bug in intro LU decomposition

Sergim96:
Hi @cmastalli, this PR fixes a bug in the current way of computing the LU decomposition.

This bug can be observed when trying to compute the kernel and the range space of this jacobian:
```c++
    MatrixXd Hu(6, 12);
    Hu << 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1;
```
The current implementation `YZ_[t] << Hu_lu_[t].matrixLU().transpose(), Hu_lu_[t].kernel();` produces the same range space as the kernel, thus resulting in a zero matrix when multiplied by the constraint jacobian.

The correct implementation should be `YZ_[t] << (Hu_lu_[t].permutationP() * data->Hu).transpose(), Hu_lu_[t].kernel();`

as you can see there are minor changes in the logs


src/core/solvers/intro.cpp
```diff
@@ -245,7 +245,8 @@ double SolverIntro::calcDiff() {
         const std::shared_ptr<crocoddyl::ActionDataAbstract>& data = datas[t];
         if (model->get_nu() > 0 && model->get_nh() > 0) {
           Hu_lu_[t].compute(data->Hu);
-          YZ_[t] << Hu_lu_[t].matrixLU().transpose(), Hu_lu_[t].kernel();
+          YZ_[t] << (Hu_lu_[t].permutationP() * data->Hu).transpose(),
+              Hu_lu_[t].kernel();
```
cmastalli, line 249: This implementation might allocate dynamic data in some compilers. Let's rewrite it as follows:
```suggestion
          YZ_[t].leftCols(Hu_rank_[t]).noalias() = (Hu_lu_[t].permutationP() * data->Hu).transpose();
          YZ_[t].rightCols(nu - Hu_rank_[t]) = Hu_lu_[t].kernel();
```

cmastalli (changes requested):
Thanks, @Sergim96 !

I suggested an improvement before merging this PR.

Sergim96:
done! I also squashed the commits

cmastalli:
> done! I also squashed the commits

Brilliant, thanks!

src/core/solvers/intro.cpp
```diff
@@ -245,7 +245,10 @@ double SolverIntro::calcDiff() {
         const std::shared_ptr<crocoddyl::ActionDataAbstract>& data = datas[t];
         if (model->get_nu() > 0 && model->get_nh() > 0) {
           Hu_lu_[t].compute(data->Hu);
-          YZ_[t] << Hu_lu_[t].matrixLU().transpose(), Hu_lu_[t].kernel();
+          YZ_[t].leftCols(Hu_rank_[t]).noalias() =
+              (Hu_lu_[t].permutationP() * data->Hu).transpose();
+          YZ_[t].rightCols(model->get_nu() - Hu_rank_[t]) = Hu_lu_[t].kernel();
+          Hu_lu_[t].kernel();
           Hu_rank_[t] = Hu_lu_[t].rank();
```
cmastalli, line 252: Hi @Sergim96 !

I forgot to mention, but we need to update `Hu_rank_[t]` first. I wonder how this is working

Sergim96: I did re-generated the logs when I updated the code
I did it again just in case and the logs do not change

cmastalli (approved):

PR #1353, opened 2025-03-07
Remove absolute path for boost library

abussy-aldebaran:
I encountered the same problem as in pinocchio (https://github.com/stack-of-tasks/pinocchio/issues/2187).
I propose to fix it with the same fix https://github.com/stack-of-tasks/pinocchio/pull/2202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+* Removed absolute path for boost library in https://github.com/loco-3d/crocoddyl/pull/1353
 * Fixed checking of positive semi-define condition in LQR problems in https://github.com/loco-3d/crocoddyl/pull/1352
 * Compute LU decomposition using permutationP in https://github.com/loco-3d/crocoddyl/pull/1351
 * Force DifferentialActionDataContactInvDynamics to update df_du for all contacts in  https://github.com/loco-3d/crocoddyl/pull/1348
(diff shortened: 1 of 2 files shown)


cmastalli:
Thanks @abussy-aldebaran ! Please update our changelog.

nim65s:
I think this is lacking include dirs.

I also think nobody cares about pkg config, and maybe it is already broken and nobody noticed, or this PR will break things and nobody will notice.

In short, as I already said in https://github.com/stack-of-tasks/pinocchio/issues/2525#issuecomment-2556553549, I think we shoud simply remove this. And this is already what I did for distribution on pip, also because of absolute paths : https://github.com/cmake-wheel/crocoddyl/commit/ff5f81639e4badf9a294043f1d5311b2b6e0b9e4

cmastalli:
> I think this is lacking include dirs.
> 
> I also think nobody cares about pkg config, and maybe it is already broken and nobody noticed, or this PR will break things and nobody will notice.
> 
> In short, as I already said in [stack-of-tasks/pinocchio#2525 (comment)](https://github.com/stack-of-tasks/pinocchio/issues/2525#issuecomment-2556553549), I think we shoud simply remove this. And this is already what I did for distribution on pip, also because of absolute paths : [cmake-wheel@ff5f816](https://github.com/cmake-wheel/crocoddyl/commit/ff5f81639e4badf9a294043f1d5311b2b6e0b9e4)

Hi @nim65s -- are you suggesting to follow your cmake-wheel commit or complete removing?

nim65s:
Short term, we can merge this PR if it help @abussy-aldebaran
Longer term, we need to work on that pinocchio issue and propagate the result

cmastalli:
> Short term, we can merge this PR if it help @abussy-aldebaran Longer term, we need to work on that pinocchio issue and propagate the result

I fully agree with this plan. Anyway, thanks for raising this point during this PR.

cmastalli (approved):

abussy-aldebaran:
Thanks for merging !

PR #1362, opened 2025-03-19
Topic/virtual binding constraint dimension

Sergim96:
Hi @cmastalli this PR set the constraint dimension getters as virtual functions for the python bindings.
Additionally, it deprecates the setters as the user should not be able to modify these values (they should come from the constructor)


cmastalli (changes requested):
Just a minor improvement is requested. Ping me when it's done. Thanks ;)

cmastalli (approved):

PR #1364, opened 2025-03-20
Fixing ROS CI in devel branch

cmastalli:


cmastalli:
@nim65s -- do you know why our ROS jobs are not passing? Can this be related with a increased compilation time?

nim65s:
Yes, ROS & Nix are both killed by OoM errors

PR #1366, opened 2025-03-24
Add explicit template instantiation

abussy-aldebaran:
I tried to do something similar to what's done in pinocchio.

Pros : 
- Fine grain selection of function to instantiate
- Not need to move around templated member functions (e.g. `cast`)
- Already proven in pinocchio
- Handles Windows export
- Necessary to export free functions

Cons : 
- Lot of boilerplate and macros

Another solution would be to instantiate the whole classes, but I'm not sure how it will work with dynamic linkage. I believe it may be ok on linux, but not on windows, which you don't support yet, but plan to.

TODO:

- [ ] installation of `config.hpp`
- [ ] naming
- [ ] formatting/coding style


abussy-aldebaran:
Note
```bash
objdump -TC libcrocoddyl.so | grep UND | grep -i croc
```
can help to check if some symbols were left undefined.

cmastalli:
Hi @abussy-aldebaran,

My idea was around instating the whole class, as this reduces the maintenance burden. I am not an expert in Windows, but it should work, right? 

Additionally, separating each scalar into different modules (as Pinocchio does) won't allow us to cast objects in Python, e.g.,
```python
import crocoddyl
model = crocoddyl.ActionModelLQR(2, 2)
model_float = model.cast(crocoddyl.DTypes.Float32)
```
Do you agree with me?

I have a WIP development that instantiates whole classes. You can have a look at: https://github.com/cmastalli/crocoddyl/tree/topic/wip-eti.

Anyone else would like to express his/her opinion like @nim65s ?

abussy-aldebaran:
I agree about not splitting `float` and `double` into separate modules. And exporting whole classes is fine by me.

As a reference, here is microsoft documentation about `dllimport/dllexport`:
https://learn.microsoft.com/en-us/cpp/cpp/using-dllimport-and-dllexport-in-cpp-classes?view=msvc-170
Especially the part
> Inheritance and Exportable Classes
All base classes of an exportable class must be exportable. **If not, a compiler warning is generated**. Moreover, all accessible members that are also classes must be exportable. This rule permits a dllexport class to inherit from a dllimport class, and a dllimport class to inherit from a dllexport class (though the latter is not recommended). As a rule, everything that is accessible to the DLL's client (according to C++ access rules) should be part of the exportable interface. This includes private data members referenced in inline functions.

It might come handy to turn this warning anto an error.

You could also split `.cpp` files for each scalar. I made a dummy example of an implementation using symlinks to avoid too much duplication :
https://github.com/abussy-aldebaran/dummy/commit/ae6d81f143d3b427577f005c24bdfca2127d4386
This will split the load to build "ETI" files. But this is probably an early optimization.

cmastalli:
> I agree about not splitting `float` and `double` into separate modules. And exporting whole classes is fine by me.

Nice, you agree with me. I am currently working on a solution based on exporting whole classes. We can change this in the future if we realize it is not the best approach.



> You could also split `.cpp` files for each scalar. I made a dummy example of an implementation using symlinks to avoid too much duplication :
> [abussy-aldebaran/dummy@ae6d81f](https://github.com/abussy-aldebaran/dummy/commit/ae6d81f143d3b427577f005c24bdfca2127d4386)
> This will split the load to build "ETI" files. But this is probably an early optimization.

This is a good idea. Currently, I have improved this idea further. Concretely, I generated `cpp` files for each scalar from `cpp.in` files. This avoids creating many `cpp` files manually. This idea can also be extended to Python bindings. However, I might not commit to this now.

cmastalli:
> It looks like you've made good progress ! Do you need help ?

No for now. However, I would love to hear your feedback when I create a PR.

abussy-aldebaran:
> No for now. However, I would love to hear your feedback when I create a PR.

Feel free to add me as a reviewer when you do !

I compiled your branch and I noticed a lot of this warning:
```bash
warning: type attributes ignored after type is already defined [-Wattributes]
```
Comparing with pinocchio, you have the wrong `DLLAPI`. It should be
```c++
extern template class CROCODDYL_EXPLICIT_INSTANTIATION_DECLARATION_DLLAPI
    crocoddyl::ActionModelAbstractTpl<double>;

template class CROCODDYL_EXPLICIT_INSTANTIATION_DEFINITION_DLLAPI
    ActionModelAbstractTpl<double>;
```

cmastalli:
> I compiled your branch and I noticed a lot of this warning:

Thanks for reviewing this! I have pushed force a fix, plus other changes. Please be aware that it is still WIP, but your comments are more than welcome.

cmastalli:
I'm closing this PR as this is resolved in https://github.com/loco-3d/crocoddyl/pull/1367.

PR #1367, opened 2025-03-25
Introducing explicit template instantiation

cmastalli:
This PR introduces template instantiation for almost all the classes in Crocoddyl. Exceptions are numdiff's classes not widely used such as state's numdiff. Additionally, it brings a mechanism to generate cpp files for each type of supported scalar.

This PR reduces substantially the compilation time and memory footprint. Early benchmarks show a 
  1. compilation time reduction of over 37% for a single core and over 62% for multiple cores compilation
  2. memory footprint reduction of over 31 %
  3. reduction of fetching memory from RAM of over 200%

These statistics are not updated, as I did when instantiating around half of what is proposed in this PR. Additionally, they were generated when using a single cpp for compiling double and floats. Therefore, the current figures should be much better, but in the interest of time, I have skipped this benchmark.

[abussy-aldebaran](https://github.com/abussy-aldebaran) -- please have a look at this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,9 @@ set(PROJECT_URL https://github.com/${PROJECT_NAMESPACE}/${PROJECT_NAME})
 
 option(INSTALL_DOCUMENTATION "Generate and install the documentation" OFF)
 
+# Set up Crocoddyl's scalar types
+set(SCALAR_TYPES float double)
+
 # Check if the submodule cmake have been initialized
 set(JRL_CMAKE_MODULES "${CMAKE_CURRENT_LIST_DIR}/cmake")
 if(EXISTS "${JRL_CMAKE_MODULES}/base.cmake")
@@ -171,6 +174,7 @@ if(BUILD_WITH_CODEGEN_SUPPORT)
         "${PACKAGE_EXTRA_MACROS}\nADD_DEFINITIONS(-DPINOCCHIO_CPPAD_REQUIRES_MATRIX_BASE_PLUGIN)"
     )
   endif()
+  list(APPEND SCALAR_TYPES CppAD::AD<CppAD::cg::CG<double>>)
 endif()
 
 # Add OpenMP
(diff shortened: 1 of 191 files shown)


cmastalli:
@nim65s -- would it be OK to impose Pinocchio v3.1.0 as the minimum version? Unfortunately, v3.4.0 is not in conda.

nim65s:
> @nim65s -- would it be OK to impose Pinocchio v3.1.0 as the minimum version?

Yes

> Unfortunately, v3.4.0 is not in conda.

Hum, I think it is: https://anaconda.org/conda-forge/pinocchio

cmastalli:
> Hum, I think it is: https://anaconda.org/conda-forge/pinocchio

Ok, I didn't see that! @traversaro or @jcarpent -- do you know what is happening with Pinocchio v3.4.0 in Conda?

cmastalli:
@nim65s and @abussy-aldebaran -- there some codegen-related fixes that I will propagate in another PR. This includes some further improvements of this PR.

cmastalli:
> I think the problem is that you are depending on the `boost` package that is deprecated/renamed since [conda-forge/boost-feedstock#164](https://github.com/conda-forge/boost-feedstock/pull/164) (unfortunately conda at the moment does not have a nice mechanism to inform downstream users that some packages have been deprecated/renamed). I think that you need to depend on `libboost-devel` (if you require boost at the C++ level) and/or `liboost-python-devel` (if you use boost-python for Python bindings). Switching to those, you should be able to get recent conda-forge packages, including pinocchio 3.4.0 .

Thanks for this insight, @traversaro. I will investigate this asap.

CMakeLists.txt
```diff
@@ -246,6 +271,46 @@ if(UNIX)
...
+    string(REPLACE "<" "_" sanitized "${sanitized}")
+    string(REPLACE ">" "" sanitized "${sanitized}")
+    string(REPLACE "," "_" sanitized "${sanitized}")
+    set(${sanitized_name}
+        "${sanitized}"
+        PARENT_SCOPE)
+  endfunction()
+  # Find all .cpp.in template files recursively
+  file(GLOB_RECURSE ETI_TEMPLATE_FILES FOLLOW_SYMLINKS
+       ${CMAKE_SOURCE_DIR}/src/**/*.cpp.in)
```
abussy-aldebaran, line 288: Maybe use `CONFIGURE_DEPENDS` ? see https://cmake.org/cmake/help/latest/command/file.html

abussy-aldebaran:
Sorry about this late review, I was out of office yesterday.

There are some files where you don't use `@SCALAR_TYPE@` or `CROCODDYL_DECLARE_EXTERN_TEMPLATE_CLASS`. I tagged them in a review I haven't submitted yet, because I don't want to pollute this PR, and you will probably include them in the upcoming PR. They might be easier to spot in this PR though.

I have a few questions, regarding factoring the code generation:

1. Is it possible to factor the declarations and the definitions in some way ?
2. Is it possible to use only one template file to generate all the `.cpp` instead of many ?
3. If you plan to support a different number of  scalar for some classes, maybe use variadic macros ?

These questions will be better answered in the upcoming PR. I'm just writing them here, so I won't forget them.
Perhaps these ideas are over-engineering.

cmastalli:
> Sorry about this late review, I was out of office yesterday.
> 
> There are some files where you don't use `@SCALAR_TYPE@` or `CROCODDYL_DECLARE_EXTERN_TEMPLATE_CLASS`. I tagged them in a review I haven't submitted yet, because I don't want to pollute this PR, and you will probably include them in the upcoming PR. They might be easier to spot in this PR though.
> 

I knew this wasn't completely right, but I was working in a large PR and I opted to wait for fixing it. A solution is provided in https://github.com/loco-3d/crocoddyl/pull/1370

> I have a few questions, regarding factoring the code generation:
> 
> 1. Is it possible to factor the declarations and the definitions in some way (e.g. the headers as well) ?

Can you explain more?

> 2. Is it possible to use only one template file to generate all the `.cpp` instead of many ?

Unfortunately, this is not possible as we need to specialize cpp.in files for classes and type of scalars we need to compile.

> 3. If you plan to support a different number of  scalar for some classes, maybe use variadic macros ?
> 

Can you explain more your idea?

> These questions will be better answered in the upcoming PR. I'm just writing them here, so I won't forget them. Perhaps these ideas are over-engineering.

Please have a look at the new PR. I must confess that I am planning to do other major refactorings in terms of Python bindings and how they are compiled.

CMakeLists.txt
cmastalli: Done in the open PR. Thanks for the feedback!

abussy-aldebaran:
I was thinking about something ilike : 
https://coliru.stacked-crooked.com/a/c23ed50e720e3e9f

PR #1370, opened 2025-04-01
Introducing CodeGen and AutoDiff in Python

cmastalli:
This PR introduces `CppADCodeGen` support in Python. Concretely, it allows us to generate code from the commodity of Python easily. The PR introduces a `05_codegenerating_a_cartpole.ipynb` notebook to demonstrate how this works. 

Additionally, we introduce a mechanism to code generate `calc` and `calcDiff` functions for terminal nodes, including the `quasiStatic` function. When code-generating our `calc` functions, we can also generate automatic derivatives. This is handy when we are not worried too much about efficiency, although this method remains competitive.

On another note, this PR also introduces precompiled headers to accelerate compilation. It finally re-structure our Jupyter notebooks to better explain how Crocoddyl works. These notebooks are still WIP.

Note that this PR fixes the reported issue in https://github.com/loco-3d/crocoddyl/pull/1367#issuecomment-2757050465 (fyi @abussy-aldebaran)


jcarpent:
I'm wondering why you have not been using our package [pycppad](https://github.com/Simple-Robotics/pycppad), which enables the seamless usage of CppAD and CppADCodeGen in Python. Pinocchio is compatible with this feature, and you can do whatever you want.

cmastalli:
We are using pycppad in our notebook to declare scalars. However, we are not using pycppad for creating the library and other inner workings.

This makes the action code generation very straightforward, as the user don't need to worry about these inner workings, which I suspect are far to be obvious, e.g., generating a library for running and another one for terminal nodes. Additionally, it is most efficient as we transfer values from CppADCodeGen to Crocoddyl's action data in C++ using Eigen. The alternative you mention will constrain us to do this using NumPy Python. Finally, there is a legacy component as we already have developed the ActionModelCodeGen class.

I am happy to hear any other feedback you have. I haven't finished this thread of work and we can shape it together.

Btw, your question is indeed very relevant and can help others to understand why Crocoddyl does this.

cmastalli:
There is another technical aspect. If we use pycppad to generate calc and calcDiff without ActionModelCodeGen, the we won't be able to run those functions in parallel. As action models written in Python deactivate this to bypass the issues with GIL.

This is because I haven't been able to find a working solution to this problem.

jcarpent:
> There is another technical aspect. If we use pycppad to generate calc and calcDiff without ActionModelCodeGen, the we won't be able to run those functions in parallel. As action models written in Python deactivate this to bypass the issues with GIL.
> 
> This is because I haven't been able to find a working solution to this problem.

Do you use parallelization over the CppAD expression or the generated code?

cmastalli:
> > There is another technical aspect. If we use pycppad to generate calc and calcDiff without ActionModelCodeGen, the we won't be able to run those functions in parallel. As action models written in Python deactivate this to bypass the issues with GIL.
> > This is because I haven't been able to find a working solution to this problem.
> 
> Do you use parallelization over the CppAD expression or the generated code?

Generated code.

PR #1372, opened 2025-04-05
Created independent Python modules associated to each scalar

cmastalli:
This PR creates independent Python libraries, allowing us to separate the compilation of different scalars per a single Python class. This leads to a reduction in compilation memory footprint and exploitation of multicore compilation. Moreover, the massive software infrastructure makes much easier the process of integrating other scalars, e.g., codegen with floats.


cmastalli:
@nim65s -- do you know why are not running the Gepetto jobs?

nim65s:
> @nim65s -- do you know why are not running the Gepetto jobs?

One worker of the server that handle this was down, no idea why. I restarted it.

PR #1343, opened 2025-02-11
Feature/improve wrench computation

rjgriffin42:

PR #1385, opened 2025-05-04
Feature/offset center

rjgriffin42:
This PR looks to allow moving the center of the wrench cone polygon in the foot. This will allow setting up the "on toes" state without having to do anything else. This could be done by changing the box length to 0 (or something small to maintain a box structure), and setting the center of the box using the new `set_center` function to half the length of the foot. 

By default the center is 0, making this change do nothing unless called.

It builds off PR https://github.com/loco-3d/crocoddyl/pull/1343.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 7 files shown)


cmastalli:
@rjgriffin42  -- There were bugs in the previous PR that I fixed. But this means that we need to rebase this PR before reviewing it. Could you rebase it and solve conflicts?

PR #1387, opened 2025-05-07
[bugfix]: base classes of DataCollectorJointActMultibodyInContact

Sergim96:
Hi @cmastalli this PR fixes a bug in the python bindings of the class `DataCollectorJointActMultibodyInContact`.
The class had the wrong base classes resulting in the `joint` attribute to not exist.


cmastalli (approved):
Brilliant, @Sergim96!

I apologise if I introduced a bug in our Python bindings.

PR #1396, opened 2025-05-30
nix: add jupyter lab app

nim65s:
Add a shortcut to run jupyter lab, eg. `nix run .#jupyter`.


cmastalli (commented):
Thanks for this, @nim65s.

As you can see, we are testing our notebooks using cmake. More details in https://github.com/loco-3d/crocoddyl/blob/devel/notebooks/CMakeLists.txt.

However, we are just running a single one for now. As I am planning to clean up and develop better notebooks. If we want to test current notebooks, I suggest relying on cmake. What are your thoughts?

nim65s:
Yes, automated CMake notebook tests is a good thing, but they can't catch everything, and sometime the best solution to reproduce a notebook issue is to spawn jupyter.

This PR is complementary to those CMake tests I think.

That can also be useful if someone want the fastest way to start jupyter on a particular crocoddyl fork/version/branch/commit.

cmastalli (approved):

cmastalli:
Fair enough. Having a second approach doesn’t hurt!

Issue #1444, opened 2025-10-22
typo in Notebook examples


Dhanushvarma:
Notebook: https://github.com/loco-3d/crocoddyl/blob/devel/notebooks/01_introduction_to_crocoddyl.ipynb

Section 2:  States in Crocoddyl

Equation 4

cmastalli:
Thanks, @Dhanushvarma! If you don't mind, it would be helpful to contribute by fixing the typo.

Dhanushvarma:
Also, I am not sure if this is a typo or intended. [here](https://github.com/loco-3d/crocoddyl/blob/c4f79e07b1b60d8e00d89fdadff8ff5f0a9bebdf/bindings/python/crocoddyl/utils/quadruped.py#L574)

since in the docstring for `createFootSwitchModel`, it says `param supportFootIds: Ids of the constrained feet`

cmastalli:
There is no typo, but I agree that the code is not intuitive to read. In any case, we will update this file in a future mayor release.

PR #1445, opened 2025-10-24
fix: jacobian difference typo

Dhanushvarma:
https://github.com/loco-3d/crocoddyl/issues/1444


cmastalli (approved):
Thanks, @Dhanushvarma !

PR #1464, opened 2025-12-19
Developing branch for articulated aerial robots

sugikazu75:


cmastalli:
Hi @sugikazu75!

There is no need for such code. If the goal is to define thrusters on the robot’s limbs, we can discuss an appropriate way to proceed. This collaboration would lead to a consistent approach with the substantial improvements we are aiming to deliver in advanced optimization, dynamics backends, and system identification/estimation capabilities.

sugikazu75:
@cmastalli 
Thanks for the comment! Also, sorry for mistakenly posting the PR and causing extra work for you to review.
As you mentioned, I'm trying to use this library to control a robot with thrusters on its rim.

This implementation allows passing thrust to ABA as fext, but it's hardcoded, so I'm looking for a better approach.

I've considered a few possibilities below, and I'd appreciate any design advice.
- Since the actuation model isn't designed to hold data, it shouldn't calculate the thruster frame Jacobian. Therefore, only compute fext (expressed in local frame) in actuation and access it from the action model.
- Construst above fext in action model.
- Implement thrust as a controllable external force. Unlike standard Contact, it lacks acceleration constraints, making it fundamentally different?

Whichever implementation we choose, I believe the Thruster class needs to be modified to hold a Frame ID, similar to Contact.

PR #1467, opened 2025-12-30
CMake: make example-robot-data more optional for ROS

### LINKED ISSUE #1466: example-robot-data breaks Crocoddyl ROS2 building
I am unaware of the details of what is happening. However, it seems that we need to create a rosdep rule for `example-robot-data`. If so, this will also require handling both installation cases: with and without the loader. 

Additionally, it might be a good idea to declare `example-robot-data` as a test_depend (i.e., `<test_depend>example-robot-data</test_depend>`).

For more details see: https://build.ros2.org/job/Rdev__crocoddyl__ubuntu_noble_amd64/5/console.

I can help with this if you guys provide guidance.

nim65s:
fix #1466

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,14 +139,24 @@ if(BUILD_PYTHON_INTERFACE)
 endif(BUILD_PYTHON_INTERFACE)
 add_project_dependency(pinocchio 3.1.0 REQUIRED PKG_CONFIG_REQUIRES
                        "pinocchio >= 3.1.0")
+
+# example-robot-data is not available in ROS and it is only used for tests,
+# examples, and benchmarks
 if(BUILD_EXAMPLES
    OR BUILD_TESTING
    OR BUILD_BENCHMARK)
-  add_project_dependency(example-robot-data 4.0.9 REQUIRED PKG_CONFIG_REQUIRES
-                         "example-robot-data >= 4.0.9")
-else()
-  add_optional_dependency(example-robot-data)
+  find_package(example-robot-data 4.0.9 QUIET)
+  if(NOT TARGET example-robot-data::example-robot-data)
+    message(STATUS "example-robot-data not found (>= 4.0.9). Let's fetch it.")
+    include(FetchContent)
+    FetchContent_Declare(
+      "example-robot-data"
+      GIT_REPOSITORY "https://github.com/gepetto/example-robot-data.git"
+      GIT_TAG v4.4.0)
+    FetchContent_MakeAvailable("example-robot-data")
+  endif()
 endif()
+
 if(BUILD_WITH_IPOPT)
   add_optional_dependency(ipopt)
 endif()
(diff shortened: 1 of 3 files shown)


package.xml
```diff
@@ -26,7 +26,7 @@
   <depend>eigenpy</depend>
   <depend>pinocchio</depend>
   <depend>coinor-libipopt-dev</depend>
-  <exec_depend>example-robot-data</exec_depend>
+  <!-- <test_depend>example_robot_data</test_depend> -->
```
cmastalli, line 29: Why is this commented?

CMakeLists.txt
```diff
@@ -139,14 +139,23 @@ if(BUILD_PYTHON_INTERFACE)
...
-                         "example-robot-data >= 4.0.9")
-else()
-  add_optional_dependency(example-robot-data)
+  find_package(example-robot-data QUIET)
+  if(NOT TARGET example-robot-data::example-robot-data)
+    message(STATUS "example-robot-data not found. Let's fetch it.")
+    include(FetchContent)
+    FetchContent_Declare(
+      "example-robot-data"
+      GIT_REPOSITORY "https://github.com/gepetto/example-robot-data.git")
```
cmastalli: Let's impose the minimal version.
```suggestion
  find_package(example-robot-data 4.0.9 QUIET)
  if(NOT TARGET example-robot-data::example-robot-data)
    message(STATUS "example-robot-data not found (>= 4.0.9). Let's fetch it.")
    include(FetchContent)
    FetchContent_Declare(
      "example-robot-data"
      GIT_REPOSITORY "https://github.com/gepetto/example-robot-data.git"
      GIT_TAG v4.0.9)
```

cmastalli:
@nim65s -- is there something blocking the merging of this PR?

CMakeLists.txt
```diff
@@ -139,14 +139,24 @@ if(BUILD_PYTHON_INTERFACE)
...
-else()
-  add_optional_dependency(example-robot-data)
+  find_package(example-robot-data 4.0.9 QUIET)
+  if(NOT TARGET example-robot-data::example-robot-data)
+    message(STATUS "example-robot-data not found (>= 4.0.9). Let's fetch it.")
+    include(FetchContent)
+    FetchContent_Declare(
+      "example-robot-data"
+      GIT_REPOSITORY "https://github.com/gepetto/example-robot-data.git"
+      GIT_TAG v4.0.9)
```
nim65s:
```suggestion
      GIT_TAG v4.4.0)
```

package.xml
nim65s: Because otherwise ROS buildfarm will refuse to build crocoddyl

nim65s:
I just forgot it

package.xml
cmastalli: Ok, thanks for the explanation. We can completely remove this IMO.

cmastalli (approved):

PR #1491, opened 2026-04-10
introduce colmpc

nim65s:
Hi,

colmpc (https://github.com/agimus-project/colmpc) implement residuals for Crocoddyl for collision avoidance for trajectory optimisation and model predictive control.

It was developped mainly by [Arthur Haffemayer](https://ahaffemayer.github.io/) during his [PhD](https://theses.hal.science/tel-05422954/) with Nicolas Mansard and Florent Lamiraux, in the frame of the Agimus project.

It include 2 contributions:
- distance: https://laas.hal.science/hal-04425002
- velocity damper: https://laas.hal.science/hal-04707324

It was developed outside of the crocoddyl repo for practical reasons (one of those being an optional dependency on mim-solvers, but if I recall correctly that was mostly built on the experience from [sobec](https://github.com/MeMory-of-MOtion/sobec) workflow), but now that it is stable, mature, and used in other projects (especially [agimus-controller](https://github.com/agimus-project/agimus_controller)), I do believe it would make sense to merge it here.

This draft PR simply import colmpc git history in a subdirectory, and add very minimal plumbing and cleaning, just as a proof of concept, to play with the idea and start discussions.

So here are the main questions:

- do we agree on the idea of merging colmpc into crocoddyl ? (I already got positive answers from @nmansard & @ahaffemayer)
- if so, technically, should we try to preserve colmpc git history inside crocoddyl repo as done in this PR, or just reference the old repo that will be archived ? (I don't have a strong opinion here)

Also, I believe that at first, keeping a first-level "colmpc" namespace (eg. in header directory and python module) would help a lot for compatibility with existing projects that currently require both crocoddyl & colmpc, but maybe later, in a second step, we'll want to homogenize the API between crocoddyl and colmpc, and for that dissolving colmpc namespace and files more tightly into crocoddyl codebase would be an option I guess.


MaximilienNaveau:
I do agree that merging colmpc into crocoddyl is a very good idea.

For the git history let's keep the current PR as it sounds good IMHO.

For the merging I would be more drastic in the merging of functionality from colmpc into crocoddyl changing the namespaces and headers directly.
For those who use both crocoddyl and colmpc they still should be able to use the archived package along crocoddyl with the added functionalities as the core API won't change (if I am not wrong).
Maybe I missed something here that would prevent one to install this PR with colmpc and make them work together.

Then again, if we want to make a small step first. I would be in favor of merging this PR as is (i.e adding the colmpc code next to crocoddyl) and then in a second step merge all colmpc API inside the crocoddyl package. Though it maybe an extra step that may not be required IMHO.

cmastalli:
@nim65s -- thanks for this PR. Give time to evaluate it. However, it might not be a good idea as we're working on a really major release to include optimal estimation.

For instance, in this new release, we won't support differentiable action models. Therefore, many of the proposed changes will be incompatible with the big release.

nim65s:
Thanks @MaximilienNaveau for your inputs.

Yes, this would cleanly break everything for people with a (crocoddyl + colmpc) package but also an off-tree colmpc one. I personally have a huge bias, as I believe packaging, dependency hell, and software distribution in general have been solved problems for decades (ref. E. Dolstra 2004), but that may still not hold for everyone. So I guess your solution is probably a better option.


And thanks @cmastalli for the review.

I was not aware of these expected breaking changes. It sounds like this would badly split the ecosystem between crocoddyl v3 users and crocoddyl v4 users. I mean, even without considering the idea of merging colmpc features in crocoddyl, would you provide an upgrade strategy for eg. colmpc to even exist in a crocoddyl v4 world ? Maybe I am misunderstanding something here.

MaximilienNaveau:
Thank you very much @cmastalli for keeping this project alive!

As @nim65s suggested, could we have a preview of the new API and a migration plan?

Otherwise we could consider an off-tree solution (as suggested by @nim65s) for keeping the current downstream software alive until the migration is properly done on the user side.

cmastalli:
@nim65s and @nim65s -- We have been working for about 2 years on optimal estimation. These are reference publications of features we already developed:
  - https://arxiv.org/pdf/2403.17161
  - https://arxiv.org/pdf/2602.16358

These are changes we made, explained at a high level:

1. **Abstraction of dynamics to handle optimal estimation problems**. Currently, the differential action approach is messy. It mixes concepts such as dynamics with cost and constraints, limiting us to introducing disturbances or parametrized dynamics. This is the major deprecation we couldn't avoid.
   
2. **Simplifying holonomic constraints**. The current development of multiple contacts and impulses is quite annoying. And if we add closed-loop kinematics, the problem becomes worse. To reduce the project complexity, we decided to introduce a holonomic-constraint abstraction as described in Ludovic's PR.

3. **Numerical integration**. Current integrators are designed for control problems, but in optimal estimation, we need to ensure disturbances satisfy the implicit holonomic constraints. This is also connected to point 1.

4. **Solving optimal estimation problems**. Optimal estimation problems perform parametrized Riccati recursions and rollouts. This requires introducing different types of problems. Currently, we support shooting problems (i.e., control problems).

@Sergim96 is in charge of this major refactoring. He is actually quite advanced in this task, but it can take several more weeks. We don't know when we'll release this code, as some of the current efforts we still plan to publish later. However, if you guys are planning to introduce some serious features in Crocoddyl, then it is indeed worth arranging a meeting and working together.

We're happy to arrange a meeting with you in two weeks. Next week is complicated for both of us. Would this work for you? If so, please let's move this discussion via email. You can find our emails at https://www.romilab.org/team/.

